Issue #30, opened 2018-07-20
Add a way to promote blocks to actual files (and the reverse)


samoht:
In some use-case we want to explain how to build a project by quoting the contents of the files in that project. We would like the instructions to be kept up-to-date automatically, so we need a way to promote the contents of the file to the documents and/or the reverse.

PR #37, opened 2018-09-07
Promote files to blocks

gpetiot:
Fix issue #30 
Work in progress: it would be better to replace the "lines" attributes with "parts" (more robust).


gpetiot:
ocaml_topexpect should be migrated to 4.06.1 to be used in mdx.

samoht:
> Edit: @samoht does this have to be tackled before merging this PR ?

Not that's not necessary. Let me know when the PR is ready to be merged!

bin/promote/main.ml
```diff
@@ -0,0 +1,139 @@
+(*
+ * Copyright (c) 2018 Thomas Gazagnaire <thomas@gazagnaire.org>
```
samoht: You should add your name here :-)

bin/promote/ocaml_topexpect.ml
```diff
@@ -0,0 +1,513 @@
+module Chunk = struct
```
samoht: can you add some headers?

test/promote_file.md
```diff
@@ -0,0 +1,32 @@
+Mdx can also understand ocaml code blocks:
+
+```ocaml file=toto.ml,part=toto
```
samoht: that should be `file=promote.ml` right?

Also you need to add the relevant runes in the `dune` file.

test/promote_file.md.corrected
```diff
@@ -0,0 +1,64 @@
+Mdx can also understand ocaml code blocks:
```
samoht: Other tests use the file extension `.expected` (e.g. that's expected that the generated `.corrected` file is similar to that file).

gpetiot:
I rebased the branch against master but it there is still some issues on the promote_file test, I think I missed something with the jbuider to dune migration (weird, CI says it is ok, but not on my own machine).

For now, based on the last modification date of the .md file and the source file (mentioned in a block), for each block:
- if the .md file is more recent, the source file is updated (in place)
- if the source file is more recent, the .md file is updated (.md.corrected generated)

Does this fit your workflow?

samoht:
After some thoughts, I think the new `promote` command could work well if the semantics of `mdx test` for blocks with a `file=..` header is modified to match the following:

A typical user interaction would look like:

    $ cat README.md
    ```ocaml file=dir/promote.ml,part=toto
    let x = 34
    let f = 42.3
    let s = "toto"
    let f x u = u x
    ```

### Normal interaction

    $ mdx test
    [ the block is skipped as promote.ml should be tested/compiled separately ]

### Promotion from ml to md

    $ mdx test README.md
    [ ok ]
    [ edit dir/promote.ml ]
    $ mdx test
    [ generate README.ml.corrected ]
    $ dune promote                     # if dune rules are set
    $ mdx promote [--direction=to-ml]  # optional as dune could handle it today

The `--direction` option has a default value of `infer-timestamp` to use timestamp detection to infer the direction of the promotion. We could later add a `infer-vcs` to check which file has changed using `git-diff` if needed.
 
### Promotion from md to ml

    $ mdx test README.md
    [ ok ]
    [ edit README.md ]
    $ mdx test
    [ generate a dir/promote.ml.corrected ]
    $ mdx promote [--direction=to-ml]  # dune promote doesn't work here (I hink) 
                                       # as it needs to know about dir/promote.ml

### Integration with dune

In the longer term, maybe the right way to integrate with dune is to have a special command generating the right test stanza for dune, something like:

```
$ mdx rule README.md
(alias
  (name runtest)
  (deps (x: README.md) (y: dir/promote.ml))
   (action (progn
      (run mdx test %{x})
       (diff %{x} %{x}.corrected)
       (diff %{y} %{y}.corrected)))
```

Note: I have no idea is that rule is enough :-) In any case, we don't need it right now, we check if I can the new commands are enough for RWO already.

bin/cli.ml
```diff
@@ -40,6 +40,24 @@ let root =
...
+             accordingly. $(b,infer-timestamp) uses the files last \
+             modification timestamps to determine the reference file \
+             and updates the least recent files." in
+  let opt_names =
+    [ "infer-timestamp", `Infer_timestamp
+    ; "to-md", `To_md
+    ; "to-ml", `To_ml ]
+  in
+  let names = ["direction"] in
+  let docv = Astring.String.concat ~sep:"|" (List.map fst opt_names) in
```
samoht: You can just use `String.concat "|"` here (no need to bring astring in)

jbuilder/jbuild
```diff
@@ -0,0 +1,24 @@
+
```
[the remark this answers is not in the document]
gpetiot: Yes but it seems this is used instead of dune in the RWO v2 repo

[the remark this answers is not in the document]
samoht: this needs to be updated -- I'll do it at one point

Makefile
```diff
@@ -3,6 +3,9 @@
 all:
 	dune build
 
+install:
+	dune install
```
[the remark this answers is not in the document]
gpetiot: I am just used to do `make install` but we can skip it if nobody else uses it.

samoht:
Is this ready to merge/review?

gpetiot:
Yes it is ready

Makefile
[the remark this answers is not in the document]
samoht: yes please remove this - this is not usually a good idea to `make install` on behalf on opam.

bin/cli.ml
```diff
@@ -40,6 +40,24 @@ let root =
...
+             and updates the least recent files." in
+  let opt_names =
+    [ "infer-timestamp", `Infer_timestamp
+    ; "to-md", `To_md
+    ; "to-ml", `To_ml ]
+  in
+  let names = ["direction"] in
+  let docv = String.concat "|" (List.map fst opt_names) in
+  let docv = "{" ^ docv ^ "}" in
+  Arg.(value & opt (some (enum opt_names)) None & info names ~doc ~docv)
```
samoht: Set the default to a sensible value here (for instance `Infer_timestamp`) - that will update the man page automatically.

bin/test/main.ml
```diff
@@ -130,9 +137,65 @@ let run_toplevel_tests c ppf tests t =
     ) tests;
   Block.pp_footer ppf ()
 
+let update_block_with_file ppf t file part =
+  Block.pp_header ppf t;
+  let lines = Mdx_top.lines_of_part ~file ~part in
+  let contents = Astring.String.concat ~sep:"\n" lines in
```
samoht: You don't need `Astring` for this.

bin/test/main.ml
```diff
@@ -130,9 +137,65 @@ let run_toplevel_tests c ppf tests t =
...
+        | Toplevel tests ->
+           let f t =
+             t.Toplevel.command |> Astring.String.concat ~sep:"\n" in
+           List.map f tests
+      in
+      let lines =
+        Mdx_top.replace_lines_of_part
+          ~file:input_file ~part ~lines in
+      let lines = List.map (Astring.String.concat ~sep:"\n") lines in
+      let lines = Astring.String.concat ~sep:"\n" lines in
```
samoht: Same here, you don't need `Astring` for this.

jbuilder/jbuild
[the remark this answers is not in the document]
samoht: I am updating RWO to generate a dune file, so this can be updated. Also please move this in `test` and a a test for it.

lib/top/mdx_top.mli
```diff
@@ -26,3 +26,15 @@ val init: verbose:bool -> silent:bool -> verbose_findlib:bool -> unit -> t
 val eval: t -> string list -> string list
 (** [eval t p] evaluates the toplevel phrase [p] (possibly spawning on
     mulitple lines) with the configuration value [t]. *)
+
+val lines_of_part: file:string -> part:string option -> string list
+(** [lines_of_part f p] returns each line of the part [p] in file [f],
```
samoht: wrong docstring

lib/top/mdx_top.mli
```diff
@@ -26,3 +26,15 @@ val init: verbose:bool -> silent:bool -> verbose_findlib:bool -> unit -> t
...
+    this function fails if [f] does not contain a part [p]. *)
+
+val replace_lines_of_part:
+  file:string ->
+  part:string ->
+  lines:string list ->
+  string list list
+(** [replace_lines_of_part f p l] returns the lines of the file [f] where the
```
samoht: wrong docstring.

lib/top/ocaml_topexpect.ml
```diff
@@ -0,0 +1,319 @@
+(*
```
samoht: rename this file to something else, as it hasn't anything to do with ocaml_topexexpt anymore. Maybe `part.ml` with 2 exposed functions: `find` and `replace`?

jbuilder/jbuild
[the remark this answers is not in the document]
gpetiot: This file itself is not bringing any value compared to the dune file, so I will delete it. I don't think it is necessary to add a test for the promote mechanism, we just use `dune promote`, I suppose it is already tested in dune, and we already test that the generated .corrected files match the .expected files.

samoht:
I am fine to merge this, we can tune the behaviour that we really want after some use.

samoht:
Thanks!

PR #38, opened 2018-09-11
Multiple environments

gpetiot:
Fix issue #31 
WIP: Pervasives bindings are lost when switching environments

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -106,6 +106,56 @@ let run_cram_tests ?root ppf temp_file pad tests t =
     ) tests;
   Block.pp_footer ppf ()
 
+let envs = Hashtbl.create 8
+
+let rec save_summary acc s =
+  let open Env in
+  match s with
+  | Env_value (summary, id, _) ->
+     save_summary (Ident.name id :: acc) summary
+  | Env_module (summary, id, _)
+    | Env_class (summary, id, _)
+    | Env_modtype (summary, id, _)
+    | Env_cltype (summary, id, _)
+    | Env_type (summary, id, _)
+    | Env_functor_arg (summary, id)
+    | Env_open (summary,
+                #if OCAML_MAJOR >= 4 && OCAML_MINOR >= 7
+                _,
+                #endif
+                Pident id)
+    | Env_extension (summary, id, _) ->
+      let acc =
+        if Ident.binding_time id >= 1000
+        then Ident.unique_toplevel_name id :: acc
+        else acc
+      in
+      save_summary acc summary
+  | Env_empty -> acc
+  | Env_constraints (summary, _)
+    | Env_open (summary,
+                #if OCAML_MAJOR >= 4 && OCAML_MINOR >= 7
+                _,
+                #endif
+                _)
+    | Env_copy_types (summary, _) -> save_summary acc summary
+
+let in_env env_name f =
+  let env, names, objs =
+    try Hashtbl.find envs env_name
+    with Not_found ->
+      let env = Compmisc.initial_env () in
+      env, [], []
+  in
+  Toploop.toplevel_env := env;
+  List.iter2 Toploop.setvalue names objs;
+  let res = f () in
+  let env = !Toploop.toplevel_env in
+  let names = save_summary [] (Env.summary env) in
+  let objs = List.map Toploop.getvalue names in
+  Hashtbl.replace envs env_name (env, names, objs);
+  res
+
 let eval_test c test =
   Log.debug (fun l ->
       l "eval_test %a" Fmt.(Dump.list string) (Toplevel.command test));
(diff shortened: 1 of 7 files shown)


gpetiot:
@let-def I did what you suggested, i.e. reloading the bindings and imports when switching environments, but I don't get the expected result yet. Do you see anything wrong with this PR?

For example the following markdown file:
```
\```ocaml env=e1
let x = 1;;
\```

\```ocaml env=e2
let x = 3;;
\```

\```ocaml env=e1
# print_int x;;
1
- : unit = ()
\```
```
Should not be modified, but instead I get an error in the last block:
```
Exception: Misc.Fatal_error.
>> Fatal error: print_int unbound at toplevel
```

gpetiot:
More progress: in my last commit I got an issue when using `Toploop.getvalue`: CamlinternalFormatBasics and Pervasives and all operators and functions defined in those modules are said to be unbound at toplevel. It is the same whether `Ident.name` or `Ident.unique_toplevel_name` is used). Those modules are in the list of imported modules and I don't get an error when they are imported.

Here is a minimal example that works but that is not using Pervasives etc. (`x` having different values in 2 environments): https://gist.github.com/gpetiot/744a8f9456419692e5f1a4b8bdd414e1

gpetiot:
Still some issues with this feature (e.g. modules and exceptions), don't merge it now

let-def:
You missed the `Env_extension` case.

gpetiot:
I ignored this case for now because it raised an error on each exception name, and I got the same result whether I use the name `Ident.name` or `Ident.unique_toplevel_name`.

bin/test/main.ml
```diff
@@ -100,6 +100,42 @@ let run_cram_tests ?root ppf temp_file pad tests t =
...
+let envs = Hashtbl.create 8
+
+let rec save_summary acc s =
+  let open Env in
+  match s with
+  | Env_value (summary, id, _) ->
+     save_summary (Ident.name id :: acc) summary
+  | Env_module (summary, id, _)
+    | Env_class (summary, id, _) ->
+     save_summary (Ident.unique_toplevel_name id :: acc) summary
```
let-def: I suggest:

```
diff --git a/bin/test/main.ml b/bin/test/main.ml
index 597a117..e445166 100644
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -108,11 +108,17 @@ let rec save_summary acc s =
   | Env_value (summary, id, _) ->
      save_summary (Ident.name id :: acc) summary
   | Env_module (summary, id, _)
-    | Env_class (summary, id, _) ->
-     save_summary (Ident.unique_toplevel_name id :: acc) summary
+    | Env_class (summary, id, _)
+    | Env_extension (summary, id, _) ->
+      let acc =
+        if Ident.binding_time id >= 1000
+        then Ident.unique_toplevel_name id :: acc
+        else acc
+      in
+      save_summary acc summary
   | Env_empty -> acc
   | Env_modtype (summary, _, _)
-    | Env_extension (summary, _, _)
     | Env_cltype (summary, _, _)
     | Env_open (summary, _)
     | Env_functor_arg (summary, _)
```

Identifiers with a binding_time between 1 and 1000 are handled differently (see `Predef` module).

gpetiot: Thank you for your help :) 
But for some reason I still have some issues with modules, I get  and `Unbound module` error when reloading the objects on an existing environment. The name of the module is not in the summary of the environment, and it is not even in the module list (when calling `Env.fold_modules`).

let-def:
Do you have an example?

gpetiot:
Here is an example: https://gist.githubusercontent.com/gpetiot/cbef10e51354c386800f33c824a4b8c8/raw/3f168441c5090ca17661a32b39eb0ba0dce9f561/example.md

The command is:
`mdx test example.md && diff example.md example.md.corrected` and the output is:
```
16c16,17
< 42
---
> Characters 10-13:
> Error: Unbound module M
```

let-def:
This has nothing to do with the environments, it is just that the way parsing is implemented, it stops at the first `;;`.

Try with:

> ```ocaml env=e1
> let x = 1;;
> ```
> 
> ```ocaml env=e1
> module M = struct let k = 42 end;;
> ```
> 
> ```ocaml env=e2
> let x = 3;;
> ```
> 
> ```ocaml env=e1
> # print_int x;;
> 1
> - : unit = ()
> # print_int M.k;;
> 42
> ```

let-def:
(Erf, I don't know how to disable markdown)

gpetiot:
You are right, thank you! Sorry if I made you waste time on this.

gpetiot:
Some work is needed to rebase this PR on master now that it is needed to compile with both 4.06 and 4.07.  In particular the type `Env.summary` has changed and OMP does not handle Env migrations.
```
Error: The constructor Env_open expects 3 argument(s),
       but is applied here to 2 argument(s)
```

samoht:
Thanks!

PR #43, opened 2018-10-10
CI: faster and run tests

gpetiot:


gpetiot:
This should be good to go

samoht:
Not sure why `ocaml-ci-scripts` is slower; would be interesting to upstream your changes if you know the differences. An other option is to remove the minor version in the `OCAML_VERSION` env var so that we don't have to bootstrap a new compiler on every test.

PR #44, opened 2018-10-11
'rule' command to generate dune rules

gpetiot:


bin/rule.ml
```diff
@@ -0,0 +1,69 @@
...
+  Fmt.pr
+    "\
+(alias\n\
+\ (name   runtest)\n\
+\ (deps   (:x %s) (:y %s.expected)\n\
+\         (:a %s) (:b %s.expected) (package mdx))\n\
+\ (action (progn\n\
+\           (run mdx test %c{x})\n\
+\           (diff? %c{y} %c{x}.corrected)\n\
+\           (diff? %c{b} %c{a}.corrected))))\n\n"
```
samoht: I would have expected to have a list of `ml` files here, and have one rule per `md` file instead of one rule per block.  I am not sure that's even work properly :-)

PR #53, opened 2018-11-05
mdx does not use the cppo library, just the binary

avsm:
Remove the dependency here, as it breaks embedding the mdx
code in a duniverse when there is a cppo override present
in the source tree.


samoht:
Thanks!

PR #49, opened 2018-10-30
test: add a failing test when creating new types

samoht:
/cc @gpetiot

PR #65, opened 2018-11-09
Add a failing test-case for objects

samoht:
/cc @gpetiot


samoht:
```
>> Fatal error: istack/1798 unbound at toplevel
mdx-test: internal error, uncaught exception:
          Misc.Fatal_error
```

Seems similar to #49

Issue #70, opened 2018-11-14
Update README.md with new features


samoht:
/cc @gpetiot

avsm:
It would be very useful to update the README synchronously all the PRs go in if possible (ideally with a 'since version of mdx'). Otherwise it becomes a forensic effort to piece together the syntax :-)

samoht:
Updated in #101

PR #85, opened 2018-11-16
Improve ellipsis support, for some value of improve

dra27:
This PR changes two features of ellipsis support.

 1. If the output has changed, mdx tries to preserve ellipsis lines - before, they would all be expanded. A test is included.
 2. If an output line ends with an ellipsis, then only the rest of the line has to match - this allows very limited partial matching of lines (but useful for things like shell error messages where the prefix is portable but the precise text of the message is not)

The second change is obviously not wholly backwards compatible (an existing `Press any key to continue...` line of output will now match `Press any key to continue...!`, for example)

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -100,7 +100,7 @@ let run_cram_tests t ?root ppf temp_file pad tests =
       let output =
         let output = List.map (fun x -> `Output x) lines in
         if Output.equal output test.output then test.output
-        else output
+        else Output.merge output test.output
       in
       Cram.pp_command ~pad ppf test;
       List.iter (function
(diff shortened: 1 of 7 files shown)


samoht:
That looks like a good idea. Thanks for the patches!

samoht:
Could you also add a few notes about this feature in README.md?

trefis:
Ping :)

dra27:
Getting there - though probably not until later in the week! @trefis - this won't be an immediate silver bullet for Merlin, since MDX needs back-ported support for OCaml 4.02-4.05 added before it can be used for the testsuite

dra27:
(sorry - juggling OCaml 4.08 reviews and, unusually, providing a taxi service for my cat.......)

samoht:
I'm happy to merge this PR now and add documentation later (see #70)

PR #86, opened 2018-11-28
Support for 4.02.3

gpetiot:
4.02.3 support for PR #85 

I tried to minimize the impact on `mdx_top.ml` but there is a lot of differences between 4.02 and 4.06/07.

Regressions:

# 4.02.3

```
--- test/padding.md	2018-11-28 09:16:32.886114184 +0700
+++ test/padding.md.corrected	2018-11-28 09:19:40.622040315 +0700
@@ -13,5 +13,5 @@
 ```sh
 $ ocaml -warn-help | egrep '\b9\b'
   9 Missing fields in a record pattern.
-  R Alias for warning 9.
+  R warning 9
 ```

```
--- test/errors.md	2018-11-28 09:16:32.886114184 +0700
+++ test/errors.md.corrected	2018-11-28 09:19:40.670040350 +0700
@@ -33,7 +33,7 @@
 # let x =
   1 + "42"
 Characters 14-18:
-Error: This expression has type string but an expression was expected of type
+Error: This expression has type bytes but an expression was expected of type
          int
 ```

``` 
--- test/mlt.md	2018-11-28 09:16:32.886114184 +0700
+++ test/mlt.md.corrected	2018-11-28 09:19:40.718040384 +0700
@@ -6,6 +6,6 @@
 val x : int = 3
 # x + "foo";;
 Characters 4-9:
-Error: This expression has type string but an expression was expected of type
+Error: This expression has type bytes but an expression was expected of type
          int
 ```

```
--- test/lines.md	2018-11-28 09:16:32.886114184 +0700
+++ test/lines.md.corrected	2018-11-28 09:19:40.798040441 +0700
@@ -13,7 +13,7 @@
 val f : int -> int = <fun>
 # let f y =
   y^"foo"
-val f : string -> string = <fun>
+val f : bytes -> bytes = <fun>
 ```
 
 ```
--- test/lines.md	2018-11-28 09:16:32.886114184 +0700
+++ test/lines.md.corrected	2018-11-28 09:19:40.798040441 +0700
@@ -24,7 +24,7 @@
   | n ->
   n + "foo"
 Characters 45-50:
-Error: This expression has type string but an expression was expected of type
+Error: This expression has type bytes but an expression was expected of type
          int
 ```

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -15,11 +15,11 @@
  *)
 
 open Mdx
+open Compat
+open Result
 
 let src = Logs.Src.create "cram.test"
 module Log = (val Logs.src_log src : Logs.LOG)
-module Result = Mdx.Migrate_ast.Result
-module String = Mdx.Migrate_ast.String
 
 let (/) = Filename.concat
 
@@ -130,10 +130,10 @@ let err_eval ~cmd lines =
 let eval_raw t ?root c ~line lines =
   let test = Toplevel.{vpad=0; hpad=0; line; command = lines; output = [] } in
   match eval_test t ?root c test with
-  | Result.Ok _    -> ()
-  | Result.Error e -> err_eval ~cmd:lines e
+  | Ok _    -> ()
+  | Error e -> err_eval ~cmd:lines e
 
-let lines = function Result.Ok x | Result.Error x -> x
+let lines = function Ok x | Error x -> x
 
 let split_lines lines =
   let aux acc s =
@@ -324,8 +324,8 @@ let run ()
                    Block.pp ppf t;
                    List.iter (fun test ->
                        match eval_test t ?root c test with
-                       | Result.Ok _    -> ()
-                       | Result.Error e ->
+                       | Ok _    -> ()
+                       | Error e ->
                          let output = List.map (fun l -> `Output l) e in
                          if Output.equal test.output output then ()
                          else err_eval ~cmd:test.command e
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -18,6 +18,8 @@ open Mdx
 
 let src = Logs.Src.create "cram.test"
 module Log = (val Logs.src_log src : Logs.LOG)
+module Result = Mdx.Migrate_ast.Result
+module String = Mdx.Migrate_ast.String
 
 let (/) = Filename.concat
 
@@ -128,10 +130,10 @@ let err_eval ~cmd lines =
 let eval_raw t ?root c ~line lines =
   let test = Toplevel.{vpad=0; hpad=0; line; command = lines; output = [] } in
   match eval_test t ?root c test with
-  | Ok _    -> ()
-  | Error e -> err_eval ~cmd:lines e
+  | Result.Ok _    -> ()
+  | Result.Error e -> err_eval ~cmd:lines e
 
-let lines = function Ok x | Error x -> x
+let lines = function Result.Ok x | Result.Error x -> x
 
 let split_lines lines =
   let aux acc s =
@@ -322,8 +324,8 @@ let run ()
                    Block.pp ppf t;
                    List.iter (fun test ->
                        match eval_test t ?root c test with
-                       | Ok _    -> ()
-                       | Error e ->
+                       | Result.Ok _    -> ()
+                       | Result.Error e ->
                          let output = List.map (fun l -> `Output l) e in
                          if Output.equal test.output output then ()
                          else err_eval ~cmd:test.command e
--- a/lib/compat.ml
+++ b/lib/compat.ml
@@ -17,11 +17,11 @@
 module String = struct
   include String
 
-#if OCAML_MAJOR >= 4 && OCAML_MINOR < 3
+#if OCAML_MAJOR = 4 && OCAML_MINOR < 3
     let equal x y = String.compare x y = 0
 #endif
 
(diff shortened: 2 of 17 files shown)


lib/migrate_ast.ml
```diff
@@ -109,6 +119,27 @@ module Pparse = struct
     |> to_406.copy_structure
 end
 
+module String = struct
+  include String
+
+#if OCAML_MAJOR >= 4 && OCAML_MINOR < 3
```
trefis: Shouldn't this be `... OCAML_MAJOR = 4 ...`?

samoht:
```
-<h1>–Test4, ?toto! a.nd. I/O</h1>
+<h1>--Test4, ?toto! a.nd. I/O</h1>
```
This is due of your version of pandoc. Try to install a more recent version to fix it.

samoht:
The regression in the test suite is perfectly fine, as it shows real changes between compiler versions!

I guess the tests need to be updated. One solution is to have enable/disable some test cases depending on the compiler version. A longer term solution (not in this scope of this PR) would be to add something like `ocaml>4.02` in the block metadata to only run the code example if the version of OCaml matches (and otherwise print a warning on the command-line).

bin/test/main.ml
```diff
@@ -18,6 +18,8 @@ open Mdx
 
 let src = Logs.Src.create "cram.test"
 module Log = (val Logs.src_log src : Logs.LOG)
+module Result = Mdx.Migrate_ast.Result
```
samoht: that's not the right way to support this -- you should use [Result](https://github.com/janestreet/result)

bin/test/main.ml
```diff
@@ -128,10 +130,10 @@ let err_eval ~cmd lines =
 let eval_raw t ?root c ~line lines =
   let test = Toplevel.{vpad=0; hpad=0; line; command = lines; output = [] } in
   match eval_test t ?root c test with
-  | Ok _    -> ()
-  | Error e -> err_eval ~cmd:lines e
+  | Result.Ok _    -> ()
```
samoht: Using `Result.Ok` is not idiomatic at all :-)

lib/migrate_ast.ml
```diff
@@ -109,6 +119,27 @@ module Pparse = struct
     |> to_406.copy_structure
 end
 
+module String = struct
```
samoht: Could you move these modules to a new `Compat` module or something similar? Not sure why they are inside `Migrate_ast`

gpetiot:
> The regression in the test suite is perfectly fine, as it shows real changes between compiler versions!
> 
> I guess the tests need to be updated. One solution is to have enable/disable some test cases depending on the compiler version. A longer term solution (not in this scope of this PR) would be to add something like `ocaml>4.02` in the block metadata to only run the code example if the version of OCaml matches (and otherwise print a warning on the command-line).

First draft in #87

avsm:
This lgtm, and helps with future integration into Dune which also supports 4.02.3. Thanks!

samoht:
Could you rebase this now that #87 is merged?

samoht:
Thanks!

lib/compat.ml
```diff
@@ -0,0 +1,106 @@
...
+  include Warnings
+
+#if OCAML_MAJOR = 4 && OCAML_MINOR < 3
+  (* Can't be overriden *)
+  let reset_fatal () = ()
+#endif
+end
+
+module Env = struct
+  include Env
```
hhugo: I suspect this break the `BLACK MAGIC` in https://github.com/realworldocaml/mdx/blob/master/lib/top/mdx_top.ml#L631
and https://github.com/realworldocaml/mdx/blob/master/lib/top/mdx_top.ml#L654.

I've no idea why we need to patch `Env.without_cmis`, It would be nice to have a comment about this.

@samoht , do you why we want to patch `Env` ?

cc @xclerc

hhugo:
I suspect this break the BLACK MAGIC in https://github.com/realworldocaml/mdx/blob/master/lib/top/mdx_top.ml#L631
and https://github.com/realworldocaml/mdx/blob/master/lib/top/mdx_top.ml#L654.

I've no idea why we need to patch Env.without_cmis, It would be nice to have a comment about this.

@samoht , do you why we want to patch Env ?

cc @xclerc

lib/compat.ml
gpetiot: This is done because the function does not exist before OCaml 4.03, so we need to define a dummy function that does nothing.

let-def:
There is a buggy interaction between `Env.without_cmis` and the printer.
In normal compilation, printing happens only when typing has finished so this is not a problem but mdx interleaves typechecking and printing.

More precisely, the problem is that `Env.without_cmis` prevent loading of CMIs and this interact badly with the lazy construction of typing environment.
If printing triggers the load of an external module, it will fail and the module will be persistently broken in the environment.

In recent versions of OCaml, @lpw25 wrote a fix that backtracks the damage that were done to the environment.

At least this ensure that the typer will behave correctly, but I am not sure this is enough for our case. Some aliases might be missed and the printing will still be wrong.

I suggest trying with and without the *BLACK MAGIC*: if the output is the same, it can probably be removed.

lib/compat.ml
let-def: Reexporting `Env` should be fine. The black magic is necessary because the `Printing` module of the compiler is already linked to the global version of Env, so we need some surgery to fix WTF behaviors.

PR #87, opened 2018-12-02
Ocaml version

gpetiot:
Version check for #86


samoht:
I'm merging this in its current state, but it would be nice to support more easily multiple versions of the compilers (e.g. by adding support to version comparisons).

samoht:
I've created https://github.com/realworldocaml/mdx/issues/88 to keep track of improvements of that feature.

Issue #104, opened 2019-02-14
Reintroduce craml syntax


trefis:
I'm using mdx only for tests, not for fancy documents, and for that purpose [craml](https://github.com/realworldocaml/craml#craml-a-tool-for-testing-command-line-applications)'s syntax was just more convenient.

I'm wondering if it could be reintroduce either with a `--syntax cram` flag, or by simply assuming that if the input is a `.t` file then it will be written in the cram syntax.

avsm:
There's nothing imcompatible about having this just integrated by default in mdx as well, is there?  It seems like a useful thing to have alongside the existing codeblock markdown format.

samoht:
Fixed by #106

PR #106, opened 2019-02-18
Reintroduce craml syntax

trefis:
Enabled by `--syntax=cram`, that seemed cleaner than matching on the file ext.

### DIFF
--- a/bin/dune
+++ b/bin/dune
@@ -1,7 +1,7 @@
 (library
  (name      cli)
  (modules   cli)
- (libraries cmdliner fmt.cli logs.fmt fmt.tty logs.cli))
+ (libraries cmdliner fmt.cli logs.fmt fmt.tty logs.cli mdx))
 
 (executable
  (name        main)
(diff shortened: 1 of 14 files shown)


trefis:
This fixes #104.
I ran it over merlin's testsuite (on this branch: https://github.com/ocaml/merlin/tree/mdx).

It's probably not the prettiest patch, but it would be greatly appreciated if we could get a release of mdx on opam including this patch (or a variation of it) so I can reenable merlin's testsuite ASAP.

samoht:
Thanks!

bin/dune
```diff
@@ -1,7 +1,7 @@
 (library
  (name      cli)
  (modules   cli)
- (libraries cmdliner fmt.cli logs.fmt fmt.tty logs.cli))
+ (libraries cmdliner fmt.cli logs.fmt fmt.tty logs.cli mdx))
```
samoht: is this needed ? Not sur how it worked before

trefis: I introduced a dependency on `Mdx.Syntax.t` because I provide cmdliner converters for it. Before that there was no dependency.

PR #107, opened 2019-02-19
Correction of tests on ellipsis/cram

clecat:
Tests are not passing on my computer:
Indeed, the new tests using "echo" does not specify the flag -e which, AFAIK, are needed for the command to interpret most of the backslash escapes.

Did the tests pass on your computer @trefis ? I thought the "-e" option was common to every implementation of echo

### DIFF
--- a/test/cram.t
+++ b/test/cram.t
@@ -10,7 +10,7 @@ Long lines can be replaced by ellipsis:
   ...
   10
 
-  $ echo "foo\"\n\nbar"
+  $ echo -e "foo\"\n\nbar"
   foo"
   
   ...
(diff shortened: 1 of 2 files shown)


trefis:
That test wasn't passing for me, but it worked on travis so I assumed the problem was on my side.

clecat:
Ok, so apparently, the implementation of echo on travis is not the same as mine (and I suppose yours).
I'm trying to remove it for printf, which, from what I rapidly checked, follows a more strict implementation according to POSIX standard

clecat:
ex: https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

samoht:
Thanks!

Issue #98, opened 2019-01-21
Uncaught exception on String.sub


gpetiot:
The following raises an error:

\`\`\`ocaml
\# print_newline ();;
\`\`\`

```
$ mdx test a.md
mdx-test: internal error, uncaught exception:
          Invalid_argument("String.sub / Bytes.sub")
```

clecat:
See this [pull request](https://github.com/realworldocaml/mdx/pull/108)

PR #108, opened 2019-02-20
Solved bug occuring when a top-level command outputs an empty line

clecat:
Short commit solving a bug occurring when a top-level command generates an empty line output, which throws an `Invalid_argument("String.sub / Bytes.sub") error`.
e.g:
```ocaml
# print_newline ();;
```
Add an additional test checking if the error does not occurs anymore.

Fix issue [#98](https://github.com/realworldocaml/mdx/issues/98)

#### Details:
In file mdx_top.ml, function trim_line, line 361 to 366:
When the string str is "\n",
 * len = 1
 * str.[0] = '\n', therefore trim_from is set to 1
 * str.[len - 1] = str.[0] = '\n', therefore trim_to is set to len - 1 = 0

Then String.sub is called with 0 and 0 - 1 = -1, which raises the error

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -363,7 +363,7 @@ let trim_line str =
   if len = 0 then str else
     let trim_from = if str.[0] = '\n' then 1 else 0 in
     let trim_to = if str.[len - 1] = '\n' then len - 1 else len in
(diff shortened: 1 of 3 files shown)


samoht:
Thanks!

Issue #100, opened 2019-02-03
rename binary to ocamlmdx


avsm:
This will likely be an ocaml-specific tool for some time, so it'll ease packaging if we rename it to ocamlmdx sooner rather than later. Especially before it starts getting embedded in 3rd party projects...

clecat:
Should it be `ocamlmdx` or `ocaml-mdx` ?

avsm:
ocaml-mdx seems cleaner

clecat:
I added a small warning when using mdx instead of ocaml-mdx: is the message good enough or do you have special expectations for his format ? @avsm

clecat:
Better with the message ;)
```
Warning: 'mdx' is deprecated and will be removed in the futur
    Use 'ocaml-mdx' instead
```

PR #110, opened 2019-02-22
Renaming mdx to ocaml-mdx

clecat:
Solves issue #100 

Generates additional binaries called 'ocaml-mdx' and 'ocaml-mdx-test' for a better clarity about what they are doing. As we do not want to break everything, we will let the former binaries for now, before removing them.

#### Details

Changed the executable rules and added the former binary in a new rule: this way, we wont need to change the rules, just remove the `install` when we will want to get rid of them.

I'm kinda new to the dune/opam stuff, so i'm not sure if it will be enough: Tell me if I actually forgot something :D

### DIFF
--- a/bin/test/dune
+++ b/bin/test/dune
@@ -1,8 +1,13 @@
 (executable
  (name        main)
- (public_name mdx-test)
+ (public_name ocaml-mdx-test)
  (package     mdx)
  (modes       byte)  ;; no native code toplevel libs yet
  (link_flags  -linkall)
  (preprocess (action (run %{bin:cppo} -V OCAML:%{ocaml_version} %{input-file})))
  (libraries   cli mdx.top))
+
+(install
+ (section bin)
+ (files
+  (main.exe as mdx-test)))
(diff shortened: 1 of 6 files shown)


bin/test/dune
```diff
@@ -1,8 +1,13 @@
...
  (package     mdx)
  (modes       byte)  ;; no native code toplevel libs yet
  (link_flags  -linkall)
  (preprocess (action (run %{bin:cppo} -V OCAML:%{ocaml_version} %{input-file})))
  (libraries   cli mdx.top))
+
+(install
+ (section bin)
+ (files
+  (main.exe as mdx-test)))
```
samoht: I think it would be nice to not have both `mdx-test` and `ocaml-mdx-test`. Nobody is using `mdx-test` directly as far as I know.

clecat: Done :)

clecat: Bad commit name, see the second one :)

samoht:
Looks good, thanks!

Issue #112, opened 2019-02-28
Support for expect tests


samoht:
We should check that one can run an expect test, and have corrections work for that test, and then include the results in the mdx.  Mostly I think this would be about getting the dune rules to work properly, rather than mdx as a tool.

(e.g. we should at least document how use ppx_expect with mdx)

clecat:
Are the ppx_expect tests usable in the top_level ? I would be glad is someone using regulartly could point me a way to do it

clecat:
I proposed [this issue](https://github.com/janestreet/ppx_expect/issues/14), which, if accepted and implmented, would allow ppx_expect to be used in a top_level.

clecat:
Coming back with a question: Are you sure that you want to run the ppx_expect tests inside of the markdown, or do you only want to write some ppx_expect use cases, and continue without taking care of the test result ?
The thing is, mdx uses the top_level, and ppx_expect is not compatible with the top_level.
- One possible solution, which should work on short-term, would be to deactivate the test if it is run from a top_level instead of raising an error (because inline tests were not activated).
- On the other hand, we could change the ppx_expect code so it would run the actual tests as a diff when run from a top_level, but it may take more time to be done (it's far from what ppx_expect currently does, but it's still doable).
What is the behaviour you want ?

clecat:
Third solution, might be the best (was actually the one I had in mind, but looks clear to me now):
- Instead of supporting the promotion, we do what I proposed in the issue mentioned earlier: We will not be able to propose the promoted source code, but we would still have the diff proposed.
Could be a good in between

samoht:
Could you give an example of the workflow you are describing? I am not sure to follow.

clecat:
Let's consider the following examples:
ex 1: Correct test
```ocaml
# let%expect_test _ =
     let f output =
       print_string output;
       [%expect {| hello world |}]
     in
     f "hello world"
```
ex2: Bad test: "error" <> "hello world"
```ocaml
# let%expect_test _ =
     let f output =
       print_string output;
       [%expect {| hello world |}]
     in
     f "error"
```

#### Solution 1: Lazy solution
Tests activated from the top_level are ignored:
The examples 1 and 2 are ignored, raising no error, but we do not know that the ex 2 if wrong

#### Solution 2: Complex solution
Tests activated from the top_level are run and the source code if corrected:
The examples 1 and 2 are properly run, raising an error for the second case, giving us the corrected code:
```ocaml
[...]

PR #118, opened 2019-04-01
Added an argument allowing to force generation of diff file

clecat:
As said, I added an argument to ocaml-mdx which forces the generation of a corrected file, even without diff.
It allows to really on the file in dune rules (e.g [this example](https://github.com/realworldocaml/mdx/issues/112))

### DIFF
--- a/lib/mdx.mli
+++ b/lib/mdx.mli
@@ -69,12 +69,12 @@ val parse_lexbuf: syntax -> Lexing.lexbuf -> t
 
 (** {2 Evaluation} *)
 
-val run: ?syntax:syntax -> force_output:bool -> f:(string -> t -> string) -> string -> unit
+val run: ?syntax:syntax -> ?force_output:bool -> f:(string -> t -> string) -> string -> unit
 (** [run ?syntax ~f n] runs the expect callback [f] over the file named
    [n]. [f] is called with the raw contents of [n] and its structured
    contents; it returns the new file contents. If the result of [f] is
-   different from the initial contents, then [$n.corrected] is created
-   with the new contents. *)
+   different from the initial contents of force_output was set to true, then
+   [$n.corrected] is created with the new contents. *)
 
 (** {2 Filtering} *)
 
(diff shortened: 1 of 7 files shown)


lib/mdx.mli
```diff
@@ -69,7 +69,7 @@ val parse_lexbuf: syntax -> Lexing.lexbuf -> t
 
 (** {2 Evaluation} *)
 
-val run: ?syntax:syntax -> f:(string -> t -> string) -> string -> unit
+val run: ?syntax:syntax -> bool -> f:(string -> t -> string) -> string -> unit
```
samoht: can you add a label here?

clecat: Done

lib/mdx.mli
```diff
@@ -69,7 +69,7 @@ val parse_lexbuf: syntax -> Lexing.lexbuf -> t
 
 (** {2 Evaluation} *)
 
-val run: ?syntax:syntax -> f:(string -> t -> string) -> string -> unit
+val run: ?syntax:syntax -> force_output:bool -> f:(string -> t -> string) -> string -> unit
```
samoht: you can even use an optional argument (with default `false`) and update the doc strings :-)

clecat: Done again, my bad for the docstring :p

lib/mdx.mli
```diff
@@ -69,12 +69,12 @@ val parse_lexbuf: syntax -> Lexing.lexbuf -> t
...
 (** {2 Evaluation} *)
 
-val run: ?syntax:syntax -> f:(string -> t -> string) -> string -> unit
+val run: ?syntax:syntax -> ?force_output:bool -> f:(string -> t -> string) -> string -> unit
 (** [run ?syntax ~f n] runs the expect callback [f] over the file named
    [n]. [f] is called with the raw contents of [n] and its structured
    contents; it returns the new file contents. If the result of [f] is
-   different from the initial contents, then [$n.corrected] is created
-   with the new contents. *)
+   different from the initial contents of force_output was set to true, then
```
samoht: s/of/and ?

clecat: or* I will definitely make it one day

samoht:
Thanks!

Issue #113, opened 2019-02-28
Add support to set environment variables


samoht:
Something like this should work:

    ```ocaml $FOO=bar
     # print_newline (Sys.getenv "FOO")
     bar
     ```

PR #120, opened 2019-04-03
Added environment variable support in code block

clecat:
As requested in [this](https://github.com/realworldocaml/mdx/issues/113) issue, it is now possible to create environment variable in a block

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -141,14 +141,15 @@ let pp ?syntax ppf b =
   pp_footer ?syntax ppf ()
 
 let labels = [
-  "dir"              , [`Any];
-  "source-tree"      , [`Any];
-  "file"             , [`Any];
-  "part"             , [`Any];
-  "env"              , [`Any];
-  "skip"             , [`None];
-  "non-deterministic", [`None; `Some "command"; `Some "output"];
-  "version"          , [`Any];
+  `Label "dir"              , [`Any];
+  `Label "source-tree"      , [`Any];
+  `Label "file"             , [`Any];
+  `Label "part"             , [`Any];
+  `Label "env"              , [`Any];
+  `Label "skip"             , [`None];
+  `Label "non-deterministic", [`None; `Some "command"; `Some "output"];
+  `Label "version"          , [`Any];
+  `Prefix "set-"            , [`Any];
 ]
 
 let pp_value ppf = function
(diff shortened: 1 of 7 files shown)


clecat:
After a second tought, it appears that we cannot declare 2 env variables in the same block:
Doing this solves it:
```
```ocaml var=FOO=bar 
`` `
```ocaml var=BAR=foo 
__your_stuff__
`` `
```
It is ugly, but we have the problem of the separator on the variables, which could actually be in the value of the variable

samoht:
Why can't we define multiple variables in the same block? We should take time to design a proper solution where this is possible. Is it a matter of syntax or code?

For the syntax, `var=FOO=bar` looks a bit weird as it is difficult to "parse" the equal signs.

Maybe something like this?

    ``` ocaml set-FOO=bar,set-BAR=foo
    ```

or (if `:` are not allowed in variable name):

    ``` ocaml var=FOO:bar,var=BAR:foo
    ```

clecat:
Actually, you can only do 1 call to the same label (the second call is ignored) and you cannot have dynamic labels:
I pushed that as a temporary solution in case some people needed that feature but I will solve those 2 issues

clecat:
Ok, so I did a fast modification, so now you can do that:
   ``` ocaml set-FOO=bar,set-BAR=foo
   ```
Tell me if you think that's enough (or if you want some more things).
But dont merge for now, I need to make this prefixed label a normal behaviour so we can use it in the futur for other purposes

lib/block.ml
```diff
@@ -141,14 +143,15 @@ let pp ?syntax ppf b =
...
 let labels = [
-  "dir"              , [`Any];
-  "source-tree"      , [`Any];
-  "file"             , [`Any];
-  "part"             , [`Any];
-  "env"              , [`Any];
-  "skip"             , [`None];
-  "non-deterministic", [`None; `Some "command"; `Some "output"];
-  "version"          , [`Any];
+  {s = "dir"              ; t = `Label ; vs = [`Any]};
```
samoht: I think it would be clearer to write:

```ocaml
let labels = [
  `Label "dir"   , [`Any];
  `Prefix "set-", [`Any];
]
```

What do you think?

clecat: I love it, would allow to put back the List.assoc and snd. Why I did not think about it before

samoht:
That looks great!

Before merging this, could you:
- squash your changes
- add an entry in README.md to advertise that new feature.

Many thanks!

clecat:
Well, I'm learning the squash, dont judge too much please :p

samoht:
Perfect, thanks!

PR #121, opened 2019-04-05
Support for 4.08

xclerc:
(Joint work with @hhugo)

This is clearly a work-in-progress, but it sounds useful to
share it, at least to avoid duplication.

This versions allows to build with 4.08, but tests fails.
There are (at least) two pain points:

- the black magic discussed by @hhugo [here](https://github.com/realworldocaml/mdx/pull/86);
- the conversion of the condition `Ident.binding_time id >= 1000`
  (I thought it meant "is not builtin", but the `is_builtin` function
  I added is apparently not correct).

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -18,6 +18,12 @@
 open Mdx.Migrate_ast
 open Mdx.Compat
 
+#if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
+let try_finally = Misc.try_finally
+#else
+let try_finally f ~always = Misc.try_finally f always
+#endif
+
 module Toploop = struct
   include Toploop
 
@@ -42,8 +48,8 @@ let redirect ~f =
     read_up_to := pos;
     Buffer.add_channel buf ic len
   in
-  Misc.try_finally (fun () -> f ~capture)
-    (fun () ->
+  try_finally (fun () -> f ~capture)
+    ~always:(fun () ->
        close_in_noerr ic;
        Unix.close fd_out;
        Unix.dup2 stdout_backup Unix.stdout;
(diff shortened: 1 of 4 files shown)


gpetiot:
I didn't manage to build your branch with 4.08 (using this switch https://discuss.ocaml.org/t/ocaml-4-08-0-beta1-release/3325), did you push everything? Are you using specific package versions?

xclerc:
I might have been using a patched compiler;
could you copy/paste the output of a failing build?

gpetiot:
I have a long sequence of:
```
File "lib/migrate_ast.ml", line 34, characters 2-9:
Error (alert deprecated): module Ast_406
Access modules via the Migrate_parsetree toplevel module. Use Migrate_parsetree.Ast_406 instead.
```
When fixed, I get:
```
File "lib/top/mdx_top.ml", line 725, characters 27-57:
Error: Unbound module Toplevel_backend
```

xclerc:
Indeed, sorry for that; I have pushed an additional
commit that should fix the build.

lib/top/mdx_top.ml
```diff
@@ -661,7 +726,8 @@ let init ~verbose:v ~silent:s ~verbose_findlib () =
   Compmisc.init_path true;
   Toploop.toplevel_env := Compmisc.initial_env ();
   Sys.interactive := false;
-  patch_env ();
+  (* CR hheuzard: fixme *)
+  (* patch_env (); *)
```
trefis: What is happening here? I think this should be uncommented.

xclerc: You are right; https://github.com/realworldocaml/mdx/pull/121/commits/6856c56cd39074ea53288861ea283b867bca8209 restores the call.

trefis:
Ping.
I think this can be merged now.

We've been using it internally at janestreet for a few days, and I have also been using it for testing merlin on 4.08.

trefis:
And ping again.
4.08 is very likely to be merged before the end of the decade, and having mdx work from day one would be nice.

samoht:
Thanks! Let me know if you need a new release.

trefis:
Thanks!
Yes, a new release would be nice. So that I could setup travis tests for 4.08 on merlin's repository without having to pin this repository.

PR #122, opened 2019-04-09
Fix dune example in README

andreypopp:


samoht:
Maybe someone should use `mdx` :p 

Many thanks for the patch!

PR #123, opened 2019-04-14
Add support for transforming syntax to Reason/OCaml

piotrdubiel:
Initial step towards https://github.com/realworldocaml/book/issues/3066

- [x] add cli flag for a print form (either ocaml or reason),
- [x] add parsing of reason blocks,
- [x] tranform blocks,
- [ ] transform toplevel phrases,
- [ ] output both syntaxes to allow switch between them.

It's still WIP since I don't know how to approach some things like outputting both blocks for switching later.


bin/output.ml
```diff
@@ -55,22 +55,47 @@ let pp_cram ppf (t:Mdx.Cram.t) =
...
         ("data-filter-output", ">");
       ]
-    | OCaml  -> Some "ocaml", pp_contents b, []
-    | Cram t -> Some "bash" , (fun ppf -> pp_list pp_cram ppf t.tests), [
+    | OCaml, `OCaml -> Some "ocaml", pp_contents b, []
+    | OCaml, `Reason -> (
+      let new_block = try_converting_to_reason b in
+      match new_block.value with
+        | Reason -> Some "reason", pp_contents new_block, []
+        | _ -> Some "ocaml", pp_contents b, [] (* transform to reason failed, so let's fallback to ocaml *)
```
avsm: Under what circumstances would a transformation from OCaml to Reason syntax fail?  Dropping back silently seems like a bad idea as it would leave some blocks in different syntaxes.

piotrdubiel: I did it, because in https://github.com/realworldocaml/book there are some examples which doesn't have a valid syntax (mostly intentionally, I guess).

For example
```ocaml
val remember : '_a -> '_a = <fun>
```
in Imperative Programming chapter.

It made sense in context of transforming the book, but I agree that for mdx as a standalone tool it's confusing. Do you want me to change it to fail fast?

avsm: > Do you want me to change it to fail fast?

We do need some alternate syntax for failing syntax fragments indeed.  I'm not sure what that is yet... some sort of alternates mode in mdx seems necessary to pass this through while selecting the right language.

piotrdubiel: What about adding label to code blocks?
For example
```
```ocaml unparsable
```

samoht: these unparsable bits are normally wrapped with a `skip` keyword to tell mdx to ignore them

avsm: I do think we need some alternates mechanism for this. I'm not keen on a 95% solution that'll leave some OCaml syntax lingering around to confuse a ReasonML-only reader...

XVilka:
There are conflicts now that should be resolved.

idkjs:
Is this ongoing? Can I help?

PR #124, opened 2019-04-14
add Jupyter output

avsm:
This is a WIP; creating a PR to make it easy for @jonludlam and me to push to our forks


avsm:
Some quality of life improvements: supports Markdown sections and raw cells now

avsm:
Noticed a problem that causes divergence between mdx evaluation and Jupyter.  The latter doesn't support multiple environments, so having two OCaml environments will cause them to be silently merged into one with the PR.

gpetiot:
Hi, should I revive this PR or is the environment issue something that will prevent it from working at all?

CraigFe:
I have a rebased form of this PR at [CraigFe/jupyter](https://github.com/CraigFe/mdx/tree/jupyter) if you do decide to go ahead with this.

avsm:
thanks, I've pushed your rebase here. @NathanReb we could go ahead with merging this; the only restriction it would be nice to have before that is to ensure that there is only one environment per notebook.

CraigFe:
From playing around with this PR, I noticed that it doesn't yet support file includes. I have a fairly hacky patch for that [here](https://gist.github.com/CraigFe/afe8530d0c4657558540d4b31e17eef1), but it probably warrants a cleaner implementation (that doesn't just `fail` for a large number of block types).

avsm:
Could we get this merged as an experimental feature @gpetiot, so that @CraigFe's variant above can be tracked? It's already useful as-is despite the limitations.

avsm:
rebased to current trunk

Issue #105, opened 2019-02-14
support mdx running in a local dune repo


avsm:
For using mdx to generate README files, it would be convenient if it had the same toplevel environment as that provided by `dune utop`.  This allows locally built packages to be used to process mdx files.

PR #127, opened 2019-06-24
[RFC] Add the --require option

Julow:
This option loads a list of packages into the toplevel, like the `-require` option of utop.

To fix https://github.com/realworldocaml/mdx/issues/105

The dune rule can then be rewritten as:

```dune
(alias
 (name   runtest)
 (deps   (package a) (:x README.mdx))
 (action (progn
           (run ocaml-mdx-test --require a --direction=infer-timestamp %{x})
           (diff? %{x} %{x}.corrected))))
```

However, the user has to specify the required packages, for example, in the mdx file (TODO).

Also, this only works with installable packages. So it not possible to write examples that requires private libraries but will be a pain in some cases (like rwo).

What do you think?

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -88,6 +88,10 @@ let force_output =
   let doc = "Force generation of corrected file (even if there was no diff)" in
   Arg.(value & flag & info ["force-output"] ~doc)
 
+let required_packages =
+  let doc = "Packages to load before executing tests, separated by a comma ($(b,,))." in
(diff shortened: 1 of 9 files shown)


samoht:
Can you add a test?

Julow:
Added a test.

The `--require` option may not be necessary after all. Specifying dependencies in the mdx file should be enough. I'll make a PR for that.

PR #130, opened 2019-06-25
Specify and load required packages

Julow:
Fixes https://github.com/realworldocaml/mdx/issues/105

This PR adds the `require` label:

```
```ocaml require=lib1,lib2
```

This will load `lib1` and `lib2` into the toplevel before executing the block. Using `findlib`.

Like with https://github.com/realworldocaml/mdx/pull/127, local libraries must be installable (have a `public_name`).

To work properly, the dune rule must be updated when the list of packages changes (eg. with `ocaml-mdx rule`).

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -744,6 +744,11 @@ let init ~verbose:v ~silent:s ~verbose_findlib () =
   silent t;
   t
 
+let load_packages _ pkgs =
+  try Result.Ok (Topfind.load_deeply pkgs)
+  with Fl_package_base.No_such_package (name, _) ->
+    Result.Error name
+
 module Part = Part
 
 let envs = Hashtbl.create 8
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -29,7 +29,7 @@ let prepend_root r b = match r with
   | None   -> b
   | Some r -> Filename.concat r b
 
-let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root options =
+let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires options =
   let ml_files = String.Set.elements ml_files in
   let ml_files = List.map (prepend_root root) ml_files in
   let dirs = match root with
@@ -44,6 +44,10 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root options =
     let f (cpt, acc) _ = cpt + 1, ("y" ^ string_of_int cpt) :: acc in
     List.fold_left f (0, []) ml_files |> snd
   in
+  let requires = String.Set.elements requires in
+  let pp_package_deps fmt name =
+    Fmt.pf fmt "\         (package %s)" name
+  in
   let pp_ml_deps fmt (var_name, ml_file) =
     Fmt.pf fmt "\         (:%s %s)" var_name ml_file
   in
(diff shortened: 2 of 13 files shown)


lib/top/mdx_top.ml
```diff
@@ -744,6 +744,11 @@ let init ~verbose:v ~silent:s ~verbose_findlib () =
   silent t;
   t
 
+let load_packages _ pkgs =
+  try Ok (Topfind.load_deeply pkgs)
```
samoht: This should be `Result.Ok` to keep compatibility with 4.02..

samoht:
It's failing on 4.02 because `'a result` was not in the standard library yet. Could you fix it and I think that's good to merge.

/cc @jonludlam if you want to try this with odoc

Julow:
Fixed.

@avsm Do this PR fixes your issue ? (https://github.com/realworldocaml/mdx/issues/105)
Cons: It uses Findlib, libraries must be installable and the dune rule must be updated when the dependencies change.

samoht:
> and the dune rule must be updated when the dependencies change.

But `mdx rule` will generate the right rule, right?

Julow:
> But `mdx rule` will generate the right rule, right?

Yes.

jonludlam:
Looks great! I'll give it a whirl.

jonludlam:
Works nicely! Is there a plan to have it work without the libraries being installed? It doesn't quite perfectly fit my use case without that (though I can certainly use this as-is now).

jonludlam:
Actually I'd not tried the dune integration - I'm getting this:

```
File "src/xref2/dune", line 6, characters 0-205:
 6 | (alias
 7 |  (name   runtest)
 8 |  (deps   (:x test.md)
 9 |          (package odoc.xref_test))
10 |  (action (progn
11 |            (run ocaml-mdx test --direction=infer-timestamp %{x})
12 |            (diff? %{x} %{x}.corrected)
13 | )))
Error: No rule found for alias .odoc.xref_test-files
     patdiff (internal) (exit 1)
```

jonludlam:
Ah I see, if I change that to get rid of the `.xref_test` it works without the libs installed. Just that to fix then. Wonderful!

jonludlam:
Hmm, how does this interact with specifying environments? I can't seem to do both.

jonludlam:
Also, if you accidentally leave a trailing space after the `require=...` it doesn't work in a quite confusing way :-)

jonludlam:
Ah, I think it might work if you do: `require=foo,env=e1` - I was putting a space in between.

In any case, I've hit an issue:

     ```ocaml require=tyxml
     # open Tyxml 
     ```
causes for me:

```
$ ocaml-mdx test test.md
/Users/jon/.opam/4.07.1/lib/seq: added to search path
/Users/jon/.opam/4.07.1/lib/re: added to search path
/Users/jon/.opam/4.07.1/lib/re/re.cma: loaded
/Users/jon/.opam/4.07.1/lib/bytes: added to search path
/Users/jon/.opam/4.07.1/lib/uchar: added to search path
/Users/jon/.opam/4.07.1/lib/uutf: added to search path
/Users/jon/.opam/4.07.1/lib/uutf/uutf.cma: loaded
/Users/jon/.opam/4.07.1/lib/tyxml/functor: added to search path
/Users/jon/.opam/4.07.1/lib/tyxml/functor/tyxml_f.cma: loaded
/Users/jon/.opam/4.07.1/lib/ocaml/compiler-libs: added to search path
/Users/jon/.opam/4.07.1/lib/tyxml/top: added to search path
/Users/jon/.opam/4.07.1/lib/tyxml/top/tyxml_top.cma: loaded
Exception:
Typetexp.Error
 ({Location.loc_start =
    {Lexing.pos_fname = ""; pos_lnum = 2; pos_bol = 1; pos_cnum = 15};
   loc_end =
    {Lexing.pos_fname = ""; pos_lnum = 2; pos_bol = 1; pos_cnum = 27};
   loc_ghost = false},
 <abstr>, Typetexp.Unbound_module (Longident.Lident "Tyxml")).
/Users/jon/.opam/4.07.1/lib/tyxml: added to search path
/Users/jon/.opam/4.07.1/lib/tyxml/tyxml.cma: loaded
```

Julow:
@jonludlam I'm investigating this error but it does not seem to be fatal, `Tyxml` is loaded and execution can continue.

I suppose it might be a bug with the packaging of Tyxml. Utop also reports an exception (less verbose) and `Tyxml_top` is not available:

```
$ utop -require tyxml
Exception: <abstr>.
utop # Tyxml_top.eval_string;;
Error: Reference to undefined global `Tyxml_top'
```

I found that `(deps (package abc))` does not work on external packages. I'll fix that.

Julow:
I pushed a commit to depend on all installable targets in the project instead of specified packages. So it's no longer needed to update the rule when dependencies change and external packages will work transparently.
This as the inconvenient of building everything in the project. What do you think @samoht ?

samoht:
I think depending on everything is too coarse grain : if you are working in a large duniverse repo I guess you don't really want/need to compile every chapter. I expect the rule to be auto-generated anyway (using `dune rule`) so it will be kept up-to-date automatically whenever you modify the markdown file.

Julow:
I agree it's not great. I reverted that commit.
The problem with `(deps (package abc))` is that it does not work with external dependencies and with libraries inside packages (eg. `package.lib`). I'm looking for a better way to specify deps.

samoht:
@Julow can you rebase this please?

would be also nice to confirm with @NathanReb  if this works with duniverse too.

Julow:
I rebased on `master` and cleaned the history.

I still have a problem with external dependencies and rule generation. Only local packages are allowed in `(package abc)` dependencies for aliases.
I did not find a way in dune to allow "library" dependencies for aliases yet.

Julow:
Requiring external/private libraries is too hard to implement currently. This may requires a new feature into dune, see https://github.com/ocaml/dune/issues/2470.
I moved the label and rule generation part in https://github.com/realworldocaml/mdx/pull/149, it should be useful for https://github.com/realworldocaml/book/pull/3100.

NathanReb:
Can this be closed (and eventually re-opened later) then?

Julow:
Yes this can be closed for now.

PR #132, opened 2019-06-26
Added unset-* label

clecat:
Added a unset-* label, allowing to remove an environment variable in a shell block.

It works the same way as set-* expect that it does not need a value:
>  '''sh unset-HOME
>    $ echo $HOME
>
> '''

It's original purpose was to allow shell script using Dune to run normally:
Dune sets a variable INSIDE_DUNE to 1 when running, which implies some modified behaviours.

So in the case of mdx being used by a dune rule, and itself launching dune in an sh block, the variable is defined before running the nested dune.

This is why, after talking about it with @samoht , I added INSIDE_DUNE to the blacklist by default.

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -150,6 +150,7 @@ let labels = [
   `Label "non-deterministic", [`None; `Some "command"; `Some "output"];
   `Label "version"          , [`Any];
   `Prefix "set-"            , [`Any];
+  `Prefix "unset-"          , [`None];
 ]
 
 let pp_value ppf = function
@@ -217,8 +218,8 @@ let get_labels t label =
 let get_prefixed_labels t prefix =
   List.fold_left (fun acc (k, v) ->
       if String.equal (String.with_range ~len:(String.length prefix) k) prefix then match v with
-        | None   -> assert false
-        | Some (e, s) -> (e, (String.with_range ~first:(String.length prefix) k, s)) ::acc
+        | None   -> (String.with_range ~first:(String.length prefix) k, None)::acc
+        | Some (e, s) -> (String.with_range ~first:(String.length prefix) k, Some (e, s))::acc
       else acc
     ) [] t.labels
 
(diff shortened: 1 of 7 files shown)


lib/block.ml
```diff
@@ -141,15 +143,16 @@ let pp ?syntax ppf b =
...
-  `Label "dir"              , [`Any];
-  `Label "source-tree"      , [`Any];
-  `Label "file"             , [`Any];
-  `Label "part"             , [`Any];
-  `Label "env"              , [`Any];
-  `Label "skip"             , [`None];
-  `Label "non-deterministic", [`None; `Some "command"; `Some "output"];
-  `Label "version"          , [`Any];
-  `Prefix "set-"            , [`Any];
+  {s = "dir"              ; t = `Label ; vs = [`Any]};
```
samoht: why did you change that type?

lib/block.ml
```diff
@@ -272,13 +273,23 @@ let environment t = match get_label t "env" with
   | Some (Some (`Eq, s)) -> s
   | Some (Some _) -> Fmt.failwith "invalid `env` label value"
 
-let variables t =
+(* ([ `Eq | `Ge | `Gt | `Le | `Lt | `Neq ] * string) option * string *)
```
samoht: you can remove that comment

lib/migrate_ast.ml
```diff
@@ -32,7 +32,7 @@ module Parsetree_ = Parsetree
 
 include (
   Migrate_parsetree.Ast_406 :
-  module type of struct
+    module type of struct
```
samoht: why did this file changes? can you clean-up the history and remove the merge commit?

samoht:
Looks good with a few minor remarks.

Could you check that this doesn't break existing mdx users, for instance merlin or rwo?

lib/block.ml
clecat: It felt clearer and cleaner, however the tags could be more explicit

lib/block.ml
clecat: Indeed, my bad

lib/migrate_ast.ml
clecat: Indeed, I was late with master, that's why, i'll do the history rework

lib/block.ml
samoht: Can you revert this change (and make an other PR if you really like the new record, even if I am not very fond of it..)  that'll make it easier to review.

clecat: Okay, no problem

samoht:
Looks great, thanks!

PR #136, opened 2019-07-10
Removed exit preventing promotion in some cases

clecat:
In the case of a promotion from .md to .ml, if the original markdown file does not contain any errors, no promotions seemed to be done in the .ml files:
Ex:
Markdown file (README.ml):
```
```ocaml file=file.ml
let t = "Hello world";
```
OCaml file (file.ml):
```ocaml
let t = "";
```
The markdown is correct, but not the ocaml file. However, the .corrected file is never created by mdx.
This is due to a call to exit in lib/misc.ml at the line 61 if the content of the .md file is the same as what we expected.
Removing this call solves the problem, allowing mdx to end properly and fully finishing  it's job.


samoht:
Can you also add a test for this? Thanks!

clecat:
I changed the existing test so it contains this situation as well

samoht:
I don't understand why the test is not failing (or why it is working) :-) I guess the promoted ml file should change too, right?

clecat:
With the ancient test, the md file was wrong, which avoided the exit call. With the new test, the md file is good, so we are sure that we are not biased by it's correction. It's dumb but it made the difference before, as the exit stopped the execution

clecat:
After so tests, it appears that the tests currently have a problem were they are not raising some cases of errors (when a `.corrected` file is expected but never produced), this is why this problem can only be saw when running ocaml-mdx by hand. I think we should merge this PR and I'll do as fast as possible a new PR correcting this problem

samoht:
Can you revert the change in the example, see that the CI is failing and update the `.expected` file instead? This is to make sure the CI is now working as expected.

clecat:
So, I pushed a commit with a test that should now break, but was not breaking before. I'll push the correction once the CI shows the problem

clecat:
I purposely did not put --force-output as it avoids the if branch calling the `exit 0`

samoht:
Perfect, the test case is very clear and it all makes sense now.

PR #137, opened 2019-07-10
Tests may ignore an error if no correction is generated when one was wanted

clecat:
Currently, if a test wants mdx to generate a `.corrected` file but mdx does not generate the file, `dune runtest` does not raise any error.
This is because we use `diff?` which doesn't need both files to exist.
One solution for now is to use force-output in the tests expecting a correction.
I also updated force-output so it also forces the corrections of .ml files in the case of an to-ml direction.

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -195,9 +195,9 @@ type file = { first: Mdx_top.Part.file; current: Mdx_top.Part.file }
 
 let files: (string, file) Hashtbl.t = Hashtbl.create 8
 
-let has_changed { first; current } =
+let has_changed ~force_output { first; current } =
   let contents = Mdx_top.Part.contents current in
-  if contents = Mdx_top.Part.contents first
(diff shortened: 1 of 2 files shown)


samoht:
LGTM 

Did you file an issue to upstream in ocaml/dune ?

Not sure how to properly test this and make sure this is working correcly now.

clecat:
I did an [issue](https://github.com/ocaml/dune/issues/2396) asking for another diff command but it was rejected, we will keep on with --force-output for now and I will try to do a clean workaround using rules and targets.

Actually, we cannot add any tests, the one existing are already enough, the bug was just ignoring some broken cases, there is no way (I dont see any at least) to check that except trusting.

The rules and targets however would make it clearer

clecat:
I pushed a broken to check if the CI works properly, will be rebased after that

clecat:
Looks like the CI is correctly failing (never expect me to say that again :laughing: ), removing the commit now

PR #139, opened 2019-07-10
Correcting bug preventing promotion of empty ml files

clecat:
When running `ocaml-mdx` with the direction `to-ml`, if a targeted ml file is empty, no corrections would be generated for this file.
This is due to the file parts parsing ignoring empty strings, therefore generating an empty list of parts if the file is totally empty.
My correcting forces the creation of at least one empty part.

The two commits corresponds to:
- Adding a test which should work but is currently broken. (using `--force-output` we can see that the generated correction is empty).
- Correcting the problem.


samoht:
That looks great, thanks!

Issue #138, opened 2019-07-10
ocaml-mdx silently depends on ocaml-mdx-test


NathanReb:
There seems to be an intricate relationship between the `ocaml-mdx` and the `ocaml-mdx-test` binaries.

The former seems to just be some proxy that re-executes the latter with its own command line arguments.

I'm not familiar enough with `mdx` to understand why it works like that but I'd be curious to know.

The pattern is currently pretty fragile and more specifically, it doesn't work very well with dune, or without opam at least.

I'm using a vendored `mdx` so when I tell dune a rule depends on `ocaml-mdx` it will know how to build it. Because `dune`'s not aware that `ocaml-mdx` depends on `ocaml-mdx-test` though, running `ocaml-mdx` just crashes with the following error:
```
ocaml-mdx: internal error, uncaught exception:
           Unix.Unix_error(Unix.ENOENT, "execvp", "../../../install/default/bin/ocaml-mdx-test")
```

While we could probably hack with dune rules to make that dependency explicit to `dune`, I'm not sure that's the right solution here.

Is there any reason why those binaries are split in such a way?

NathanReb:
It could also be fixed by changing the rules generated by `ocaml-mdx rule` to make that dependency clear if we absolutely need those two separated binaries!

avsm:
Interesting -- it would be far easier if these were the same binary indeed

samoht:
The binaries are split because bytecode is needed for the toplevel libraries (as ocamlnat doesn't work realiably yet). The main code doing the parsing needs to be in native otherwise it's too slow (it was like this initially but people noticed it was slow).

So adding an explicit dependency rule makes sense here.

maiste:
Hi,
Currently working on building a `locked` file with `opam monorepo`, I have to face the same issue as @NathanReb did.
```sh
(cd _build/default && ../install/default/bin/ocaml-mdx test -o .mdx/README.md.corrected README.md)
ocaml-mdx: internal error, uncaught exception:
           Unix.Unix_error(Unix.ENOENT, "create_process", "../install/default/bin/ocaml-mdx-test")
```

When pulling the `duniverse` and then building it, the `test` file doesn't seem to be built so the `ocaml-mdx` is not able to find the file at the right place. 

Is there a way to fix (even a temporary one) this issue with any hack?

NathanReb:
Yes this is a known limitation when using the MDX stanza v0.1 in a monorepo project.

[...]

PR #141, opened 2019-07-15
Add a dependency on mdx package in mdx generated rules

NathanReb:
This prevents error due to duplicate binary from happening when vendoring mdx. It's a quick fix rather than a long term solution to #138 since technically we still need to make `ocaml-mdx` depend on `ocaml-mdx-test`.

### DIFF
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -79,7 +79,8 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root options =
       "\
 (alias\n\
 \ (name   %s)\n\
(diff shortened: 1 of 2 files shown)


NathanReb:
I need to try that in a duniverse context to make sure it works properly!

samoht:
even if that doesn't fix duniverse, that looks like a good fix anyway so merging.

NathanReb:
Yeah I just forgot to update but it does work with duniverse or at least it does fix the missing `ocaml-test-mdx` binary error.

It might be worth cutting a `1.4.1` release!

Issue #144, opened 2019-08-01
Exception when running on OCaml 4.08


anuragsoni:
Hi,

I use mdx for one of my projects ([routes](https://github.com/anuragsoni/routes)) and I notice that for travis builds running against 4.08 there is the following error:

```
- (cd _build/default && /home/opam/.opam/4.08/bin/ocaml-mdx test --direction=infer-timestamp README.md)
- ocaml-mdx-test: internal error, uncaught exception:
-                 Failure("incomplete OCaml version")
```
Link to the failed build: https://api.travis-ci.com/v3/job/221454058/log.txt

When i try to run the tests locally under 4.08 i don't have the same exception. The builds for ocal versions 4.05 through 4.07 also succeed (https://api.travis-ci.com/v3/job/221454057/log.txt -> build log for 4.07)

I'm not sure if this is a problem with `mdx` or with my test setup and i'd be happy to help debug more.

Thanks!

gpetiot:
Can you try with #145 ? I didn't expect the OCaml version to not always be of the form x.y.z.

anuragsoni:
I'll see if i can try it with travis soon (i think i can pin your branch there), but trying this locally everything is fine as before. For some reason i only saw this exception in the travis build. When trying to see the OCaml version in the top level on my local switch, i see "4.08.0".

anuragsoni:
Just tried a travis build with your changes, the build passes now: https://api.travis-ci.com/v3/job/221462948/log.txt

XVilka:
So can be closed I guess?

anuragsoni:
Closing as a fix was merged

PR #145, opened 2019-08-01
Fix version comparison with short OCaml version (like 4.08)

gpetiot:
Attempt at fixing #144 (confirmed by the reporter)


samoht:
Merging, but would be nice to have a proper test for this.

PR #149, opened 2019-08-02
Specify required packages

Julow:
A trimmed version of https://github.com/realworldocaml/mdx/pull/130
I do not try to load packages into the toplevel, `#require` are required (hehe). They are only used to generate the list of dependencies in the dune rule.

Dune's `(deps (package abc))` only work for public packages in the current project. Package name is not necessarily library name and it don't work with external packages (installed with opam).

This is enough to work with duniverse.

https://github.com/realworldocaml/mdx/pull/130 needs changes to dune (https://github.com/ocaml/dune/issues/2470) and can be rebased on top of this later.

### DIFF
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -119,14 +119,8 @@ let run () md_file section direction prelude prelude_str root =
     | Block b when active b ->
       let files, dirs, nd, requires = acc in
       let requires =
-        (* Only depend on the base package, eg. 'a.b' become 'a' *)
-        let base_pkg pkg =
-          match String.cut ~sep:"." pkg with
-          | Some (base, _) -> base
-          | None -> pkg
-        in
         Mdx.Block.required_packages b
-        |> List.fold_left (fun s e -> String.Set.add (base_pkg e) s) requires
+        |> List.fold_left (fun s e -> String.Set.add e s) requires
       in
       let nd = nd || match Mdx.Block.mode b with
         | `Non_det _ -> true
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -29,7 +29,7 @@ let prepend_root r b = match r with
   | None   -> b
   | Some r -> Filename.concat r b
 
-let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root options =
+let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires options =
   let ml_files = String.Set.elements ml_files in
   let ml_files = List.map (prepend_root root) ml_files in
   let dirs = match root with
@@ -44,6 +44,10 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root options =
     let f (cpt, acc) _ = cpt + 1, ("y" ^ string_of_int cpt) :: acc in
     List.fold_left f (0, []) ml_files |> snd
   in
+  let requires = String.Set.elements requires in
+  let pp_package_deps fmt name =
+    Fmt.pf fmt "\         (package %s)" name
+  in
   let pp_ml_deps fmt (var_name, ml_file) =
     Fmt.pf fmt "\         (:%s %s)" var_name ml_file
   in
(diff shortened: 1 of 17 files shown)


NathanReb:
What's the status on this?

Julow:
This was extracted from https://github.com/realworldocaml/mdx/pull/150 and to be discussed and merged faster.
It's ready for reviews.

bin/rule.ml
```diff
@@ -57,22 +61,18 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root options =
...
   let deps =
-    let sep1 = if var_names <> [] then "\n" else "" in
-    let sep2 = if dirs <> [] && prelude <> "" then "\n" else "" in
-    let x = Fmt.strf "%a%s%a%s%s"
-        Fmt.(list ~sep:(unit "\n") pp_ml_deps) (List.combine var_names ml_files)
-        sep1
-        Fmt.(list ~sep:(unit "\n") pp_dir_deps) dirs
-        sep2
-        prelude
+    let x =
```
NathanReb: If we start refactoring this part how about a better name than `x` like `depends` maybe?

bin/rule.ml
```diff
@@ -117,7 +117,11 @@ let run () md_file section direction prelude prelude_str root =
   let on_item acc = function
     | Mdx.Section _ | Text _ -> acc
     | Block b when active b ->
-      let files, dirs, nd = acc in
+      let files, dirs, nd, requires = acc in
+      let requires =
+        Mdx.Block.required_packages b
+        |> List.fold_left (fun s e -> String.Set.add e s) requires
```
NathanReb: This is `String.Set.of_list`!

test/dune
```diff
@@ -268,3 +268,12 @@
 (alias
  (name runtest)
  (action (diff dune_rules.inc dune_rules.inc.gen)))
+
+(alias
```
NathanReb: Is this generated or added by hand?

lib/block.ml
```diff
@@ -288,6 +289,15 @@ let unset_variables t =
   in
   List.map f (get_prefixed_labels t "unset-")
 
+let required_packages t =
+  let f = function
+    | `Eq, "" ->
+      Fmt.failwith "invalid `require-package` label value: requires a value"
+    | `Eq, pkg -> pkg
+    | _ -> Fmt.failwith "invalid `env` label value"
```
NathanReb: Shouldn't it be `` invalid `require-package` label value ``?

bin/rule.ml
Julow: This adds to `requires` so it's `of_list` and `union`. I think this is better.

lib/block.ml
Julow: Nice catch

bin/rule.ml
Julow: It's not necessary at all. This part of the code need to be cleaned up.

bin/rule.ml
NathanReb: The fact that it's going to be refactored later doesn't mean we shouldn't properly name variables now. It's rather the opposite actually, making the code easier to understand helps with later refactoring.

I don't mind if you leave it like that as it is in a way unrelated to this PR. I just think it's good practice to improve the code we're modifying.

bin/rule.ml
NathanReb: The fact that it's going to be refactored later doesn't mean we shouldn't properly name variables now. It's rather the opposite actually, making the code easier to understand helps with later refactoring.

I don't mind if you leave it like that as it is in a way unrelated to this PR. I just think it's good practice to improve the code we're modifying.

bin/rule.ml
NathanReb: Do you mean it's better performance-wise or it reads better?

I'm not sure it reads really well, in particular:
```ocaml
let requires_from_block = String.Set.of_list (Mdx.Block.required_packages b) in
let requires = String.Set.union requires requires_from_block in
...
```
would be much clearer than the current where I have to mentally parse the fold left.

Anyway don't bother with that as I think that while rebasing #150 on top of this, I refactored it that way because it introduce different sources for required packages and it made sense to extract intermediate variables there.

bin/rule.ml
Julow: I removed the binding. My second sentence was not an argument but a complain.
I'm working on improving that module.

test/dune
Julow: By hand for now. See https://github.com/realworldocaml/mdx/pull/162

NathanReb:
This looks good to me! @samoht want to have a final look or should I merge?

samoht:
as long as it makes the use-case of RWO simpler (and does not add too much cruft) go for it!

NathanReb:
Cool, let's merge then!

PR #150, opened 2019-08-05
Duniverse mode

NathanReb:
This PR adds a `--duniverse-mode` flag to `mdx rule` to generate rules that works within a dune-only context and therefore with a `duniverse`.

This is done by parsing the `#require` statements within OCaml toplevel code blocks, adding `(package ...)` dependencies to the `mdx test` rule for any package found that way and by setting `findlib`'s config so that it looks in the `_build/install` directory instead.

This is more of an hack rather than a long term solution but it gives us something to iterate on and it solves some issues for [realworldocaml/book](https://github.com/realworldocaml/book).

Let me know what you think!

### DIFF
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -137,6 +137,30 @@ let pp_prelude_str fmt s = Fmt.pf fmt "--prelude-str %S" s
 
 let add_opt e s = match e with None -> s | Some e -> String.Set.add e s
 
+let aggregate_requires ~require_from l =
+  let open Rresult.R.Infix in
+  Mdx.Util.Result.List.fold
+    ~f:(fun acc x -> require_from x >>| Mdx.Library.Set.union acc)
+    ~init:Mdx.Library.Set.empty
+    l
+  >>| Mdx.Library.Set.to_package_set
+
+let requires_from_prelude_str prelude_str_list =
+  let require_from prelude_str =
+    let (_, prelude) = Mdx.Prelude.env_and_file prelude_str in
+    let lines = String.fields ~is_sep:(function '\n' | '\r' -> true | _ -> false) prelude in
+    Mdx.Block.require_from_lines lines
+  in
+  aggregate_requires ~require_from prelude_str_list
+
+let requires_from_prelude prelude_list =
+  let require_from prelude =
+    let (_, prelude) = Mdx.Prelude.env_and_file prelude in
+    let lines = Mdx.Util.File.read_lines prelude in
+    Mdx.Block.require_from_lines lines
+  in
+  aggregate_requires ~require_from prelude_list
+
 let run (`Setup ()) (`File md_file) (`Section section) (`Syntax syntax) (`Direction direction)
   (`Prelude prelude) (`Prelude_str prelude_str) (`Root root)
   (`Duniverse_mode duniverse_mode) =
@@ -183,7 +207,10 @@ let run (`Setup ()) (`File md_file) (`Section section) (`Syntax syntax) (`Direct
   let on_file file_contents items =
     let empty = String.Set.empty in
     let req_res =
-      Mdx.Util.Result.List.fold ~f:on_item ~init:(empty, empty, false, String.Set.empty) items
+      requires_from_prelude prelude >>= fun prelude_requires ->
+      requires_from_prelude_str prelude_str >>= fun prelude_str_requires ->
+      let requires = String.Set.union prelude_requires prelude_str_requires in
+      Mdx.Util.Result.List.fold ~f:on_item ~init:(empty, empty, false, requires) items
     in
     match req_res with
     | Error s ->
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -39,7 +39,7 @@ let pp_action fmt = function
   | `Diff_corrected var -> Fmt.pf fmt "(diff? %%{%s} %%{%s}.corrected)" var var
   | `Run args -> Fmt.pf fmt "(run @[<hov 2>%a@])" Fmt.(list ~sep:sp string) args
 
-let pp_rules ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires
+let pp_rules ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~packages
     fmt options =
   let ml_files = List.map (prepend_root root) (String.Set.elements ml_files) in
   let dirs = match root with
@@ -61,7 +61,7 @@ let pp_rules ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires
   let root = match root with None -> [] | Some r -> ["--root=" ^ r] in
   let deps =
     let packages =
-      List.map (fun p -> `Package p) (String.Set.elements requires)
+      List.map (fun p -> `Package p) (String.Set.elements packages)
     and ml_files =
       List.map2 (fun name p -> `Named (name, p)) var_names ml_files
     and dirs = List.map (fun p -> `Source_tree p) dirs
@@ -137,7 +137,7 @@ let pp_prelude_str fmt s = Fmt.pf fmt "--prelude-str %S" s
 let add_opt e s = match e with None -> s | Some e -> String.Set.add e s
 
 let run (`Setup ()) (`File md_file) (`Section section) (`Syntax syntax) (`Direction direction)
-    (`Prelude prelude) (`Prelude_str prelude_str) (`Root root) =
+    (`Prelude prelude) (`Prelude_str prelude_str) (`Root root) (`Duniverse_mode _) =
   let active =
     let section = match section with
       | None   -> None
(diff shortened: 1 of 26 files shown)


NathanReb:
cc @Julow

lib/library.ml
```diff
@@ -0,0 +1,39 @@
+type t =
```
samoht: missing headers

lib/prelude.mli
```diff
@@ -0,0 +1 @@
+val env_and_file : string -> string option * string
```
samoht: what does this function do?

lib/util.ml
```diff
@@ -0,0 +1,27 @@
+module Result = struct
+  module List = struct
+    open Rresult.R.Infix
```
samoht: if you redefine a few `Result` function you probably don't need to depend on `Rresult` and can inline bind and map directly.

lib/block.mli
```diff
@@ -94,6 +94,9 @@ val set_variables: t -> (string * string) list
 val unset_variables: t -> string list
 (** [unset_variable t] is the list of environment variable to unset *)
 
+val required_libraries: t -> (Library.Set.t, string) Result.result
```
Julow: This should not return a result. Other functions in this module either raise to let the user known or ignore errors.

Julow: This function does not need to use `Library.Set`. It should return an `string list` and let the calling code handle this.

test/require/require.md
```diff
@@ -0,0 +1,17 @@
+# Using local library
+
+```ocaml
+# #require "example_lib";;
```
samoht: I would tend to favour the removal of the `#require` statements as they: 1/ force a dependency towards ocamlfind and 2/ could be guessed properly in most of the cases (see utop or [down](https://github.com/dbuenzli/down)). 

So I think it's better to only use `require=...` in the block metadata to indicate which OCaml library should be added to the scope of the current environment and remove all `#require` completely.

Julow: There is a lot of `#require` in rwo, that's why we added that. In duniverse mode, they are used to specify package deps, otherwise they do nothing.

samoht: are `#require` used elsewhere than in preludes in rwo?

NathanReb: I think `#require` statement should appear in toplevel code blocks so that the reader can eventually copy and paste the content of those blocks into a toplevel and get it to work.

There are indeed quite a few `#require` statement in code blocks in RWO.

Also note that parsing them doesn't necessarily mean that they have to be evaluated, to the contrary, we could strip them from the actual lines passed on to the toplevel and load the libraries there ourselves.

lib/prelude.mli
NathanReb: This function was moved from the `bin/main/test.ml` where it was used to parse passed through `--prelude` or `--prelude-str` to separate the filename/content from the env into which the prelude must be evaluated.

That's where the `Prelude` name comes from. I extracted it here because it is now also used in `bin/rule.ml`. If you have a better suited module in mind, I'll happily move it there.

lib/block.mli
NathanReb: My reasoning was that actually parsing library names here is more future proof and could later be used to load those libraries rather than simply adding a dependency on the package.

I'm not sure whether you're suggesting to move the parsing to `Library.t` values and the building of the set elsewhere or not to do this at all.

In the former case I'm not sure I understand why it would be better but I can move it if you'd like.

NathanReb: Ok, I'll let it throw.

test/require/require.md
samoht: IThere are not so many `#require` statements in RWO so these can be changed if needed:
```
$ git grep "#require" book/**/*.md | grep -v prelude
book/concurrent-programming/README.md:# #require "async"
book/concurrent-programming/README.md:We first load the Async package in the toplevel using `#require`, and then
book/concurrent-programming/README.md:# #require "cohttp.async"
book/foreign-function-interface/README.md:# #require "ctypes.foreign"
book/foreign-function-interface/README.md:# #require "ctypes.top"
book/json/README.md:# #require "yojson"
book/lists-and-patterns/README.md:# #require "core_bench"
book/records/README.md:# #require "re.posix"
book/runtime-memory-layout/README.md:# #require "compiler-libs"
```
Also, the "copy&paste" arguments doesn't really hold as you can't copy&paste the results.

In any case, what I meant is that we *must not* rely on ocamlfind in mdx for anything. This is just craft and will make it harder to remove in the feature if people start to use it.

NathanReb: The issue here is that I don't think we can do without `findlib` atm.

A better and more future proof solution is discussed [here](https://github.com/ocaml/dune/issues/2470) and it's not clear whether it implies relying on `findlib` or not yet.

Anyway this is just here to get working tests in RWO. I'd understand if you don't want it merged because of it's fragile nature, I just wanted to avoid having to backport fixes or new features to a fork/branch. We can keep it separate for now.

samoht: > The issue here is that I don't think we can do without findlib atm.

I am discussing about the API we expose to the user ; I am fine to use findlib internally at the moment, but again, we should not rely on it being present in the near future. Putting it in the block header makes it easier to switch to a different system (directly provided by dune) in the near future.

I am fine to merge that PR to ease maintenance, but before this I'd like to make sure we are not explicitely dependending on findlib/ocamlfind or at least these parts have to be clearly identified and separated so that they can be removed easily in a subsequent PR (which should happen before the next release).

NathanReb:
Small fix that we need here: not to add a package dependency to `compiler-libs`!

Julow:
(Rebased on master and merged https://github.com/realworldocaml/mdx/pull/155 and https://github.com/realworldocaml/mdx/pull/156)

bin/rule.ml
```diff
@@ -120,7 +105,7 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~packages ~duniverse_
 \           (diff? %%{x} %%{x}.corrected))))\n"
       name
       md_file
-      conf_dep
+      ""
```
samoht: probably better to just remove that argument

Julow:
I cleaned the history and rebased on master.

NathanReb:
Ok this PR is now ready for a final pass of review, @samoht do you want to have a look at it?

NathanReb:
Okay wait, it's not, we messed the rebase and removed the extra `package mdx` dependency, I'll fix that!

NathanReb:
Ok so I further cleaned that up.

It now depends on #184 but I expect we merge it soon.
I fixed:
- the compatibility issues with older versions
- the lack of tests
- the forgotten extra `(package mdx)` dependency that must be added in duniverse mode

Once #184 is merged I think we can give this a last review round and merge. If you want to see it in action, you can look at: https://github.com/realworldocaml/book/pull/3185

NathanReb:
Just rebased this.

Given that the next scheduled release is `2.0.0` (I'm happy to backport bug fixes if needed) I think it's okay to merge this even if it's not yet perfect so that RWO depends on master.

A quick last round of review just to make sure I haven't broken anything by accident should be enough, I'm willing to change this incrementaly in subsequent PRs!

samoht:
My comments still hold: please add a licence header to new files and a bit of documentation to the new mlis. Also, if you depend on `rresult` for just one function it's probably better to just redefine that function instead of bringing a new dependency. Other than that, that seems ok.

NathanReb:
I just added the license headers, documentation to `Prelude` and `Library` and removed the dependency on `Rresult`!

lib/block.ml
```diff
@@ -300,6 +301,37 @@ let required_packages t =
...
+let require_from_line line =
+  let open Util.Result.Infix in
+  let re = Re.compile require_re in
+  match Re.exec_opt re line with
+  | None -> Ok Library.Set.empty
+  | Some group ->
+    let matched = Re.Group.get group 1 in
+    let libs_str = String.cuts ~sep:"," matched in
+    Util.Result.List.map ~f:Library.from_string libs_str >>| fun libs ->
+    List.fold_left (fun acc l -> Library.Set.add l acc) Library.Set.empty libs
```
Julow:
```suggestion
    Library.Set.of_list libs
```

lib/block.ml
```diff
@@ -300,6 +301,37 @@ let required_packages t =
...
+let require_from_lines lines =
+  let open Util.Result.Infix in
+  Util.Result.List.map ~f:require_from_line lines >>| fun libs ->
+  List.fold_left Library.Set.union Library.Set.empty libs
+
+let required_libraries = function
+  | { value = Toplevel _; contents; _} -> require_from_lines contents
+  | { value = (Raw | OCaml | Error _ | Cram _); _ } -> Ok Library.Set.empty
+
+let required_packages t =
```
Julow: This function is not used.
It's not useful to keep it to not break the API because its type change and it doesn't do the same thing.

lib/library.ml
```diff
@@ -0,0 +1,62 @@
...
+  Fmt.(pair ~sep:(cst "; ") string (option ~none:(cst "None") string)) fmt (base_name, sub_lib);
+  Fmt.string fmt " }"
+
+let from_string s =
+  let invalid () = Error (Printf.sprintf "Invalid library name: %S" s) in
+  match Astring.String.cuts ~sep:"." s with
+  | [""] | [""; _] | [_; ""] -> invalid ()
+  | [base_name] -> Ok {base_name; sub_lib = None}
+  | [base_name; sl] -> Ok {base_name; sub_lib = Some sl}
+  | _ -> invalid ()
```
Julow: Do you know if this is defined ? Why having `.` in sub lib an error ?

NathanReb: I don't think you can define a `a.b.c` library, at least not in dune but I'll check!

NathanReb: Also might be worth merging anyway since we're only keeping this as a temporary solution.

NathanReb: Dune doesn't seem to forbid it although I'm not sure what that means for the main library file and module name. I'll remove that error for now.

NathanReb:
I think this should be ready to merge!

Issue #148, opened 2019-08-01
Split a submodule into parts with mdx in to-ml mode


NathanReb:
I recently ran into an issue with mdx. I have an `.md` file containing several ocaml code snippets pointing to different parts of an `.ml` file. I'm using `mdx` with `--direction to-ml` and for some reasons, it gets the first right but doesn't write the other ones to the `.ml` file.

My first guess was that the first part starts at the file's toplevel but introduces the declaration of a submodule and the subsequent parts all are within that submodule. I thought this could be the reason for the issue because `[@@@...]` attributes are structure_items of a different part of the AST but I doubt `mdx` actually relies on that.

You can take a look at the `.md` file [here](https://github.com/NathanReb/book/blame/build-and-test-with-duniverse/book/first-class-modules/README.md#L666) and the current target file [here](https://github.com/NathanReb/book/blob/build-and-test-with-duniverse/examples/code/fcm/query_handler_loader/query_handler_core.ml). Running `dune runtest` there result in the `diff` action on the `.ml` file suggesting to remove anything from `[@@@part "2"]` (included) to the end of the file. I'm not sure exactly why this is happening. I'll to reproduce that in a simpler example if I find the time to!

PR #155, opened 2019-08-06
Fix parts parsing

Julow:
Fixes https://github.com/realworldocaml/mdx/issues/148.

If a part delimiter occur inside a nested structure or signature, it will be ignore:

```ocaml
[@@@part "1"]

module A =
struct

[@@@part "2"]

  type t = A

[@@@part "3"]

end
```

The parts 2 and 3 won't be parsed and everything will be the body of the part 1, which will be replaced by just:

```ocaml
[@@@part "1"]

module A =
struct
```

This PR fixes this by parsing the part delimiters line by line using `Re` instead of using OCaml's parser. This has the advantage of not failing and not depending on an unstable library.

### DIFF
--- a/test/sync_to_md.md.expected
+++ b/test/sync_to_md.md.expected
@@ -37,6 +37,7 @@ let () =
   ()
 ;;
 
+
 let x = 34
 let f = 42.3
 let s = "toto"
@@ -46,6 +47,7 @@ let () =
   print_int x;
   print_float f
 ;;
+
 
 let () =
   print_string s
--- a/test/sync_to_md.md.expected
+++ b/test/sync_to_md.md.expected
@@ -37,7 +37,6 @@ let () =
   ()
 ;;
 
-
 let x = 34
 let f = 42.3
 let s = "toto"
@@ -47,7 +46,6 @@ let () =
   print_int x;
   print_float f
 ;;
-
 
 let () =
   print_string s
--- a/test/sync_to_ml.ml.expected
+++ b/test/sync_to_ml.ml.expected
@@ -19,6 +19,7 @@ let f = "hello world!"
 
 [@@@part "42"] ;;
 
+
 let () =
   f x print_int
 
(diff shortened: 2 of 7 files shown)


samoht:
I've restarted the failing CI job.

test/sync_to_md.md.expected
```diff
@@ -37,6 +37,7 @@ let () =
   ()
 ;;
 
+
```
samoht: why is there extra new lines in the expected files?

samoht:
I think I'm ok with the approach -- it allows to put part in invalid ast fragments too, which is a nice feature (and allow to extract parts of structures as you demonstrated it).

I am not sure to understand where do the extra newlines come from.

Also, what happens if `[@@@part  ...]` doesn't start at the beginning of the line? I got the impression that your patch is able to detect the part but then it removes these spaces in the expected file? I guess this will break ocamlformat-ed files, if you have a part deep into a structure.

test/sync_to_md.md.expected
Julow: Fixed.
Part parsing no longer remove empty lines at the end of parts, this was a bug in the code that update md files.

Julow:
The regex that matches `[@@@part ...]` allow trailing whitespaces but they are not printed back. I didn't change the printing code so this is unchanged, except in the case the file did not parse and the whole file is printed as-is, including part separators.

Julow:
I added a commit (a21f0fa) to preserve the indentation of `[@@@part ...]`.
Can be reverted. What do you think?

test/sync_to_ml.ml.expected
```diff
@@ -19,5 +19,6 @@ let f = "hello world!"
 
 [@@@part "42"] ;;
 
+
```
samoht: a new extra newline ? :-)

Julow: Nice catch. Changed the code to trim sooner.

Julow:
Rebased. @samoht Do you think we can merge this PR ?

samoht:
LGTM

Issue #154, opened 2019-08-06
Mdx ignore new parts when syncing in to-ml direction


Julow:
I have a `README.md` defining a part to sync:

````md
Blabla:

```ocaml file=test.ml,part=1
let f x = x
```
````

and a `test.ml`:

```ocaml
[@@@part "1"]

let f x = x
```

If I add a new part in the README, eg.

````md

Part 2:

```ocaml file=test.ml,part=2
let g f x = f x
```
````

running mdx with `--direction=to-ml` won't update the `test.ml` file.

NathanReb:
I believe this is related to #148 !

PR #156, opened 2019-08-06
Fix parts syncing

Julow:
Fixes https://github.com/realworldocaml/mdx/issues/154.

If a part does not exist it is appended to the file.
A bug remain: Parts are not inserted in the order they appear in the markdown file.


samoht:
Thanks! Could you also add a test to exercise that feature?

Julow:
Did.

samoht:
Thanks!

PR #161, opened 2019-09-04
Modify 'pp' to respect the 'skip' label

### LINKED ISSUE #159: `pp` subcommand doesn't respect the `skip` label
The [Irmin tutorials](https://github.com/tarides/irmin.io/pull/7/files#diff-6ca3c6684ec06eee8be2c61772e7f672R5) are currently using the `pp` subcommand to verify that each one compiles. We currently have one block which fails CI, and so it would be nice to use the `skip` label to avoid outputting it to the `.ml` file; however, the `pp` subcommand simply ignores the `skip` label.

Perhaps the `pp` command was never intended to be used for a test workflow such as this, but until https://github.com/realworldocaml/mdx/issues/158 is fixed it is not possible to use the `test` subcommand to achieve the right result.

CC. @Julow

CraigFe:
This fixes #159.

Perhaps there are less hacky ways to do this? I was trying to minimise the diff.

### DIFF
--- a/bin/pp.ml
+++ b/bin/pp.ml
@@ -34,16 +34,18 @@ let run () file section =
         | Mdx.Section _ | Text _ -> ()
         | Block b ->
           let b = Mdx.Block.eval b in
-          Log.debug (fun l -> l "pp: %a" Mdx.Block.dump b);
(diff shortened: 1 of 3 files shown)


samoht:
Looks good

PR #162, opened 2019-09-09
Generate test rules

Julow:
To better test rule generation and avoid eventually outdated rules.
Rules that were using a `.expected` file are kept in `test/dune`, mdx cannot generate them.

The fist commit fixes the printing of `--prelude-str="code"` which is parsed as 2 words, `--prelude-str=` and `code` by dune.


test/dune.inc
```diff
@@ -0,0 +1,189 @@
+(alias
+ (name   runtest)
+ (deps   (:x pp.md)
+         (package mdx))
+ (action (progn
+           (run ocaml-mdx test --direction=infer-timestamp %{x})
```
samoht: are you sure we want `--direction=infer-timestamp`: this doesn't work very well with version controlled files.

Julow: This is the default value. Tests were already run like this.

Julow: These tests probably don't use synchronization so it would not make sense to set a direction.
Should we change the default ?

NathanReb: Hmm indeed so that means that `ocaml-mdx rule` translate the default into an explicit argument?

That seems like bad behaviour. It means we have to update the default in two places if we want to change it.

Anyway that's unrelated to this PR, I'll add an issue!

NathanReb:
Can you please rebase and fix the conflict and then we can merge!

test/dune.inc
NathanReb: Hm you have a good point, the test are expected to succeed anyway. I guess it would be safer to use a direction yes!

Julow:
Rebased and waiting for the CI.

test/dune.inc
NathanReb: Changing mdx's default would be sensible as well, especially if we're not too confident in `infer-timestamp`'s behaviour but that's a different matter and it would be a breaking change.

samoht: I think that's a good idea to change the default and probably nobody (apart from rwo) is using the promotion mechanism so we are safe.

PR #163, opened 2019-09-09
Cleanup rule printing

Julow:
Refactor printing code to use `Format` features, making it more robust and easier to change.

Rebased on top of https://github.com/realworldocaml/mdx/pull/162

### DIFF
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -29,6 +29,12 @@ let prepend_root r b = match r with
   | None   -> b
   | Some r -> Filename.concat r b
 
+let pp_dep fmt = function
+  | `Named (name, path) -> Fmt.pf fmt "(:%s %s)" name path
+  | `Source_tree path -> Fmt.pf fmt "(source_tree %s)" path
+  | `Package p -> Fmt.pf fmt "(package %s)" p
+  | `Path path -> Fmt.pf fmt "%s" path
+
 let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires options =
   let ml_files = String.Set.elements ml_files in
   let ml_files = List.map (prepend_root root) ml_files in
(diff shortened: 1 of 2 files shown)


Julow:
Rebased on master and ready for reviews.

bin/rule.ml
```diff
@@ -142,7 +144,8 @@ let run () md_file section direction prelude prelude_str root =
     let options =
       List.map (Fmt.to_to_string pp_prelude) prelude @
       List.map (Fmt.to_to_string pp_prelude_str) prelude_str @
-      [Fmt.to_to_string pp_direction direction]
+      [Fmt.to_to_string pp_direction direction] @
+      (match section with Some s -> [Fmt.(to_to_string pp_section) s] | None -> [])
```
samoht: why is this needed?

samoht:
It's a good idea to simplify that part of the code -- I just don't understand why the patch adding the section is part of the changesets. I guess this is fixing a different issue?

bin/rule.ml
Julow: That's unrelated.
The `-s` option is used by `rule` but not passed to `test`, that will not work if ignored sections need extra deps.
Other options are correctly passed (`prelude`, `direction`, `root`...).

Should I move this commit to an other PR?

samoht: yes please, let's keep that PR focused on cleaning up the code without any bug fixes.

Julow: I moved that commit to https://github.com/realworldocaml/mdx/pull/176

CraigFe:
I feel that we could do with direct tests of the `ocaml-mdx rule` command (i.e. generated rules which are _not_  for propping up other tests). This would be enabled by #178.

Julow:
I agree testing that way is not great. We still benefit from generating test rules but we need an other way to test every features of `ocaml-mdx rule`.

bin/rule.ml
```diff
@@ -29,9 +29,18 @@ let prepend_root r b = match r with
   | None   -> b
   | Some r -> Filename.concat r b
 
+let pp_dep fmt = function
+  | `Named (name, path) -> Fmt.pf fmt "(:%s %s)" name path
```
NathanReb: Is there any particular reason to use polymorphic variants here? I think I'd prefer we define types for dependencies and actions here.

Julow: That's because these values are created a few lines below and directly passed to this function. So it's not worth the overhead of defining the type.
What do you think ?

NathanReb: I think that if we extract proper types, move those to the library we can then unit test how they are printed rather than using end to end testing for that.

We can probably do that separately though. Let's keep it like this for now, it's already an improvement!

Issue #146, opened 2019-08-01
mdx sometimes add trailing whitespaces


NathanReb:
Mdx appear to sometime add trailing whitespaces to toplevel evaluations. See an example of this behaviour [here](https://github.com/realworldocaml/book/pull/3100/commits/17abdfa156d7d91fb7d80dea9091dc0ed61efecf#diff-0ddf0bd6914c9962843f0b23ed6009caR710).

I haven't had time to investigate why this happened but I'm guessing it could be fixed by systematically stripping trailing whitespaces from toplevel evaluations as there's no value in keeping them around even if the toplevel did return them.

samoht:
@clecat can you have a look at this? I think we changed that code recently.

clecat:
I'll give a look at it

clecat:
We should close this issue now

PR #166, opened 2019-09-11
Trim the end of output

clecat:
Some cram tests or toplevel executions generates trailing whitespaces in the output, conflicting with code editors deleting them.

This small pull request simply trims the end of the output strings.

It solves [this issue](https://github.com/realworldocaml/mdx/issues/146) and [this one](https://github.com/realworldocaml/book/issues/3164) too

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -17,6 +17,7 @@
 open Mdx
 open Compat
 open Result
+open Astring
 
 let src = Logs.Src.create "cram.test"
 module Log = (val Logs.src_log src : Logs.LOG)
@@ -24,10 +25,10 @@ module Log = (val Logs.src_log src : Logs.LOG)
 let (/) = Filename.concat
 
 let prelude_env_and_file f =
-  match Astring.String.cut ~sep:":" f with
+  match String.cut ~sep:":" f with
   | None        -> None, f
   | Some (e, f) ->
-    if Astring.String.exists ((=) ' ') e
+    if String.exists ((=) ' ') e
     then None  , f
     else Some e, f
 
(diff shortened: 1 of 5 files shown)


bin/test/main.ml
```diff
@@ -60,6 +60,19 @@ let ansi_color_strip str =
...
+  let rec f i =
+    if i > l
+    then
+      i
+    else
+      match s.[l - i - 1] with
+      | ' ' | '\t' | '\011'..'\013' -> f (succ i)
+      | _ -> i
+  in
+  String.sub s 0 (l - (f 0))
```
Julow: This project depends on Astring so you can implement this like:
```ocaml
let trim_end s =
  String.drop ~rev:true ~sat:(function ' ' | '\t' | '\011'..'\013' -> true | _ -> false) s
```

NathanReb: You can probably go even further and make it:
```ocaml
let trim_end s = String.drop ~rev:true ~sat:Ascii.Char.is_white s
```

Or Ascii.Char.is_blank, not sure which one suits this better!

bin/test/main.ml
```diff
@@ -118,7 +131,7 @@ let run_cram_tests ?syntax t ?root ppf temp_file pad tests =
       let n = run_test ?root blacklist temp_file test in
       let lines = read_lines temp_file in
       let output =
-        let output = List.map (fun x -> `Output x) lines in
+        let output = List.map (fun x -> `Output (trim_end x)) lines in
```
NathanReb: Might be worth extracting a `output_from_lines` function so we don't have to remember to update all those call sites in the future!

PR #167, opened 2019-09-11
Remove extra package deps on mdx

NathanReb:
This reverts #141. Package dependencies cannot be resolved to opam packages by dune as of today. This so-called fix introduces a bug and the package dependency on mdx should only be added when vendoring mdx itself, eg in a duniverse.

I'll re-enable that in the `--duniverse-mode`!

### DIFF
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -78,8 +78,7 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires options =
       "\
 (alias\n\
 \ (name   %s)\n\
-\ (deps   (:x %s)\n\
-\         (package mdx)%s)\n\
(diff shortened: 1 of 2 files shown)


NathanReb:
This what you would get if using current master without vendoring mdx:
```
$ dune runtest
File "infer-timestamp/dune.inc", line 1, characters 0-262:
1 | (alias
2 |  (name   runtest)
3 |  (deps   (:x README.md)
4 |          (package mdx)
5 |          (:y0 example.ml))
6 |  (action (progn
7 |            (run ocaml-mdx test --direction=infer-timestamp %{x})
8 |            (diff? %{y0} %{y0}.corrected)
9 |            (diff? %{x} %{x}.corrected))))
Error: No rule found for alias .mdx-files
```

Issue #169, opened 2019-09-11
Remove the infer-timestamp direction, at least temporarily


NathanReb:
The `--direction=infer-timestamp` is not working very well. In particular it can't be used in dune because dune copies all the files to the `_build` directory first and therefore messes with the timestamps.

While I like the idea of having such an option, I'm quite struggling to find a good workflow for it outside of dune and its promotion mechanism.

I'd suggest we simply remove it for now until we have that figured out.

Note that if anyone's actually using this option and has a good workflow to integrate it outside of dune I'd love to hear about it. If that were the case an alternative would be to remove it from `ocaml-mdx rule` at least so that it only produces options that actually work fine with dune.

NathanReb:
This has been fixed!

PR #171, opened 2019-09-16
Remove infer-timestamp direction

Julow:
Implement https://github.com/realworldocaml/mdx/issues/169

Instead of setting `to-ml` or `to-md` as default, `direction` is now an option. An exception is raised if this option is needed but not provided.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,10 @@
 - Remove trailing whitespaces at the end of toplevel or bash evaluation lines
   (#166, @clecat)
 
+#### Removed
+
+- Remove the `infer-timestamp` direction (#171 @Julow)
+
 ### 1.4.0 (2019-06-11)
 
 - Add `--force-output` option to force generation of diff file (#118 @clecat)
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -254,13 +254,10 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
   in
   let direction =
     match direction with
-    | `To_md -> `To_md
-    | `To_ml -> `To_ml
-    | `Infer_timestamp ->
-       let md_file_mtime = (Unix.stat md_file).st_mtime in
-       let ml_file_mtime = (Unix.stat ml_file).st_mtime in
-       if ml_file_mtime < md_file_mtime then `To_ml
-       else `To_md
+    | Some `To_md -> `To_md
+    | Some `To_ml -> `To_ml
+    | None ->
+        Fmt.failwith "--direction must be set to update this file."
   in
   match direction with
   | `To_md ->
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -252,15 +252,6 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
     | None   -> dir / ml_file
     | Some r -> r / dir / ml_file
   in
-  let direction =
-    match direction with
-    | Some `To_md -> `To_md
-    | Some `To_ml -> `To_ml
-    | None        ->
-        Fmt.failwith "This block must be synchronised with %s, \
-                      the --direction option must be provided."
-          ml_file
-  in
   match direction with
   | `To_md ->
      update_block_with_file ppf block ml_file (Block.part block)
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -256,8 +256,10 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
     match direction with
     | Some `To_md -> `To_md
     | Some `To_ml -> `To_ml
-    | None ->
-        Fmt.failwith "--direction must be set to update this file."
+    | None        ->
+        Fmt.failwith "This block must be synchronised with %s, \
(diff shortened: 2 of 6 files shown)


bin/test/main.ml
```diff
@@ -250,13 +250,10 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
...
-    | `To_ml -> `To_ml
-    | `Infer_timestamp ->
-       let md_file_mtime = (Unix.stat md_file).st_mtime in
-       let ml_file_mtime = (Unix.stat ml_file).st_mtime in
-       if ml_file_mtime < md_file_mtime then `To_ml
-       else `To_md
+    | Some `To_md -> `To_md
+    | Some `To_ml -> `To_ml
+    | None ->
+        Fmt.failwith "--direction must be set to update this file."
```
NathanReb: I like that it makes `--direction` optional!

Do you think we could include the location of the block in the markdown file that requires the option to be set? The filename at least would be great. I also think the error message could be slightly more explicit, what would you think about somethng like: "This block must be synchronised with <ml_file>, you must set a promotion direction for <md_file>. See --direction documentation." ?

NathanReb:
Oh, I'd also like to start keeping the changelog up-to-date from now on. Would you mind adding an entry there with your name and the PR number, under a removed section?

NathanReb:
I'd like to get @samoht approval for this first as well!

bin/test/main.ml
Julow: Improving errors need a bit of work, I'll submit an other PR for easier reviews. For the tests too.
The message can be improved.

NathanReb: Ok, let's leave the location out for now. Could you, as part of this PR, improve the error message a little bit and still add a test just to make sure the error is properly triggered and ensure we don't inadvertently break it in the future?

I believe this can be done quite easily as both file names are available in the scope at this point.

CHANGES.md
```diff
@@ -1,3 +1,6 @@
+### Master
+- Remove the `infer-timestamp` direction (#171 @Julow)
```
Julow: I added a CHANGES line.
Should I mark that it's a breaking change ? And, how ?
@NathanReb what do you mean by a "removed" section ?

bin/test/main.ml
Julow: I changed the message.
Location/testing in https://github.com/realworldocaml/mdx/pull/172. Not done in this PR because the output would not looks good (uncaught exception) and should be discussed separately.

CHANGES.md
NathanReb: I recenlty updated the changelog, can you rebase on top of that and put that under a `#### Removed` section.

From there I'd say it's pretty clear that removed features are a breaking change.

Julow: Done

bin/test/main.ml
NathanReb: I'm not sure I understand how that prevents you from adding a test to that PR, especially since the first commit in #172 is pretty much what I'd expect from this PR.

Julow: Currently the output is an uncaught exception message from OCaml's runtime, which is not nice. But more importantly, adding this test requires to add a mechanism to run it that may require discussion.
Would you prefer if I reorder the PRs ? #172 first and this one rebased on it ?

samoht:
If I understand properly, the option is now mandatory in the presence of `file=...` metadata? Why not pick a default option instead?

Julow:
That was to avoid choosing the wrong default. Users may not notice this option exists and 50% of them may be surprised of the default since most other options are passed directly as label on each blocks.
Which value should be the default?

NathanReb:
If we must pick a default I'd vote for `to-md` as I think that's the prefered choice in most cases. It makes more sense to edit ocaml code in `.ml` and `.mli` files.

As a matter of fact I'm wondering if there is a good reason for wanting to do it the other way around. Maybe now is the right time to decide whether this option should exist at all.

Julow:
Alternatively, this option could be set on each block, through a label.

NathanReb:
> Alternatively, this option could be set on each block, through a label.

I think it make sense to provide that option if we want to allow promotion both ways but that's definitely in the scope for a separate PR.

samoht:
I initially thought that `to-ml` would make more sense as it allow people to fix typos when reading the `.md` file only but it's apparently not the preferred one in rwo. In any case we should make the rwo worflow easier by default so `to-md` as the default seems fine to me.

Julow:
I changed the default to `to-md`.
I liked that it had no defaults as it's not clear which value make more sense as default and that this option is not working the same way as other options in Mdx (cli instead of per-block).
What do you think about adding a `direction` label ? Alternatively, to have `to-file` and `from-file` instead of the label `file` to explicitly choose the direction ?

samoht:
I like the `to-file` and `from-file` labels

Julow:
I'll implement that.
In the meantime, I think we should merge this. The `direction` option will have to stay a bit as a transition, so we cannot replace it entirely now.

PR #172, opened 2019-09-18
Improve error reporting of ocaml-mdx test

Julow:
Rebased on top of https://github.com/realworldocaml/mdx/pull/171

- Improve the printing of errors
  Expected errors were shown as uncaught exceptions
- Show the line number and file name for errors occurring while testing a block
- Add tests for error messages

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -258,6 +258,8 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
   | `To_ml ->
      update_file_with_block ppf block ml_file (Block.part block)
 
+exception Test_block_failure of Block.t * string
+
 let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     (`Not_verbose not_verbose) (`Syntax syntax) (`Silent silent)
     (`Verbose_findlib verbose_findlib) (`Prelude prelude)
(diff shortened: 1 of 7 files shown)


bin/test/main.ml
```diff
@@ -373,13 +381,21 @@ let run_exn ()
...
-    prelude_str file section root direction
+    prelude_str file section root direction force_output
+  : int
   =
     try
     run_exn () non_deterministic not_verbose syntax silent verbose_findlib
-      prelude prelude_str file section root direction
-    with Failure f -> prerr_endline f; exit 1
- 
+      prelude prelude_str file section root direction force_output
```
Julow: This function was partially applied, exception were never catched.

test/dune
```diff
@@ -156,3 +156,21 @@
...
+
+; Test failures
+
+(rule
+ (targets failures.outputs)
+ (deps    (package mdx) failures/empty.ml)
+ (action
+  (with-outputs-to %{targets}
+   (progn
+    (system "%{bin:ocaml-mdx} test %{dep:failures/direction_not_set.md} || echo")
```
samoht: I'm not sure `||` is very portable on windows

NathanReb: It's not indeed but I don't know any work around for that. I think a `run!` action in dune would help there. We should probably open an issue about that. I even wonder if I haven't opened one already as I'm using the same trick to test PPXs' errors!

Julow: I think it will work fine.

Julow:
Rebased.

test/dune
```diff
@@ -163,3 +163,21 @@
...
+
+; Test failures
+
+(rule
+ (targets failures.outputs)
+ (deps    (package mdx) failures/empty.ml)
+ (action
+  (with-outputs-to %{targets}
+   (progn
+    (system "%{bin:ocaml-mdx} test \"%{dep:failures/direction_not_set.md}\" || echo")
```
Julow: I'm adding quotes to avoid problems with spaces. But some characters will still be interpreted by bash (`$`, backquote, `\`, `"`) and cmd (`%`, `!`).
I don't known any syntax that would fix that and be parsed by both bash and cmd.

samoht: if that's just for the tests, maybe it's not necessary to add `"` as well.

bin/test/main.ml
```diff
@@ -254,20 +256,21 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
...
-    | `To_ml -> `To_ml
-    | `Infer_timestamp ->
-       let md_file_mtime = (Unix.stat md_file).st_mtime in
-       let ml_file_mtime = (Unix.stat ml_file).st_mtime in
-       if ml_file_mtime < md_file_mtime then `To_ml
-       else `To_md
+    | Some `To_md -> `To_md
+    | Some `To_ml -> `To_ml
+    | None        ->
+        Fmt.failwith "This block must be synchronised with %s, \
```
samoht: Can you pick `To_md` as the default option to ease rwo?

test/dune
Julow: Yes. Looks like `dune` only use relative paths so this should not be a problem.
Reverted.
Still, the ideal would be to have an equivalent of `(run ..)` that don't check exit code.

NathanReb: Or that checks that it fails!

NathanReb:
Can you rebase now that #171 has been merged please?

Julow:
Rebased and cleaned.

test/dune
CraigFe: If #178 is merged, we could do something similar to the solution in Alcotest for this: https://github.com/mirage/alcotest/blob/master/test/e2e/expect_failure.ml. Agreed that it would be nice for the `dune` `run` DSL to support this; I've been meaning to make an issue about this for a while.

Julow: Yes that would be great. I'll PR your PR.

bin/test/main.ml
```diff
@@ -371,8 +380,15 @@ let run setup non_deterministic not_verbose syntax silent verbose_findlib
...
     run_exn setup non_deterministic not_verbose syntax silent verbose_findlib
       prelude prelude_str file section root direction force_output
-    with Failure f -> (prerr_endline f; exit 1)
- 
+    with
+    | Failure f ->
+      prerr_endline f;
+      1
+    | Test_block_failure (block, msg) ->
+      Fmt.epr "Error in block at line %d in %s:@\n%s\n"
```
NathanReb: I think the error message could be slightly improved here by:
- Using `code block` instead of block, I think it sounds a bit vague
- Adding the kind of block, OCaml, OCaml toplevel or shell
- printing locations the other way around, file first, line second

bin/test/main.ml
```diff
@@ -258,6 +260,8 @@ let update_file_or_block ?root ppf md_file ml_file block direction =
   | `To_ml ->
      update_file_with_block ppf block ml_file (Block.part block)
 
+exception Test_block_failure of Block.t * string
```
NathanReb: Again, I think I'd prefer if we used results instead of exceptions there but given this is already an improvement we can refactor that later.

NathanReb:
Can you add a changelog entry as well?

bin/test/main.ml
Julow: Changed it. What do you think?

bin/test/main.ml
```diff
@@ -306,73 +364,41 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
 
+let report_error_in_block block msg =
+  let kind =
+    match block.Block.value with
+    | Raw | Error _ -> ""
+    | OCaml -> "OCaml "
+    | Cram _ -> "cram "
+    | Toplevel _ -> "toplevel "
+  in
+  Fmt.epr "Error in the %sblock in %s at line %d:@]\n%s\n"
```
NathanReb: Could you make that "%scode block" ? I think it's the right terminology for those, block is too generic!

Julow: This will be "OCaml block" or "toplevel block" already. This may be too much words. Also "block" is not generic at all.

Julow: Agreed "code block" is better.

Julow:
Rebased. Also cleaned the history a bit.

Issue #157, opened 2019-08-07
Support accepting multiple files as CLI arguments


CraigFe:
The CLI subcommands (`output`, `pp`, `rule`, `test`) currently all accept only a single file as input; it might be nice if they could accept more than one at a time.

This would be helpful when verifying a _set_ of markdown files, as in the [Irmin tutorials](https://github.com/CraigFe/irmin.io/tree/mdx-verification/data/tutorial). Currently, I think this requires an unsightly `bash` action to do it in a single Dune action:

```
  (action (with-stdout-to %{targets}
            (bash "echo %{tutorials} | xargs --max-args=1 ocaml-mdx pp")))
```
Whereas it would be nice to get away with simply:
```
  (action (with-stdout-to %{targets}
            (ocaml-mdx pp %{tutorials})))
```
What do you think?

samoht:
that sounds like a good idea!

NathanReb:
That would also be good for the dune integration I think!

NathanReb:
It's actually better to invoke mdx once per file in dune so nevermind my previous comment.

gpetiot:
Does the recent `mdx` dune stanza makes it easier or is this issue still relevant?

CraigFe:
I haven't ported the Irmin tutorials to use the `mdx` stanza yet: it's waiting for the release of the HTML-syntax for MDX attributes  (also currently blocking https://github.com/mirage/irmin/pull/991), and is a non-trivial change due to needing the workaround described in https://github.com/realworldocaml/mdx/issues/158.

If the `pp` command is going to be deprecated in favour of the `mdx` stanza, this issue can probably be closed. Having said that, I still find `pp` a better literate programming workflow than the stanza + explicit file/part labels, so would only switch grudgingly :wink:

shonfeder:
Also worth noting that supporting multiple files is a nice feature for using this in tests outside of dune.

NathanReb:
> Also worth noting that supporting multiple files is a nice feature for using this in tests outside of dune.

I think at the moment the dune stanza is the recommended way to use MDX and we're more headed towards dropping the mdx binary than the other way around.

> If the pp command is going to be deprecated in favour of the mdx stanza, this issue can probably be closed. Having said that, I still find pp a better literate programming workflow than the stanza + explicit file/part labels, so would only switch grudgingly ￼

[...]

PR #173, opened 2019-09-18
Support multiple files as CLI arguments

CraigFe:
This fixes https://github.com/realworldocaml/mdx/issues/157, but is currently untested.

### DIFF
--- a/bin/pp.ml
+++ b/bin/pp.ml
@@ -17,35 +17,42 @@
 let src = Logs.Src.create "cram.pp"
 module Log = (val Logs.src_log src : Logs.LOG)
 
-let run (`Setup ()) (`File file) (`Section section) =
-  let t = Mdx.parse_file Normal file in
-  let t = match section with
-    | None   -> t
-    | Some s ->
-      let re = Re.Perl.compile_pat s in
-      match Mdx.filter_section re t with
-      | None   -> []
-      | Some t -> t
+let filter_by_section section lines =
+  match section with
+  | None -> lines
+  | Some s ->
+    let re = Re.Perl.compile_pat s in
+    match Mdx.filter_section re lines with
+    | None -> []
+    | Some lines -> lines
+
+let run (`Setup ()) (`Files files) (`Section section) =
+  let t =
+    List.map (fun p ->
+        (p, Mdx.parse_file Normal p |> (filter_by_section section))
+      ) files
   in
   match t with
   | [] -> 1
   | _  ->
-    List.iter (function
-        | Mdx.Section _ | Text _ -> ()
-        | Block b ->
-          let b = Mdx.Block.eval b in
-          if not (Mdx.Block.skip b) then (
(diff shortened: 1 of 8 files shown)


bin/pp.ml
```diff
@@ -17,42 +17,49 @@
...
-            match b.value with
-            | Toplevel _ -> Fmt.pr "%a\n" pp_lines contents
-            | OCaml      ->
-              Fmt.pr "%a\n%a\n"
-                Mdx.Block.pp_line_directive (file, b.line)
-                pp_lines contents
-            | _          -> ()
-          )
-      ) t;
+    t |> List.iter (fun (file, blocks) ->
```
NathanReb: I don't necessarily mind the change of style but could you please leave this (and the other similar change in `bin/test/main.ml`) out of this PR to ease the review?

CraigFe: I've split the PR into two commits:
 - https://github.com/realworldocaml/mdx/pull/173/commits/607342e02fe59b91b32703f9fd437b78364a0d46, which contains the new feature;
 - https://github.com/realworldocaml/mdx/pull/173/commits/ca595dc7e5bd40cec4756c1ad822518c2699fe6c, which changes the style. I've taken the opportunity to be more suggestive with the style change; I find this far easier to understand, particularly in the case of nested `List.iter`s.

Happy to split into another PR altogether, but perhaps a commit split is enough?

NathanReb: A separate commit is fine! Thanks!

CraigFe:
I will delay writing tests for this PR until #178 is either merged or closed, since it would save some time.

bin/test/main.ml
```diff
@@ -24,6 +24,8 @@ module Log = (val Logs.src_log src : Logs.LOG)
 
 let (/) = Filename.concat
 
+let (>>-) x f = (|>) x (List.iter f)
```
Julow: Why not defining this like this `List.iter f x` ?
Also, I'm not fan of defining this as an operator as you had to define it twice in 2 modules.
I agree having the function as the last argument is slightly better, `abc >>- (fun x ->` could be written `abc |> List.iter (fun x ->`. What do you think?

CraigFe: The definition I've given reads in the way I think about it, so I think it's easier :slightly_smiling_face:

I think in files where there are lots of occurrences of this pattern it makes sense to define an infix operator for it, but I don't particularly mind `|> List.iter` as an alternative.  I also think it's fairly natural to need to redefine this sort of helper in multiple files, but again I don't particularly mind if you think otherwise.

CraigFe: https://github.com/realworldocaml/mdx/pull/173/commits/865fe919e93547601fe88adc01ee546aa056d538 rewrites the `>>-` definitions as `List.iter f x` for simplicity :slightly_smiling_face:

CraigFe:
Rebased over #174.

CraigFe:
Closing as stale.

PR #174, opened 2019-09-18
Refactor handling of CLI parameters

CraigFe:
This adds polymorphic variant wrappers to all of the various CLI parameters in an effort to prevent them from being re-ordered along the way.

More verbose, but hopefully safer w.r.t. future changes.

### DIFF
--- a/bin/output.ml
+++ b/bin/output.ml
@@ -84,7 +84,7 @@ let pp_block ppf (b:Mdx.Block.t) =
   Fmt.pf ppf "<div class=\"highlight\"><pre%a><code%a>%t</code></pre></div>"
     pp_attrs () pp_lang () pp_code
 
-let run () file output =
+let run (`Setup ()) (`File file) (`Output output) =
   let t = Mdx.parse_file Normal file in
   match t with
   | [] -> 1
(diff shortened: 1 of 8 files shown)


bin/test/main.ml
```diff
@@ -375,12 +376,10 @@ let run_exn ()
   Hashtbl.iter (write_parts ~force_output) files;
   0
 
-let run ()
-    non_deterministic not_verbose syntax silent verbose_findlib prelude
-    prelude_str file section root direction
-  =
+let run setup non_deterministic not_verbose syntax silent verbose_findlib
+    prelude prelude_str file section root direction =
```
Julow: Here the `force_output` argument is missing, so the call to `run_exn` below is partially applied and exceptions are never catch.
Could you add it and add a type annotation to the function (`int`) to prevent it in the future?

bin/output.ml
```diff
@@ -118,7 +118,9 @@ open Cmdliner
 let output =
   let doc = "Write output to $(b,FILE) instead of stdout.  If $(b,FILE) is -, \
              output will go to stdout," in
-  Arg.(value & opt (some string) None & info ["o";"output"] ~doc ~docv:"FILE")
+
```
Julow: Why empty line?

bin/test/main.ml
CraigFe: Good catch; done.

samoht:
Looks good, thanks!

PR #176, opened 2019-09-19
Rule: Print the -s option

Julow:
`ocaml-mdx rule` accept the `-s` option but it does not print it back.
That means it will generate rules that test the whole file instead of the section the rule was generated for.

As a test, I generate the rule for `section.md`.


CraigFe:
This is very related to #177; perhaps we need a better long-term solution.

CraigFe:
Also, as mentioned in #163, I feel that testing `ocaml-mdx rule` only via tests that do other things is not great: I find overloading tests makes them hard to reason about. #178 would be a step towards writing explicit tests of this behaviour.

Julow:
> This is very related to #177; perhaps we need a better long-term solution.

Do you suggest adding a mechanism to pass option directly to `ocaml-mdx test` ? This is what is already done with `prelude`, `prelude-str`, `direction`. What is the problem with that interface ?
The `-s` option have to be handled by both `rule` and `test`.

CraigFe:
I know `-s` and `--syntax` are different flags; it just so happens that `rule` currently supports neither of them when `test` does. 

I think the current state of _some_ options to `rule` being used to customise the rule generation and _some_ options being implicitly passed through to `ocaml-mdx test` in the generated rule is confusing and doesn't scale well (what would happen if we added an `-s` flag to `ocaml-mdx rule` for instance?). In #177 I suggest having a single option for `ocaml-mdx rule` whose sole responsibility is to pass options to the internal `ocaml-mdx test` call.

CraigFe:
I agree that not all of the options can be directly passed through to the `ocaml-mdx test` case: in some cases you really want to change the way `rule` behaves. However, I think:
 - if the option _doesn't_ change the structure of the generated rule then it should be exposed via a passthrough option because (a) we don't want to have to extend `rule` to account for changes to `test` and (b) it's easier to reason about / makes for cleaner code
 - if the options _does_ change the rule generation, then this should be properly documented in the Cmdliner.Arg, rather then borrowing options from the `Cli` module which have documentation intended for use in `ocaml-mdx test` instead.

test/dune.inc
```diff
@@ -171,4 +171,11 @@
  (action (progn
            (run ocaml-mdx test --direction=infer-timestamp %{x})
 
+           (diff? %{x} %{x}.corrected))))
+(alias
+ (name   runtest)
+ (deps   (:x section.md))
+ (action (progn
+           (run ocaml-mdx test --direction=infer-timestamp -s "Testing" %{x})
+
```
NathanReb: We really need to fix that empty line bug!

test/dune.inc
```diff
@@ -171,4 +171,11 @@
  (action (progn
            (run ocaml-mdx test --direction=infer-timestamp %{x})
 
+           (diff? %{x} %{x}.corrected))))
+(alias
+ (name   runtest)
+ (deps   (:x section.md))
+ (action (progn
+           (run ocaml-mdx test --direction=infer-timestamp -s "Testing" %{x})
```
NathanReb: There's no such direction anymore !

bin/rule.ml
```diff
@@ -97,21 +97,23 @@ let pp_direction fmt = function
...
-  let section = match section with
-    | None   -> None
-    | Some p -> Some (Re.Perl.compile_pat p)
-  in
-  let active b = match section, Mdx.Block.section b with
-    | None   , _      -> true
-    | Some re, None   -> Re.execp re ""
-    | Some re, Some s -> Re.execp re (snd s)
+  let active =
+    let active re b =
```
NathanReb: I'm not sure it reads any better that way, I think it'd be clearer as a single pattern matching:
```ocaml
let active b =
  match section with
  | None -> true
  | Some p ->
   let re = Re.Perl.compile_pat p in
   let block_section = Util.Option.get_value ~default:"" (Mdx.Block.section b) in
   Re.execp re block_section
```

You can add the `Option.get_value` to the `Util` module, we'll probably need it again in the future!

Also I'm not sure the `-s` option behaviour when there's no section is really intuitive to users. It should probably be documented. That's unrelated to this PR but if you feel like adding that in a separate commit it'd be great!

bin/rule.ml
```diff
@@ -143,7 +145,8 @@ let run (`Setup ()) (`File md_file) (`Section section) (`Direction direction)
     let options =
       List.map (Fmt.to_to_string pp_prelude) prelude @
       List.map (Fmt.to_to_string pp_prelude_str) prelude_str @
-      [Fmt.to_to_string pp_direction direction]
+      [Fmt.to_to_string pp_direction direction] @
+      (match section with Some s -> [Fmt.(to_to_string pp_section) s] | None -> [])
```
NathanReb: Adding the `Option.get_value` (or whatever bette name you can come up with) will also be useful here!

bin/rule.ml
NathanReb: Correct me if I'm wrong but I feel like this contains a bug fix so it should probably sit in its own commit!

bin/rule.ml
```diff
@@ -97,21 +97,23 @@ let pp_direction fmt = function
   | `To_md -> Fmt.pf fmt "--direction=to-md"
   | `To_ml -> Fmt.pf fmt "--direction=to-ml"
 
+let pp_section fmt s = Fmt.pf fmt "-s %S" s
```
NathanReb: I think it'd be better to use the long form here: `--section=%S`

Julow: No. This will result in `--section="abc"` which will be parsed as 2 atoms by dune ("--section=" and "abc").

NathanReb: My point was about the output rule being more readable so we can go for `--section` and `"abc"` can't we?

test/dune.inc
Julow: Fixed on master.

test/dune.inc
Julow: Was not rebased on that. Now it is.

bin/rule.ml
Julow: I changed to write it like syntax. What do you think?

bin/rule.ml
Julow: I think I didn't change any behavior. What do you mean?

This code also unpack the tuple and I don't want to come up with a "get_value_map" function when a pattern matching is as easy to use.

bin/rule.ml
Julow: Yes that's better.

bin/rule.ml
Julow: I originally changed the code to avoid shadowing `section`. Just changed so it's similar to how it was.

NathanReb: Ok nevermind I misread the change, thought we change when the re was compiled but it's not the case!

I do prefer the old version so we'll move this within the `let active = ...` to prevent the shadowing of the `section` argument, as discussed offline!

bin/rule.ml
NathanReb: Looks good thanks!

NathanReb:
Please update the changelog and I'll merge!

bin/rule.ml
```diff
@@ -96,6 +96,10 @@ let options_of_syntax = function
   | Some Mdx.Cram -> [ "--syntax=cram" ]
   | None -> []
 
+let options_of_section = function
+  | Some s -> [ Fmt.str "--section %S" s ]
```
NathanReb: There's a compatibility issue with `Fmt.str` which is only available since `fmt.0.8.7` and `fmt.0.8.7` is for OCaml `>= 4.05`.

For this next release we'd like not to break compatibility with older OCaml versions. Could you replace this `Fmt.str` use with something else please?

Julow: Build for 4.02 is green. Waiting for all the builds and merging.

PR #177, opened 2019-09-19
Add support for '--syntax' flag in 'ocaml-mdx rule'

CraigFe:
`ocaml-mdx rule` accepts some options solely in order to pass them to the `ocaml-mdx test` rule that it generates (`--direction`, `--prelude` etc.), but it doesn't allow for passing the `--syntax` flag. This adds the `--syntax` flag.

This is very similar to #176, making me think this logic probably warrants a refactor to make it easier to understand; a pass-through option like `--test-options="--syntax=cram"` might work.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 ### unreleased
 
+#### Added
+
+- Add `--syntax` option to `rule` subcommand to allow generating rules for cram
+  tests (#177, @craigfe)
+
 #### Fixed
 
 - Remove trailing whitespaces at the end of toplevel or bash evaluation lines
--- a/bin/rule.ml
+++ b/bin/rule.ml
@@ -91,6 +91,11 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires options =
   pp "runtest" [];
   if nd then pp "runtest-all" ["--non-deterministic"]
 
+let options_of_syntax = function
+  | Some Mdx.Normal -> [ "--syntax=normal" ]
+  | Some Mdx.Cram -> [ "--syntax=cram" ]
+  | None -> []
+
 let pp_direction fmt = function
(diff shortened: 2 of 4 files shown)


CraigFe:
This is tested by #178. I can add a test directly here, but if the tests are just about to be refactored then it doesn't seem worthwhile to me.

CraigFe:
Rebased over #171.

CraigFe:
Rebased over #163.

Julow:
I prefer having `rule` take the same options as `test`. So there is no different syntaxes between options that rule use and the ones it does not. In the future, `rule` may need to known the syntax.

I think we can merge this :)

What do you think @NathanReb, @samoht ?

CraigFe:
> I prefer having `rule` take the same options as `test`. 

FWIW, I don't like this very much for two reasons:
 - The doc-strings for those options explain the use in `test` but not in `test`.
 - There are existing collisions that make this impossible: `--color`, `--quiet`, `--verbose` interact directly with the `rule` command and cannot be passed to `test` AFAICS. I think collisions like this are inevitable.

CraigFe:
The test exists in #178; I was trying to avoid writing it here only to immediately delete it. Would you like it here too?

NathanReb:
I really prefer when the feature/fix comes with its tests, especially since there's not much to discuss here but there might be in #178 (haven't reviewed it yet).

I totally trust that #178 is an improvement but given that it doesn't change `mdx`'s behaviour and that I'd like to release a new version soon, adding a test here would allow me to merge this with a free mind and be able to delay the review of #178 a bit (just a bit I promise)!

In short: if it's too much of a hassle don't bother, if it's easy enough that'd be great indeed!

NathanReb:
Can you also add a changelog entry please?

CraigFe:
Added test and changelog entry.

PR #178, opened 2019-09-19
Clean up end-to-end testing

CraigFe:
This cleans up the end-to-end test plumbing:
 - it categorises the tests according to the feature-under-test, rather than managing all of the tests in the same `dune` file;

 - it splits up test files that were being used to test multiple features (in particular, `section.md` has been gradually accumulating more responsibilities... see [#161](https://github.com/realworldocaml/mdx/pull/161/files#diff-2f436029b34085924e44afe01dd27275) and [#167](https://github.com/realworldocaml/mdx/pull/176/files#diff-c3c5bb93b454c1c2eefcabcb46b0e659));

 - it adds a new `gen_dune_rules.ml` executable responsible for generating some of the more common patterns. This picks up any test-specific CLI options from `<test>.opt` files;

 - it adds `runtest` triggers for `.md` files that were previously being ignored (e.g. `environment_variable.md`). The introduction of the auto-generation script prevents this problem from re-occurring in the future.

This is based on #177 (unmerged) for convenience, and also tests the new functionality in that PR via the [`test_fixpoint/cram`](https://github.com/realworldocaml/mdx/pull/178/files#diff-bb60c2e01d9c07aaa2007299870ea595) test case.


test/dune
```diff
@@ -1,165 +0,0 @@
...
-    (run ocaml-mdx rule %{dep:lwt.md})
-    (run ocaml-mdx rule %{dep:non-det.md})
-    (run ocaml-mdx rule %{dep:code.md})
-    (run ocaml-mdx rule %{dep:set_environment_variable.md})
-    (run ocaml-mdx rule %{dep:unset_environment_variable.md})
-    (run ocaml-mdx rule --prelude-str "#require \"lwt\"" --prelude-str "toto:let x = \"42\"" %{dep:prelude.md})
-    (run ocaml-mdx rule --prelude prelude.ml %{dep:prelude_file.md})
-    (run ocaml-mdx rule --direction=to-ml %{dep:dune_rules.md})
-    (run ocaml-mdx rule %{dep:require/require-package.md})
-   ))))
```
Julow: That part is gone. It generates the rules for a few tests and include them.
Is it intended?

Julow: Sorry I missed that you do this from `gen_dune_rules.ml`.

CraigFe:
Rebased over #171.

CraigFe:
Rebased over #163.

test/bin/gen_dune_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+    | _ -> assert false
+  in
+  let rec search_dot i =
+    if i < 0 || is_dir_sep name i then invalid_arg "Filename.chop_extension"
+    else if name.[i] = '.' then String.sub name 0 i
+    else search_dot (i - 1) in
+  search_dot (String.length name - 1)
+
+let options_of_file file =
+  let options_file = chop_extension file ^ ".opts" in
```
Julow: This is inconsistent with `.expected` files where the extension is not removed.
Why is this needed?

test/bin/gen_dune_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+  if not (Sys.file_exists expected_file) then
+    failwith (
+      Fmt.strf "No %s file found when generating an expect rule" expected_file
+    );
+  let options = options_of_file file in
+  Fmt.pr
+    {|
+(alias
+  (name runtest)
+  (deps (:x %s) (:y %s.expected) (package mdx))
```
Julow: For `y`, you can use `expected_file` instead.

test/bin/gen_dune_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+  (action
+    (progn
+      (run ocaml-mdx test --force-output %a %s)
```
Julow: You should use `%%{x}` instead of passing `file` again

test/bin/gen_dune_rules.ml
CraigFe: I think it's not inconsistent: the `.md.expected` files _are_ markdown files, so the extension is kept to indicate their format. There are also tests with both `.ml.expected` and `.md.expected` files (see `test_sync/sync_to_ml`), so it's necessary to keep those two distinct.

I don't keep the `.md`/`.t` extensions on the `.opts` files because those are not Markdown/Craml files.

test/bin/gen_dune_rules.ml
```diff
@@ -0,0 +1,125 @@
...
+  |> List.iter (match rule_type with
+      | `Test_fixpoint -> test_fixpoint_rule
+      | `Test_expect -> test_expect_rule
+    )
+
+open Cmdliner
+
+let cmds =
+  Term.[
+    const (rule_gen `Test_fixpoint) $ const (), info "test_fixpoint";
```
Julow: Here we could pass the function instead of a variant matching its name. What do you think?

Julow:
I add the tests from https://github.com/realworldocaml/mdx/pull/172 in https://github.com/CraigFe/mdx/pull/1

test/bin/gen_dune_rules.ml
Julow: That makes sense.

test/bin/gen_dune_rules.ml
CraigFe: I see why you'd want to pass the function here: it gets rid of an indirection. However, this was a deliberate indirection: I don't like mingling logic with Cmdliner specifications wherever possible. The current pattern has the advantage that `rule_gen`'s behaviour is explicit in its type and is independent of Cmdliner.

If you don't mind too much, I'd like to keep it how it is.

Julow: Yes that makes sense, I agree. I initially thought it was point less because `rule_gen` is referenced here, so why not the `test_*_rule` function that are kind of entry points.

CraigFe:
Rebased.

CraigFe:
CI failing for unrelated reasons; can't re-trigger it myself.

Julow:
Restarted it and it's passing.

test/bin/gen_dune_rules.ml
```diff
@@ -0,0 +1,151 @@
...
+  let rec search_dot i =
+    if i < 0 || is_dir_sep name i then invalid_arg "Filename.chop_extension"
+    else if name.[i] = '.' then String.sub name 0 i
+    else search_dot (i - 1) in
+  search_dot (String.length name - 1)
+
+let options_of_file file =
+  let options_file = chop_extension file ^ ".opts" in
+  if not (Sys.file_exists options_file) then
+    [""]
```
Julow: Why not `[]` ? It's only passed to `pp_options` and it should render the same.

test/bin/gen_dune_rules.ml
```diff
@@ -0,0 +1,151 @@
...
+let pp_options =
+  Fmt.(list ~sep:(const string " ") string)
+
+(** Tests that the output of 'ocaml-mdx test [options] <file>' is equal to the
+    contents of '<file>'. Uses the rule generator build into ocaml-mdx. *)
+let test_fixpoint_rule file =
+  let options = options_of_file file in
+  let command =
+    let args = file::options in
+    Fmt.strf "ocaml-mdx rule %a" pp_options args
```
Julow: This is calling `ocaml-mdx` I have on my system if I `dune clean`. You should add `(package mdx)` on every rules calling this program.

CraigFe: Thanks :slightly_smiling_face: It's only necessary in the `test_fixpoint/dune` case, since other rule generation patterns don't rely on `ocaml-mdx rule`. Have added it there.

NathanReb: I'm not super convinced by this one. I feel like we should either test `ocaml-mdx test` or `ocaml-mdx rule` but not both at the same time.

If I want to test `ocaml-mdx rule`, I will compare its output with a `.expected` file. 

None of the tests in `test_fixpoint` actually require calling `ocaml-mdx rule` and this could just be another case of `test_expect` from what I can see. I might be wrong so let me know if it is in fact required but if it's not, I think we could detect whether the `.expected` file exists and compare to input on if it doesn't!

NathanReb: Another advantage of my suggested approach is that we'd get rid of the annoying `execvp` errors!

CraigFe: Regarding testing ocaml-mdx rule separately: I agree completely, but think this is something we should do _next_. The test directories are structured so that a `test/rule_expect` will fit in there nicely. The current state of affairs is how it was being done before, and I kept it this way to keep the diff smaller (not introducing a bunch of additional tests in the middle of a refactor).

Regarding making `test_fixpoint` a conditional behaviour of `test_expect` in the case where the expected file doesn't exist: I'd rather not do this. I'd rather the behaviour was kept explicit, and that each `test_foo` be kept as simple as possible. The code can be simplified in another way by testing `ocaml-mdx rule` generation separately and then not depending on `ocaml-mdx rule` to generate tests of other properties of the binary.

NathanReb: I agree about keeping the behaviour explicit!

My main point was that I think you could just make the current `test_fixpoint_rule` very similar to `test_expect_rule`. I agree it changes how things were done but I don't think it will change what's tested too much and we could add proper `ocaml-mdx rule` tests in a separate PR.

Let's merge it like this, I'll improve it later on, you already had to rebase this quite a few times by now!

PR #184, opened 2019-10-02
Format dune rules output by rule command

### LINKED ISSUE #183: Format output of ocaml-mdx rule
The output of `ocaml-mdx rule` isn't correctly formatted.

Rather than trying to maintain that in `mdx` we should probably find a way to use `dune format-dune-file` for that, although I don't think it can read from the standard input unfortunately.

We could write to a temporary file though!

NathanReb:
This PRs passes the generated rules through `dune format-dune-file` to output formatted rules.

Fixes #183


test/dune.inc
```diff
@@ -1,164 +1,218 @@
...
- (action (progn
-           (run ocaml-mdx test --direction=to-md --syntax=cram %{x})
-           (diff? %{x} %{x}.corrected))))
-(alias
- (name   runtest)
- (deps   (:x section.md))
- (action (progn
-           (run ocaml-mdx test --direction=to-md --section "Testing" %{x})
-           (diff? %{x} %{x}.corrected))))
+ (name runtest)
```
samoht: that's really ugly :p

NathanReb: Haha, I don't have strong opinion on the beauty of either output but at least it's consistent with how dune formats its own files and I think that's the most important part.

It also shields us against eventually breaking formatting since we're not responsible for it anymore!

bin/rule.ml
```diff
@@ -88,8 +89,34 @@ let print_rule ~nd ~prelude ~md_file ~ml_files ~dirs ~root ~requires options =
...
-  if nd then pp "runtest-all" ["--non-deterministic"]
+  pp fmt "runtest" [];
+  if nd then
+    pp fmt "runtest-all" ["--non-deterministic"]
+
+let print_format_dune_rules pp_rules =
+  let open Unix in
+  let read_pipe, write_pipe = pipe () in
+  set_close_on_exec read_pipe;
+  set_close_on_exec write_pipe;
```
Julow: `~⁠cloexec:true` on `pipe` is not working ?

bin/rule.ml
```diff
@@ -156,7 +183,10 @@ let run (`Setup ()) (`File md_file) (`Section section) (`Syntax syntax) (`Direct
       options_of_syntax syntax @
       options_of_section section
     in
-    print_rule ~md_file ~prelude ~nd ~ml_files ~dirs ~root ~requires options;
+    let pp_rules fmt () =
+      pp_rules ~md_file ~prelude ~nd ~ml_files ~dirs ~root ~requires fmt options;
```
Julow:
```suggestion
      pp_rules ~md_file ~prelude ~nd ~ml_files ~dirs ~root ~requires fmt options
```

bin/rule.ml
NathanReb: It's not available on OCaml `< 4.05` :'(

bin/rule.ml
NathanReb: good catch!

bin/rule.ml
Julow: :(

NathanReb:
I just rebased, let's wait for the CI to go green and merge!

PR #185, opened 2019-10-02
Run tests through opam in CI

NathanReb:
This will prevent the need for debugging failures at release time in the future!

### DIFF
--- a/mdx.opam
+++ b/mdx.opam
@@ -10,7 +10,7 @@ doc:          "https://realworldocaml.github.io/mdx/"
 build: [
  ["dune" "subst"] {pinned}
  ["dune" "build" "-p" name "-j" jobs]
- ["dune" "runtest" "-p" name] {with-test}
+ ["make" "test"] {with-test}
 ]
 
 depends: [
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - TO_TEST=tests OPAMBUILDTEST=true OCAML_VERSION=4.07.1
   - TO_TEST=tests OPAMBUILDTEST=true OCAML_VERSION=4.08.1
   - TO_TEST=tests OPAMBUILDTEST=true OCAML_VERSION=4.09.0
+  - TO_TEST=install OPAMBUILDTEST=true OCAML_VERSION=4.09.0
 before_install:
   # Download and use opam2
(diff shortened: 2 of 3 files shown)


samoht:
What problem does this fix?

NathanReb:
The tests couldn't be run through opam and we didn't realize because we never ran them through opam in the CI!

NathanReb:
To clarify, this PR does two things:
- Add a CI build that basically runs `opam install -t mdx`, see the last line in the env matrix of the `.travis-ci.yml`
- Fixes the opam file so that when opam runs our test they don't fail

NathanReb:
Is it good to merge?

mdx.opam
```diff
@@ -10,7 +10,7 @@ doc:          "https://realworldocaml.github.io/mdx/"
 build: [
  ["dune" "subst"] {pinned}
  ["dune" "build" "-p" name "-j" jobs]
- ["dune" "runtest" "-p" name] {with-test}
+ ["make" "test"] {with-test}
```
samoht: I don't understand this change -- why is this needed?

NathanReb: `dune runtest -p mdx` doesn't work because some of our tests rely on a different local package which doesn't get built because of the `-t` option.

`dune runtest` alone doesn't work because `ocaml-mdx` depends on `ocaml-mdx-test` and we can't specify dependencies on executables in dune.

Finally running `dune build @install @runtest` can fail because of a race condition ie some tests are run before `ocaml-mdx-test` is built.

Since we were using `make test` to run the test in the CI I switched to that. I also made it clear in the makefile that `dune runtest` can only be reliably run after `dune build`.

NathanReb: I tried other solutions with dune but couldn't get any of them working and always ended up with a race condition, I think using the Makefile to wrap dune commands the way we want it is okay but if there's any issue with that we can look for an alternative!

samoht: Can you report this upstream (and point to that PR to see it's annoying for us)? It would be nice if we didn't have to add hacks to make the tests work.

NathanReb: We have a related issue on dune already, I'll update it!

NathanReb: Did that here: https://github.com/ocaml/dune/issues/2470#issuecomment-537998717

PR #186, opened 2019-10-02
Used module_presence type to check if module is in the topevel env

clecat:
When generating the list of the objects to load from the toplevel environment (function `save_summary` in file `lib/top/mdx_top`), mdx was not checking if the module was supposed to be present in the environment.

Indeed, a new information defined by the type `module_presence` was introduced in 4.08.0 in the `Ocaml.Env.summary` type, but it is actually ignored with a wildcard.

I simply added a check, ignoring the module if it's presence was set to `Mp_absent`.

The problem could for example append when dealing with module aliases:
```ocaml
# module Test = String;;
module Test = String
```

This PR should therefore solve the mdx tests of those two packages:
- [containers](https://github.com/ocaml/opam-repository/pull/14791#issuecomment-529057288)
- [datalog](https://github.com/ocaml/opam-repository/pull/14970#issuecomment-537434474)

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -761,7 +761,7 @@ let rec save_summary acc s =
   | Env_value (summary, id, _) ->
      save_summary (Ident.name id :: acc) summary
 #if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
-  | Env_module (summary, id, _, _)
+  | Env_module (summary, id, Mp_present, _)
 #else
   | Env_module (summary, id, _)
 #endif
@@ -779,6 +779,9 @@ let rec save_summary acc s =
         else acc
       in
       save_summary acc summary
+#if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
+  | Env_module (_, _, Mp_absent, _)
+#endif
   | Env_empty -> acc
 #if OCAML_MAJOR >= 4 && OCAML_MINOR >= 4
   | Env_constraints (summary, _)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 
 #### Fixed
 
+- Used module_presence information on Env.summary to prevent fetching absent modules from the toplevel env (#186, @clecat)
 - Remove trailing whitespaces at the end of toplevel or bash evaluation lines
   (#166, @clecat)
 - Improve error reporting of ocaml-mdx test (#172, @Julow)
--- a/test/dune
+++ b/test/dune
@@ -19,6 +19,7 @@
     (run ocaml-mdx rule %{dep:semisemi.md})
     (run ocaml-mdx rule %{dep:exit.md})
     (run ocaml-mdx rule %{dep:padding.md})
+    (run ocaml-mdx rule %{dep:module_alias.md})
     (run ocaml-mdx rule %{dep:multilines.md})
     (run ocaml-mdx rule %{dep:lines.md})
     (run ocaml-mdx rule %{dep:lwt.md})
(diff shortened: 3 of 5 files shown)


clecat:
A great thanks to @let-def, who greatly helped me in pointing me to the module_presence information

avsm:
Ignore the failing `ocaml-ci` one -- @talex5 and I are working o that.  This is clear to merge as Travis passed.

avsm:
Any chance you could push a CHANGES entry for this as well @clecat ?

clecat:
Done, and thanks for the information :D

lib/top/mdx_top.ml
```diff
@@ -779,6 +779,9 @@ let rec save_summary acc s =
         else acc
       in
       save_summary acc summary
+#if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
+  | Env_module (_, _, Mp_absent, _)
+#endif
   | Env_empty -> acc
```
trefis: I'm not sure I understand what the purpose of this `save_summary` function is, but here you're stopping the recursion at the first absent module that you encounter, and that seems like a mistake.

trefis:
FTR: I think absent modules were already in the summary before 4.08 even though the presence information wasn't available in it.
So it seems somewhat unlikely (though not impossible) that that would be the issue.

trefis:
For instance:
```
# include struct module Fields = struct let a = 3 end end;;
```
Also gives:
```
>> Fatal error: Fields/89 unbound at toplevel
                Misc.Fatal_error
                Raised at file "utils/misc.ml", line 22, characters 20-31
                Called from file "list.ml", line 92, characters 20-23
                Called from file "lib/top/mdx_top.ml", line 659, characters 13-44
                Called from file "lib/top/mdx_top.ml", line 680, characters 25-37
                Called from file "bin/test/main.ml", line 351, characters 12-103
                Re-raised at file "bin/test/main.ml", line 347, characters 25-50
                Called from file "list.ml", line 110, characters 12-15
                Called from file "bin/test/main.ml", line 345, characters 4-402
                Called from file "lib/mdx.ml", line 99, characters 18-39
                Called from file "lib/mdx.ml", line 109, characters 23-48
                Called from file "bin/test/main.ml", line 360, characters 20-67
                Called from file "bin/test/main.ml", line 382, characters 4-151
                Called from file "cmdliner_term.ml", line 25, characters 19-24
                Called from file "cmdliner.ml", line 117, characters 32-39
```
And that's because of the use of `Ident.unique_toplevel_name`.

Basically, the issue is that in the summary, you don't have enough information to deduce how things are stored by the toplevel.
```
module M = <ME>
```
will be stored using the toplevel unique name when it is present **and** at the toplevel. A present module inside an `include struct end` or `open struct end` will be stored using its name. An absent module won't be stored, regardless of where it comes from (so the idea behind the initial patch was somewhat correct).

A temporary patch could be to try getting the unique name, catching the fatal error, trying the normal name, and if it errors again, raising the initial exception.

But in the long run, I think the current `save_summary` approach is a bad idea.

trefis:
> A temporary patch could be to try getting the unique name, catching the fatal error, trying the normal name, and if it errors again, raising the initial exception.

Looking at the code in more details, I'm not sure such an approach would work.

Btw, a "simpler" example also shows the problem
```
# include Hashtbl;;
>> Fatal error: Make/255 unbound at toplevel
                Misc.Fatal_error
```

I'll stand by my earlier comment: `save_summary` (as it currently is) doesn't seem like a good idea. I don't know what you're trying to do there, but you should probably do it differently.

---
PS: I might have given the impression that I would dig into this and produce a patch. To lift any such impression: I won't.

lib/top/mdx_top.ml
clecat: From what I remember, I think this function is used to be able to have several environments at the same time.
But indeed the recursion is broken by my changes, I'll get that solved

PR #187, opened 2019-10-02
Switch to using ocaml-ci-scripts

NathanReb:
This seems like a more reliable CI until we can use `ocaml-ci`.

This will allow us to easily add revdeps builds btw which could prove useful once we want to stabilize mdx's API and avoid breaking users' tests!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,19 @@
 language: c
+sudo: false
+services:
+  - docker
+install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
+script: bash ./.travis-docker.sh
 env:
-  # Check build and unit tests
-  # NOTE: testing needs OPAMBUILDTEST=true so that test deps are installed
-  - TO_TEST=tests OPAMBUILDTEST=true OCAML_VERSION=4.02.3
-  - TO_TEST=tests OPAMBUILDTEST=true OCAML_VERSION=4.03.0
-  - TO_TEST=tests OPAMBUILDTEST=true OCAML_VERSION=4.04.2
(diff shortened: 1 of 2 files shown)


NathanReb:
Opam has a `make` built-in variable that `opam lint` suggests we use instead of the raw `"make"`.

Now that we use the ci-scripts, it's properly picked up and triggers a build failure if we don't have a compliant opam file.

samoht:
Ok I've read https://github.com/realworldocaml/mdx/pull/185#discussion_r330673627 which answers my question. However this is quite fragile as it will break if you run this in a duniverse setting (with a toplevel call to `dune runtest`).

NathanReb:
What do you mean? Do you mean that it will break if we vendor `mdx` in a duniverse? Aliases aren't resolved within the duniverse so it won't run the tests in this case.

I agree it is a bit fragile and I wish there was a way to tell dune about the `ocaml-mdx` -> `ocaml-mdx-test` binary dependency but unfortunately there isn't any atm. I tried a couple things and they ended up suffering from the same race condition.

Also when using `mdx` in duniverse mode, `ocaml-mdx rule` adds a `(package mdx)` dependency to the generated rules, thus solving the dependecny issue. We can't do that outside a duniverse because `(package ...)` deps only work for local packages, not opam ones.

There's one last thing I can try which is to add an explicit dependency on the `install` alias for all `runtest` aliases that rely on `ocaml-mdx` but that is a bit tedious and doesn't solve the issue in a generic way either but at least running `dune runtest` should work then. Let's merge that as it already fix the opam test runs and I'll try that solution in a separate PR!

NathanReb:
Just had a quick look and the `(alias install)` dependency propably won't work as we generate most of the rules for `runtest`.

samoht:
I just mean that we should report that issue upstream as it should be fixed properly at one point :-)

PR #188, opened 2019-10-11
[WIP] Add latex generation

clecat:
This small PR adds support for latex file generation.
When launching the `output` command, the user can specify the option `-t` followed by either `html` or `latex` to choose the output format.

Keep in mind that generated files needs are not build as standalone.


clecat:
As mentionned [there](https://github.com/realworldocaml/book/pull/3215), I'll close this PR and move all the latex generation to the realworldocaml book

Issue #179, opened 2019-09-20
mdx swallows code on syntax errors, in some cases


yminsky:
This is an example that showed up in the nascent testing chapter for RWO:

 ```ocaml env=main
 # let gen_shape =
-	let open Quickcheck.Generator.Let_syntax in
-	let module G = Quickcheck.Generator in
-	let circle =
-	  let%map radius = Float.gen_positive in
-	  Circle { radius }
-	in
-	let rect =
-	  let%map height = Float.gen_positive
-	  and width = Float.gen_positive
-	  in
-	  Rect { height; width }
-	in
-	let poly =
-	  let%map points =
-		List.gen (G.both Float.gen_positive Float.gen_positive)
-	  in
-	  Poly points
-	in
-	G.union [circle; rect; poly];;
-val gen_shape : shape Quickcheck.Generator.t = <abstr>
+Characters 16-18:
+Error: Syntax error
 ```
Note how it deletes the input code and replaces it with a syntax error. Maybe it doesn't know what the output is for some reason?

PR #191, opened 2019-10-19
Convert tabulations to spaces to fix the parsing of code blocks in toplevel

gpetiot:
This allows to execute a code using tabulations instead of spaces (see #179).
Not sure this should close the issue as it would be better to print back the input when an exception has been raised, in addition to the error message.

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -349,6 +349,9 @@ let guess_ocaml_kind b =
 let toplevel ~file ~line lines = Toplevel (Toplevel.of_lines ~line ~file lines)
 
 let eval t =
+  let re_tab = Re.(compile (char '\t')) in
+  let contents = List.map Re.(replace_string re_tab ~by:"      ") t.contents in
+  let t = {t with contents} in
   match check_labels t with
   | Error e -> { t with value = Error e }
   | Ok ()   ->
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -360,7 +360,7 @@ let eval t =
     let s' =  Re.replace_string re_tab ~by:"    " s in
     if not (String.equal s s') then
       Misc.warn (location t)
-        ("Warning: tabulation of line " ^ Int.to_string i ^
+        ("Warning: tabulation of line " ^ string_of_int i ^
          " automatically converted to 4 spaces, please update the source.");
     s'
   in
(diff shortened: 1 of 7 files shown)


lib/block.ml
```diff
@@ -324,6 +324,9 @@ let guess_ocaml_kind b =
 let toplevel ~file ~line lines = Toplevel (Toplevel.of_lines ~line ~file lines)
 
 let eval t =
+  let re_tab = Re.(compile (char '\t')) in
+  let contents = List.map Re.(replace_string re_tab ~by:"      ") t.contents in
```
Julow: Could you add a comment saying why ?

gpetiot: I added a warning, it sounds enough to me (can be verbose but we don't have a non-duplicate mechanism yet)

lib/block.ml
```diff
@@ -324,6 +330,17 @@ let guess_ocaml_kind b =
 let toplevel ~file ~line lines = Toplevel (Toplevel.of_lines ~line ~file lines)
 
 let eval t =
+  let re_tab = Re.(compile (char '\t')) in
+  let on_line i s =
+    let s' =  Re.replace_string re_tab ~by:"    " s in
```
Julow: This is replacing all the tabs and we only need to replace the leading tabs. Also, tabs that are not at the beginning of the line doesn't always have the same size.
Maybe we don't need the full power of `Re` here. It may also make things more complicated here.

test/bin/test_expect/dune.inc
```diff
@@ -59,6 +59,16 @@
   )
 )
 
+(alias
+  (name runtest)
+  (deps (:x tabs.md) (:y tabs.md.expected) (package mdx))
+  (action
+    (progn
+      (run ocaml-mdx test --force-output  %{x})
```
Julow: The warnings are printed to stderr and are not checked and spam the test outputs.
Maybe after https://github.com/realworldocaml/mdx/pull/194 you could redirect both outputs to the output file ?

lib/block.ml
```diff
@@ -324,6 +330,17 @@ let guess_ocaml_kind b =
...
+    if not (String.equal s s') then
+      Misc.warn (location t)
```
Julow: Maybe printing the warning for every lines with tabs is a bit overkill ?

gpetiot:
Closing this one as I'm not satisfied with this solution, feel free to re-open to take ownership or start again from scratch.

PR #194, opened 2019-11-15
Add a --output/-o option to specify mdx test output

### LINKED ISSUE #193: Allow more configuration of the test command
Currently the `ocaml-mdx test` command isn't very flexible with regard to its output.

It always writes it to the input file with a `.corrected`. While this is a good default, I think it'd be good to:
1. Document it in the `test` command man page.
2. Add a `--to-stdout`/`-O` option to write to stdout instead
3. Add a `--output <file>`/`-o <file>` option to write to a different file

These would all be backward compatible changes with the current behaviour.

While I believe we should, in the future, try to promote the dune workflow over the `ocaml-mdx test` one, I think until we have a proper integration these additions would make more usable to everyone, including ourselves since I actually met that limitation while working on the test!

What do you think? @samoht @Julow

NathanReb:
Fixes #193 

This PR adds a `--output/-o` option to set the output for the `ocaml-mdx test` command.

It used to always write to `<input_file>.corrected` so the default behaviour is preserved, i.e. if you do not pass the `--output` option, it still writes to that file.
If passed `--output -`, the command will write the corrected file to stdout instead of writing it to a file. Setting stdout as the output implies `--force-output`. This seems like the best behaviour if you want to pipe it into some other command or redirect it to a file for later diffing.

I deleted the code of `Misc.run_expect_test` (which isn't part of the pulbic API) since it was only used in `Mdx.run`. Instead I created a more generic function `Mdx.run_str` which returns a pair composed of the test result (whether the corrected version is equal to the input or not) and the corrected content.

It is used by two new functions `Mdx.run_to_stdout` and `Mdx.run_to_file`. The former just writes the corrected content to stdout no matter what, the second preserves the behaviour of `Misc.run` except that it takes an `outfile` argument.

I wasn't too sure about what needed to be done with the "clean up" behaviour which takes care of deleting the corrected file if it exists and the test succeeded so I kept it for any output file, whether it's the default are an explicitly specified one. I don't think it's necessary to do that when used through dune as it doesn't use the file presence alone to determine which files need to be promoted but that's unrelated to this PR.

Finally, to preserve the public API, I exposed these two new functions and re-implemented `Mdx.run` on top of `Mdx.run_to_file` so that its behaviour and signature remain the same.

I also added tests to ensure that this option works properly.

Let me know what you think!

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -372,13 +372,12 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     Format.pp_print_flush ppf ();
     Buffer.contents buf
   in
-  let run_to_file outfile =
-    Mdx.run_to_file ?syntax ~force_output ~outfile ~f:gen_corrected file
-  in
   (match (output : Cli.output option) with
    | Some Stdout -> Mdx.run_to_stdout ?syntax ~f:gen_corrected file
-   | Some (File outfile) -> run_to_file outfile
-   | None -> run_to_file (file ^ ".corrected"));
+   | Some (File outfile) ->
+     Mdx.run_to_file ?syntax ~outfile ~f:gen_corrected file
+   | None ->
+     Mdx.run ?syntax ~force_output ~f:gen_corrected file);
   Hashtbl.iter (write_parts ~force_output) files;
   0
 
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -266,7 +266,7 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     (`Not_verbose not_verbose) (`Syntax syntax) (`Silent silent)
     (`Verbose_findlib verbose_findlib) (`Prelude prelude)
     (`Prelude_str prelude_str) (`File file) (`Section section) (`Root root)
-    (`Direction direction) (`Force_output force_output) =
+    (`Direction direction) (`Force_output force_output) (`Output output) =
   let c =
     Mdx_top.init ~verbose:(not not_verbose) ~silent ~verbose_findlib ()
   in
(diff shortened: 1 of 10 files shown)


bin/test/main.ml
```diff
@@ -353,25 +353,33 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
-            List.iter (fun (k, v) -> Unix.putenv k v) (Block.set_variables t);
-            try
-              Mdx_top.in_env (Block.environment t)
-                (fun () -> test_block ~ppf ~temp_file t)
-            with Failure msg ->
-              raise (Test_block_failure (t, msg))
-        ) items;
-      Format.pp_print_flush ppf ();
-      Buffer.contents buf);
+  let _ = Misc.remove_file in
```
Julow: What is this for ?

bin/test/main.ml
```diff
@@ -353,25 +353,33 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
+  let gen_expected file_contents items =
```
Julow: Should this be `gen_corrected` ?

bin/test/main.ml
```diff
@@ -353,25 +353,33 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
+    Format.pp_print_flush ppf ();
+    Buffer.contents buf
+  in
+  let run_to_file outfile =
+    Mdx.run_to_file ?syntax ~force_output ~outfile ~f:gen_expected file
+  in
+  (match (output : Cli.output option) with
+   | Some Stdout -> Mdx.run_to_stdout ?syntax ~f:gen_expected file
+   | Some (File outfile) -> run_to_file outfile
+   | None -> run_to_file (file ^ ".corrected"));
```
Julow: You should call `Mdx.run` here. It does the `^ ".corrected"`.

lib/mdx.mli
```diff
@@ -69,6 +69,16 @@ val parse_lexbuf: syntax -> Lexing.lexbuf -> t
...
 
+val run_to_stdout : ?syntax: syntax -> f:(string -> t -> string) -> string -> unit
+
+val run_to_file :
+  ?syntax: syntax ->
+  ?force_output: bool ->
+  f:(string -> t -> string) ->
+  outfile: string ->
+  string ->
+  unit
```
Julow: Can you add a little documentation ?
I see no problem in moving this function below `run` so it can reference it.

bin/test/main.ml
NathanReb: Yep, I'll change that!

bin/test/main.ml
NathanReb: Hmm some debug leftover, I'll remove it!

bin/test/main.ml
NathanReb: `Mdx.run` is still in the API for compatibility reasons, I'd rather not use it internally and use the more up-to-date equivalent instead.

Ideally we'll deprecate it in a little while!

lib/mdx.mli
NathanReb: I did it the other way around, I documented the `run_to_file` and `run_to_stdout` functions and describe `run` in term of `run_to_file`

samoht:
Can you use the compat layer to keep support for 4.02? Thanks

```
# File "bin/cli.ml", line 113, characters 17-19:
# Error: Unbound constructor Ok
```

NathanReb:
I think `force-output` should always be true, that would spare some effort on our hand and offer a more consistent behaviour but let's keep that for a separate PR!

NathanReb:
> Can you use the compat layer to keep support for 4.02? Thanks

Done!

Julow:
> I think `force-output` should always be true, that would spare some effort on our hand and offer a more consistent behaviour but let's keep that for a separate PR!

I meant setting `force-output` when `--output` is set. Changing the default behavior requires an other PR but this is the behavior of `--output` that I'm talking about. Let's not implement legacy code.

NathanReb:
I understood what you meant, my point was that we can just tackle one thing at a time. Having `--force-output` implied when output is specified means having completely different behaviours for `--output` and the default which currently is just `--output <input>.corrected`.

I'd rather drop `force-output` completely in a future release. Let's keep the breaking changes to a minimum for the next one.

Julow:
I don't agree. The default feels like `--output <input>.corrected` but it doesn't have to be !
My point was that this has nothing to do with the default behavior and it should be discussed and implemented correctly the first time.
This behavior (not writing a file, or worse, removing it when, as a user, I wrote it on the command-line) is not right for me.
You already implemented a special case for `--output -`.

NathanReb:
Yeah it's a good point, I just fixed it!

PR #195, opened 2019-11-18
Remove output subcommand

NathanReb:
This removes the `ocaml-mdx output` subcommand since it has now been moved to https://github.com/realworldocaml/book.

It shouldn't be an issue as RWO is the only user of this command!


samoht:
Looks good but can you also remove https://github.com/realworldocaml/mdx/blob/master/mdx.opam#L30 so that the pass could pass?

samoht:
> Is there a replacement in RWO yet ?

Yes, it's here: https://github.com/realworldocaml/book/blob/master/bin/bin/convert_md.ml

NathanReb:
> Looks good but can you also remove https://github.com/realworldocaml/mdx/blob/master/mdx.opam#L30 so that the tests could pass?

Good catch! I'll remove that

NathanReb:
Both CI are green! Let's merge!

PR #196, opened 2019-11-20
Improve test infrastructure

NathanReb:
Depends on #194 

This PRs refactors the tests even further. The main motivation is to have deterministic, well defined and race condition-free tests in mdx. The recent attempts to release it to opam showcased that the tests were a bit fragile.

The first effect of this PR is that tests dependencies are properly specified and running `dune runtest` from scratch now works. Note that the root cause, i.e. the dependency between the native and bytecode binaries, isn't solved but that the tests depend on `(package mdx)`.

Each test case has been moved to its own folder and no file is used by different tests that can run in parallel anymore. We should make sure that new tests have all the files they need in their own folder and that they never try to reach the outside.

Tests for different commands have been split up. That means we don't test `ocaml-mdx pp` through an `ocaml-mdx test` test anymore. Separate concerns, separate tests. Same goes for `ocaml-mdx rule` which we used as a test utility. It should be tested on its own. You'll note that it's not yet the case in this PR. I was thinking about adding tests for the `rule` subcommand separately given the size of this PR but I'm happy to add them there if we feel like it is necessary.

I'm also planning to add a `CONTRIBUTING.md` describing the tests architecture and how to add new ones.


NathanReb:
Need to fix some compatibility issues!

test/bin/mdx-pp/dune
```diff
@@ -0,0 +1,3 @@
+(executable
+ (name gen_dune_rules)
+ (libraries fmt gen_rule_helpers))
```
Julow: Why not move this executable to where it's used, in `expect/` ? `expect/dune` uses a relative path (`../gen_dune_rule.exe`) to access this and it is not used in the `misc` directory.

test/bin/mdx-pp/misc/compile/dune
```diff
@@ -0,0 +1,10 @@
+(executable
```
Julow: Why an extra `misc/` level ? `compile` would fit near `expect` in the parent directory, I think.

test/bin/mdx-pp/gen_dune_rules.ml
```diff
@@ -0,0 +1,28 @@
...
+
+(** Tests that 'cd <dir> && ocaml-mdx pp [options] <file>' exits with a
+    failing code and that its output is equal to the content of the
+    '.expected' file in <dir>. *)
+let pp_failure_action fmt dir =
+  Fmt.pf fmt
+    {|
+  (with-outputs-to %%{target}
+   (chdir %s
+    (system "! ocaml-mdx pp %a%s)))|}
```
Julow: Repeating this strings, indenting and passing more arguments to pf could be avoided by having utils that print the `(with-outputs-to ..)` or `(chdir ..)` part.
It would be used like this: ``pp_with_outputs_to `Target (pp_chdir dir.dir_name (pp_system ("!" :: "ocaml-mdx" :: "pp" :: options_as_list @ [test_file]))) fmt``
Indentation is handled by boxes so it's no longer a concern.

What do you think ?

test/bin/mdx-pp/dune
NathanReb: I try to maintain consistent structure between the various sub commands end to end tests, that is:
```
test/bin/mdx-test/
├── dune
├── gen_dune_rules.ml
├── expect
│   ├── ... # test case folders
│   ├── dune
│   └── dune.inc
├── failure
│   ├── ... # test case folders
│   ├── dune
│   └── dune.inc
└── misc
    └── ... # test case folders
```

And that even if we don't have error test cases for `ocaml-mdx pp` yet. Adding some remains simpler and I believe it makes it easier to navigate the end to end tests if each command folder has the same structure.

test/bin/mdx-pp/misc/compile/dune
NathanReb: For the same reasons I mentioned above, I think it's better if all test case folders have the same "depth" in the folder structure, that way you don't have to try and interpret the names of the folders in `test/bin/mdx-pp/` to figure out which one are used to group similar test cases and which ones are individual test cases.

test/bin/mdx-pp/gen_dune_rules.ml
NathanReb: I agree having to care about indentation isn't great and I wouldn't do it like that if it was part of the core of mdx but I'm willing to sacrifice that here, especially given that it makes it more readable than with `pp` combinators. I.e. it's easier to read the plain dune field as a string, even though you proposal uses self-explanatory names.

Eventually I'd prefer piping it into `dune format-dune-file` so that the script code doesn't have to care about formatting at all.

Would that work for you?

test/bin/mdx-pp/dune
Julow: I understood that. I was talking of:

```
test/bin/*/
├── expect
│   ├── ... # test case folders
│   ├── dune
│   ├── gen_dune_rules.ml
│   └── dune.inc
├── failure
│   ├── ... # test case folders
│   ├── dune
│   └── dune.inc
└── misc
    └── ... # test case folders
```

- `failure` and `misc` seems not to need `gen_dune_rules`
- Generated rules are very specific to "expect" tests

test/bin/mdx-pp/misc/compile/dune
Julow: I see, makes sense.

test/bin/mdx-pp/dune
NathanReb: `failure` does use `gen_dune_rules`

test/bin/mdx-pp/gen_dune_rules.ml
Julow: I don't agree with your first point. I think `pp`s will be more readable, reduce boilerplate, avoid confusing printf and disallow invalid outputs (because they are written once and combinations are restricted to what the API allows).
Also, I think the `pp` approach is easier to review because it's not strings and avoid confusions with format strings syntax.

For example, on that line, the closing `"` is missing.

I wasn't making a point on indentation, but will still be a burden and can be avoided. I think indenting ocaml code (applications of `pp`s) is easier than indenting a string stripped of its context.

test/bin/mdx-pp/dune
Julow: You are right.

test/bin/mdx-pp/gen_dune_rules.ml
NathanReb: I think the current implementation is fine enough and a huge improvement over the previous situation.

We can refactor that later but I'd like to get these improvement merged and try to release again.

NathanReb:
To summarize our offline discussions, we're merging but we'll submit subsequent PRs to:
- Split binaries between failure and expect
- Improve the `Gen_rule_helpers`'s API to use proper pretty printers for dune rules rather than raw format strings

test/bin/mdx-pp/gen_dune_rules.ml
Julow: Yes, I agree.

PR #197, opened 2019-11-22
top: move compatibility functions to `compat_top`

emillon:
This isolates `cppo` directives to a single file, as done in `mdx` itself. In particular, the corresponding `.mli` does not have any preprocessing directives.

Reasons for doing that:
- it gives us a single place to clean up when changing compatibility
- it makes it possible to format `mdx_top` in the future
- if we want to get rid of `cppo`, this limits the scope of things to change

The approaches are a bit different from function to function:
- some are just constructors (`sig_class`, `attribute`)
- some are deconstructors (`match_env`)
- some split the work between a function that stays in `mdx_top` and a mapper in `compat_top` (`map_error_loc`)

This can be tested locally by using the following `dune-workspace.dev` file:

```
(lang dune 1.0)
(context (opam (switch 4.02.3)))
(context (opam (switch 4.03.0)))
(context (opam (switch 4.04.2)))
(context (opam (switch 4.05.0)))
(context (opam (switch 4.06.1)))
(context (opam (switch 4.07.1)))
(context (opam (switch 4.08.0)))
```

and:

    dune build @install @runtest --workspace dune-workspace.dev

Let me know what you think.

cc @NathanReb

### DIFF
--- a/lib/top/compat_top.ml
+++ b/lib/top/compat_top.ml
@@ -1,0 +1,285 @@
+open Mdx.Migrate_ast
+
+#if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
+let try_finally ~always f = Misc.try_finally f ~always
+#else
+let try_finally ~always f = Misc.try_finally f always
+#endif
+
+let map_error_loc ~f (error : Location.error) =
+#if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
+  let f_msg (msg : Location.msg) =
+    { msg with loc = f msg.loc}
+  in
+  { error with main = f_msg error.main;
+                sub = List.map f_msg error.sub; }
+#else
+  let rec aux (error : Location.error) =
+     {error with sub = List.map aux error.sub;
+                 loc = f error.loc}
+  in
+  aux error
+#endif
+
+let error_of_exn exn =
+#if OCAML_MAJOR >= 4 && OCAML_MINOR > 5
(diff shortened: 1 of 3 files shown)


emillon:
note: the logic to compute version numbers is buggy in places, for example `MAJOR >= 4 && MINOR >= 3` does not do the right thing for `5.0`, but that does not matter too much for now.

lib/top/compat_top.ml
```diff
@@ -0,0 +1,285 @@
+open Mdx.Migrate_ast
+
+#if OCAML_MAJOR >= 4 && OCAML_MINOR >= 8
```
Julow: Cppo supports tuples.
So you can define `#define VERSION (OCAML_MAJOR, OCAML_MINOR)` and compare `#if VERSION >= (4, 8)`.

emillon: Nice! I didn't know that. I'll submit a PR to change that once this one is merged.

NathanReb:
This seems to break compatibility with `4.09.0`!

emillon:
Yes, I haven't even tried with 4.09 locally :/ I'll update this.

emillon:
Done! The failure is interesting - I was opening `Parsetree` under a `#if OCAML >= 4.08`. This works in 4.08 because `Parsetree` happens to be the OMP 4.08 AST. But we want the current AST, `Parsetree_`. The 4.09 build caught that. Neat!

PR #198, opened 2019-11-26
cppo: only compare `OCAML_VERSION` as tuples

emillon:
To test if the version is greater or equal than 4.03.0,
The code base uses variants of `OCAML_MAJOR >= 4 && OCAML_MINOR >= 3`.
This is incorrect after 5.0.0. The right way is to use lexicographic
comparison, in extension:

    (OCAML_MAJOR > 4) || (OCAML_MAJOR == 4 && OCAML_MINOR >= 3)

But fortunately `cppo` supports tuples natively (thanks @Julow for the
tip!), so we can express that as just `OCAML_VERSION >= (4, 3, 0)`.

In addition, this commit standardizes the order in which comparisons are
done:

- conditionals with an else are written with >=:

```
#if OCAML_VERSION >= xxx
  ...
#else
  ...
#endif
```

- polyfills for former versions of ocaml are written with <:

```
#if OCAML_VERSION < xxx
  ...function introduced in xxx...
#endif
```

### DIFF
--- a/lib/compat.ml
+++ b/lib/compat.ml
@@ -17,11 +17,11 @@
 module String = struct
   include String
 
-#if OCAML_MAJOR = 4 && OCAML_MINOR < 3
-    let equal x y = String.compare x y = 0
+#if OCAML_VERSION < (4, 3, 0)
+  let equal x y = String.compare x y = 0
 #endif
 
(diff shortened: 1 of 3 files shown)


emillon:
@NathanReb done!

PR #203, opened 2020-01-08
Feature / Include any file

voodoos:
Allow `mdx` to include content from any file, not only OCaml files.
When including non-OCaml files part splitting is not supported and raises an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Here is an examples of toplevel OCaml code:
 ### File sync
 `mdx` is also capable of two-way synchronization of fenced code blocks with
 OCaml files using
-the labels `file`. When an OCaml file is included it can be automaticcaly sliced
+the label `file`. When an OCaml file is included it can be automatically sliced
 if it contains
 annotations such as `[@@@part "partName"]` and if the block has the label `part=partName`:
 
--- a/README.md
+++ b/README.md
@@ -96,20 +96,19 @@ Here is an examples of toplevel OCaml code:
     ```
 
 ### File sync
-`mdx` is also capable of two-way synchronization of fenced code blocks with
-OCaml files using
-the label `file`. When an OCaml file is included it can be automatically sliced
-if it contains
-annotations such as `[@@@part "partName"]` and if the block has the label `part=partName`:
+`mdx` is also capable of including content from files in fenced code blocks
+using the label `file`. When an OCaml file is included it can be automatically
+sliced if it contains annotations such as `[@@@part "partName"]` and if the
+block has the label `part=partName`:
 
     ```ocaml file=sync_to_md.ml,part=partName
     ```
 
-Non-OCaml files can also be read and included (file to md sync only) in a block:
+Non-OCaml files can also be read and included in a block:
 
     ``` file=any_file.txt
     ```
-Part splitting is only supported for OCaml files.
+However, part splitting is only supported for OCaml files.
 
 ### Pre-processing
 
--- a/README.md
+++ b/README.md
@@ -95,6 +95,22 @@ Here is an examples of toplevel OCaml code:
     42
     ```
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -95,6 +95,22 @@ Here is an examples of toplevel OCaml code:
     42
     ```
 
+### File sync
+`mdx` is also capable of two-way synchronization of fenced code blocks with
+OCaml files using
+the labels `file`. When an OCaml file is included it can be automaticcaly sliced
```
NathanReb: tiny typos:
```suggestion
the label `file`. When an OCaml file is included it can be automatically sliced
```

README.md
```diff
@@ -95,6 +95,22 @@ Here is an examples of toplevel OCaml code:
     42
     ```
 
+### File sync
+`mdx` is also capable of two-way synchronization of fenced code blocks with
```
NathanReb: I wouldn't mention that it can be done from the md file to the ml one as we're planning on dropping that.

bin/test/main.ml
```diff
@@ -291,8 +298,16 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     match active, non_deterministic, Block.mode t, Block.value t with
     (* Print errors *)
     | _, _, _, Error _ -> Block.pp ?syntax ppf t
-    (* Skip raw blocks. *)
-    | true, _, _, Raw -> Block.pp ?syntax ppf t
+    (* Skip or copy raw blocks. Without parts support *)
+    | true, _, _, Raw -> 
+    (match Block.part t with
```
NathanReb: I would probably swap the two nested pattern matching here as the error mentions files (as in the ones passed to the `file` label) and you can still get that error if you just wrote:
```
\```scheme part=2
\```
```
We can also make the error a bit more generic, "Parts are not supported for non-OCaml code blocks".
What do you think?

Note that I'm just nit-picking here as I believe the error should actually be reported at parsing time as we shouldn't allow some labels on some blocks anyway but that's a bit out of scope for this PR.

test/bin/mdx-test/expect/casual-file-inc/test-case.opts
```diff
@@ -0,0 +1 @@
+--direction=to-md
```
NathanReb: I think it makes sense for this test to remove this option as it is unrelated to its behaviour.

To clarify, this test has value whether the `--direction` option exists or not whereas the second one is here explicitly to show it's behaviour is independent from the option value, in particular when it is `to-ml`.
When we remove the option, we'll delete the second test but should be able to leave this one unchanged.

What do you think?

bin/test/main.ml
```diff
@@ -190,6 +194,18 @@ let read_parts file =
...
+  match Mdx_top.Part.find parts.current ~part with
+  | None       ->
+    Fmt.failwith "Cannot find part %S in %s"
+      (match part with None -> "" | Some p -> p)
+      file
+  | Some lines ->
+    let contents = String.concat ~sep:"\n" lines in
+    String.trim contents
+
+
```
NathanReb: Unnecessary extra empty line!
```suggestion
```

README.md
voodoos: Done in 54fcc23da04fa1542910cf0caec3d5688123fe8e

bin/test/main.ml
voodoos: Option 2: 
> We can also make the error a bit more generic, "Parts are not supported for non-OCaml code blocks".

Implemented in 5ae04dd

test/bin/mdx-test/expect/casual-file-inc/test-case.opts
voodoos: Agreed, done in 32e4952

PR #206, opened 2020-01-09
Add support for mli files

jsomers:
Teach mdx to process the code blocks inside docstrings in mli files.

For instance, the following:

```ocaml
(** {[# List.map (fun x -> x * x) [1; 2; 3]]} *)
```

is corrected in-place to become:

```ocaml
(** {[# List.map (fun x -> x * x) [1; 2; 3]
      - : int list = [1; 4; 9]
    ]} *)
```

## Implementation

1. We lex mli files in two phases: first, we use the OCaml lexer (via `ppxlib`) to extract docstrings, then we use the odoc parser to extract code blocks within them. We then slice the input into contiguous regions of `Text` and `Block` items, just like the lexer for `Normal` syntax.

2. We add a new `Syntax.Mli` and thread `~syntax` arguments through a few more places so that we can properly format the output of mli's (avoiding newlines, etc.).

3. Since mli files are often written with formatters like ocamlformat and ocp-indent, and since these formatters might clobber the output of mdx, we make mdx's diff for the `Mli` syntax whitespace-invariant.

## Notes

* There is currently no support for header attributes (like `env=main`), because the odoc parser doesn't currently support an extended `{env=main[ let x = ... ]}` syntax for code blocks. This will have to be added later.

* We made changes to mdx to get the library to build within our environment (we're using Jenga). I also made the changes in this pull request against an earlier version of mdx, and wrote tests for them according to our internal conventions. I did my best to port the changes to this new version. But there are currently no tests and I'm not sure that this will build!

So I'd appreciate any instructions on how to get mdx running with dune (from scratch, more or less), and how to run the test suite.

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -18,7 +18,7 @@ let syntax =
   in
   let syntax = (parse, Mdx.Syntax.pp) in
   let doc =
-    "Which syntax to use. Either 'markdown' (also 'normal') or 'cram'."
+    "Which syntax to use. Either 'markdown' (also 'normal'), 'cram', or 'mli'."
   in
   named
     (fun x -> `Syntax x)
@@ -47,6 +47,12 @@ let silent_eval =
   named
     (fun x -> `Silent_eval x)
     Arg.(value & flag & info [ "silent-eval" ] ~doc)
+
+let record_backtrace =
+  let doc = "Print backtraces when evaluating toplevel phrases." in
+  named
+    (fun x -> `Record_backtrace x)
+    Arg.(value & flag & info [ "record-backtrace" ] ~doc)
 
 let silent =
   let doc = "Do not show any (phrases and findlib directives) results." in
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -120,7 +120,7 @@ let run_cram_tests ?syntax t ?root ppf temp_file pad tests =
         output;
       Cram.pp_exit_code ~pad ppf n)
     tests;
-  Block.pp_footer ?syntax ppf ()
+  Block.pp_footer ?syntax ppf t
 
 let eval_test ?block ?root c test =
   Log.debug (fun l ->
@@ -152,7 +152,7 @@ let split_lines lines =
   in
   List.fold_left aux [] (List.rev lines)
 
-let eval_ocaml ~block ?root c ppf ~line lines errors =
+let eval_ocaml ~block ?syntax ?root c ppf ~line lines errors =
   let test =
     Toplevel.{ vpad = 0; hpad = 0; line; command = lines; output = [] }
   in
@@ -163,7 +163,7 @@ let eval_ocaml ~block ?root c ppf ~line lines errors =
     | _ -> assert false
   in
   match eval_test ?root ~block c test with
-  | Ok _ -> Block.pp ppf (update ~errors:[] block)
+  | Ok _ -> Block.pp ?syntax ppf (update ~errors:[] block)
   | Error lines ->
       let errors =
         let lines = split_lines lines in
@@ -176,12 +176,12 @@ let eval_ocaml ~block ?root c ppf ~line lines errors =
               | `Output x -> `Output (ansi_color_strip x))
             (Output.merge output errors)
       in
-      Block.pp ppf (update ~errors block)
+      Block.pp ?syntax ppf (update ~errors block)
 
 let lines = function Ok x | Error x -> x
 
-let run_toplevel_tests ?root c ppf tests t =
-  Block.pp_header ppf t;
+let run_toplevel_tests ?syntax ?root c ppf tests t =
+  Block.pp_header ?syntax ppf t;
   List.iter
     (fun test ->
       let lines = lines (eval_test ?root ~block:t c test) in
@@ -200,7 +200,7 @@ let run_toplevel_tests ?root c ppf tests t =
               Output.pp ~pad ppf (`Output line))
         output)
     tests;
-  Block.pp_footer ppf ()
+  match syntax with Some Syntax.Mli -> () | _ -> Block.pp_footer ?syntax ppf t
 
 type file = { first : Mdx.Part.file; current : Mdx.Part.file }
 
@@ -242,16 +242,16 @@ let write_parts ~force_output file parts =
       flush oc;
       close_out oc
 
-let update_block_content ppf t content =
-  Block.pp_header ppf t;
+let update_block_content ?syntax ppf t content =
+  Block.pp_header ?syntax ppf t;
   Output.pp ppf (`Output content);
-  Block.pp_footer ppf ()
+  Block.pp_footer ?syntax ppf t
 
-let update_file_or_block ?root ppf md_file ml_file block part =
+let update_file_or_block ?syntax ?root ppf md_file ml_file block part =
   let root = root_dir ?root ~block () in
   let dir = Filename.dirname md_file in
   let ml_file = resolve_root ml_file dir root in
-  update_block_content ppf block (read_part ml_file part)
+  update_block_content ?syntax ppf block (read_part ml_file part)
 
 exception Test_block_failure of Block.t * string
 
@@ -275,10 +275,12 @@ let eval_prelude ?root top prelude prelude_str =
   | _ -> Fmt.failwith "only one of --prelude or --prelude-str shoud be used"
 
 let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
-    (`Silent_eval silent_eval) (`Syntax syntax) (`Silent silent)
-    (`Verbose_findlib verbose_findlib) (`Prelude prelude)
-    (`Prelude_str prelude_str) (`File file) (`Section section) (`Root root)
-    (`Force_output force_output) (`Output output) =
+    (`Silent_eval silent_eval) (`Record_backtrace record_backtrace)
+    (`Syntax syntax) (`Silent silent) (`Verbose_findlib verbose_findlib)
+    (`Prelude prelude) (`Prelude_str prelude_str) (`File file)
+    (`Section section) (`Root root) (`Force_output force_output)
+    (`Output output) =
+  Printexc.record_backtrace record_backtrace;
   let syntax =
     match syntax with Some syntax -> Some syntax | None -> Syntax.infer ~file
   in
@@ -292,15 +294,17 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
       | Raw _ -> print_block ()
       | Include { file_included; file_kind = Fk_ocaml { part_included } } ->
           assert (syntax <> Some Cram);
-          update_file_or_block ?root ppf file file_included t part_included
+          update_file_or_block ?syntax ?root ppf file file_included t
+            part_included
       | Include { file_included; file_kind = Fk_other _ } ->
           let new_content = read_part file_included None in
-          update_block_content ppf t new_content
+          update_block_content ?syntax ppf t new_content
       | OCaml { non_det; env; errors } ->
           let det () =
             assert (syntax <> Some Cram);
             Mdx_top.in_env env (fun () ->
-                eval_ocaml ~block:t ?root c ppf ~line:t.line t.contents errors)
+                eval_ocaml ~block:t ?syntax ?root c ppf ~line:t.line t.contents
+                  errors)
           in
           with_non_det non_deterministic non_det ~command:print_block
             ~output:det ~det
@@ -316,7 +320,11 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
             ~det:(fun () ->
               run_cram_tests ?syntax t ?root ppf temp_file pad tests)
       | Toplevel { non_det; env } ->
-          let tests = Toplevel.of_lines ~file:t.file ~line:t.line t.contents in
+          let tests =
+            let syntax = Util.Option.value syntax ~default:Normal in
+            Toplevel.of_lines ~syntax ~file:t.file ~line:t.line ~column:t.column
+              t.contents
+          in
           with_non_det non_deterministic non_det ~command:print_block
             ~output:(fun () ->
               assert (syntax <> Some Cram);
@@ -336,7 +344,7 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
             ~det:(fun () ->
               assert (syntax <> Some Cram);
               Mdx_top.in_env env (fun () ->
-                  run_toplevel_tests ?root c ppf tests t))
+                  run_toplevel_tests ?syntax ?root c ppf tests t))
     else print_block ()
   in
   let gen_corrected file_contents items =
@@ -346,7 +354,7 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     let ppf = Format.formatter_of_buffer buf in
     List.iter
       (function
-        | (Section _ | Text _) as t -> Mdx.pp_line ?syntax ppf t
+        | (Mdx.Document.Section _ | Text _) as t -> Mdx.pp_line ?syntax ppf t
         | Block t -> (
             List.iter (fun (k, v) -> Unix.putenv k v) (Block.set_variables t);
             try test_block ~ppf ~temp_file t
@@ -376,11 +384,12 @@ let report_error_in_block block msg =
   Fmt.epr "Error in the %scode block in %s at line %d:@]\n%s\n" kind block.file
     block.line msg
 
-let run setup non_deterministic silent_eval syntax silent verbose_findlib
-    prelude prelude_str file section root force_output output : int =
+let run setup non_deterministic silent_eval record_backtrace syntax silent
+    verbose_findlib prelude prelude_str file section root force_output output :
+    int =
   try
-    run_exn setup non_deterministic silent_eval syntax silent verbose_findlib
-      prelude prelude_str file section root force_output output
+    run_exn setup non_deterministic silent_eval record_backtrace syntax silent
+      verbose_findlib prelude prelude_str file section root force_output output
   with
   | Failure f ->
       prerr_endline f;
@@ -399,9 +408,9 @@ let cmd =
   let doc = "Test markdown files." in
   ( Term.(
       pure run $ Cli.setup $ Cli.non_deterministic $ Cli.silent_eval
-      $ Cli.syntax $ Cli.silent $ Cli.verbose_findlib $ Cli.prelude
-      $ Cli.prelude_str $ Cli.file $ Cli.section $ Cli.root $ Cli.force_output
-      $ Cli.output),
+      $ Cli.record_backtrace $ Cli.syntax $ Cli.silent $ Cli.verbose_findlib
+      $ Cli.prelude $ Cli.prelude_str $ Cli.file $ Cli.section $ Cli.root
+      $ Cli.force_output $ Cli.output),
     Term.info "ocaml-mdx-test" ~version:"%%VERSION%%" ~doc ~exits ~man )
 
 let main () = Term.(exit_status @@ eval cmd)
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -73,6 +73,7 @@ type value =
 
 type t = {
   line : int;
+  column : int;
   file : string;
   section : section option;
   dir : string option;
@@ -108,10 +109,12 @@ let dump_value ppf = function
   | Toplevel _ -> Fmt.string ppf "Toplevel"
   | Include _ -> Fmt.string ppf "Include"
 
-let dump ppf ({ file; line; section; labels; contents; value; _ } as b) =
+let dump ppf ({ file; line; column; section; labels; contents; value; _ } as b)
+    =
   Fmt.pf ppf
-    "{@[file: %s;@ line: %d;@ section: %a;@ labels: %a;@ header: %a;@\n\
-    \        contents: %a;@ value: %a@]}" file line
+    "{@[file: %s;@ line: %d;@ column: %d;@ section: %a;@ labels: %a;@ header: \
+     %a;@\n\
+    \        contents: %a;@ value: %a@]}" file line column
     Fmt.(Dump.option dump_section)
     section
     Fmt.Dump.(list Label.pp)
@@ -121,13 +124,34 @@ let dump ppf ({ file; line; section; labels; contents; value; _ } as b) =
     Fmt.(Dump.list dump_string)
     contents dump_value value
 
-let pp_lines syntax =
+let pp_lines syntax t =
   let pp =
-    match syntax with Some Syntax.Cram -> Fmt.fmt "  %s" | _ -> Fmt.string
+    match syntax with
+    | Some Syntax.Cram -> Fmt.fmt "  %s"
+    | Some Syntax.Mli ->
+        let i = ref (t.column + 2) in
+        let s = ref Fmt.string in
+        while !i > 0 do
+          s := Fmt.prefix (Fmt.unit " ") !s;
+          i := !i - 1
+        done;
+        !s
+    | _ -> Fmt.string
   in
   Fmt.(list ~sep:(unit "\n") pp)
 
-let pp_contents ?syntax ppf t = Fmt.pf ppf "%a\n" (pp_lines syntax) t.contents
+let lstrip string =
+  let hpad = Misc.hpad_of_lines [ string ] in
+  Astring.String.with_index_range string ~first:hpad
+
+let pp_contents ?syntax ppf t =
+  match syntax with
+  | Some Syntax.Mli ->
+      if List.length t.contents = 1 then
+        Fmt.pf ppf "%s" (String.concat "\n" t.contents)
+      else Fmt.pf ppf "\n%a" (pp_lines syntax t) (List.map lstrip t.contents)
+  | Some Cram | Some Normal | None ->
+      Fmt.pf ppf "%a\n" (pp_lines syntax t) t.contents
 
 let pp_errors ppf t =
   match t.value with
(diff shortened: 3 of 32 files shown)


NathanReb:
Thanks for working on this!

Regarding building and testing there's not much to know, `dune build` and `dune runtest` should do the trick. I'm happy to help you further with this dune if you have any further/more detailed questions.

The tests suite is plit between unit tests and end to end tests.

Unit tests live in [`test/lib/`](https://github.com/realworldocaml/mdx/tree/master/test/lib). We use [aclotest](https://github.com/mirage/alcotest) there. If you want to write unit tests you can take inspiration from [`test_block.ml`](https://github.com/realworldocaml/mdx/blob/master/test/lib/test_block.ml) for instance.

The end to end tests uses dune rules to call `mdx` subcommands and compare their output with reference files. They live in [`test/bin/`](https://github.com/realworldocaml/mdx/tree/master/test/bin).
You'll be interested in the [`mdx-test`](https://github.com/realworldocaml/mdx/tree/master/test/bin/mdx-test) subfolder since that's where the test for `ocaml-mdx test` are.
It's divided in three folders: `expect`, `failure` and `misc`. Each of those folder in turn contains one folder per test case. The `expect` one contains tests that ensure that the output of `ocaml-mdx test` on the input markdown file, using the options specified in `test-case.opts` is equal to `test-case.expected` (or to the input if there's no .expected file).
The `failure` is very similar except that it is used to test cases where `ocaml-mdx` will exit with a non-zero status.
Both of these folders dune rules are generated. If you wish to write a test case that don't fit any of these scenarios you can do it in the `misc` folder.

It's likely that you'll have to edit the rule generation a bit so that it works with `.mli` files as well. The shared parts of the rule generation can be found in this [`test/bin/gen_rule_helpers/`](https://github.com/realworldocaml/mdx/tree/master/test/bin/gen_rule_helpers).

I've been planning on adding a CONTRIBUTING.md to document the end to end test, I'll work on that ASAP!

jsomers:
Thanks! I think I've nearly got it building and running tests. I went through [many steps](https://gist.github.com/jsomers/a329588e1d23ba377e6b4ddb516996e3), and ultimately the tests are still exiting with status 1:
```sh
$ dune runtest
   ocaml-mdx test/bin/mdx-test/expect/errors.actual (exit 1)
(cd _build/default/test/bin/mdx-test/expect/errors && ../../../../../../install/default/bin/ocaml-mdx test --output - test-case.md) > _build/default/test/bin/mdx-test/expect/errors.actual
```

Any idea why that might be?

NathanReb:
Oh sorry, should have told you about setting up the dependencies. For future reference, once you have a working opam switch, you can simply run `opam install -t --deps-only .` from the root of the repo and it will install both regular and tests dependencies!

I'll add that to the README!

Also note that it should work with `lwt.5.x`, you might have been confused by lwt's warning!

I'm a bit confused by the tests errors but I'll give it a look!

NathanReb:
Ok so I dug into it a bit and it seems one of the new dependencies you added sets `Printexc.record_backtrace true` and it breaks the tests.

I suspect it comes from `ppxlib` (maybe indirectly) as I remember having to disable it myself there a few month ago, see https://github.com/ocaml-ppx/ppxlib/commit/622e1a30f0f0edddfe93879df78cd6c1ac671a6c#diff-8b1ef87e8b49371d4138edec195db0e5.

We probably want to disable them in mdx as well so my current work around was to do so in ``test/bin/main.ml` but I think we'd rather avoid the ppxlib dependency as it currently makes it hard to stay compatible with new versions of the compiler and requires a lot of maintenance to do so.

I'd suggest relying on minimal parser for mli files that doesn't depend on the AST. How doable do you think that is?

NathanReb:
In the meantime, I pushed the small fixes I had to make to get it to build and pass the tests to my fork, see: https://github.com/NathanReb/mdx/commit/a956cb53b529145bf3c261ca7be073bc02a1af2d

jsomers:
Thanks for those fixes, I've incorporated them into my branch (now pushed here). What's weird is that I had seen the test failures on a commit *before* my changes; I wonder if it has something to do with my environment.

Now, after pulling in your changes (and reintroducing the `ppxlib` dependency) I seem to be doing okay, except that I still get an `(exit 1)` on the tests:

```sh
$ dune runtest
   ocaml-mdx test/bin/mdx-test/expect/errors.actual (exit 1)
(cd _build/default/test/bin/mdx-test/expect/errors && ../../../../../../install/default/bin/ocaml-mdx test --output - test-case.md) > _build/default/test/bin/mdx-test/expect/errors.actual
```

There doesn't seem to be anything in that `errors.actual` file.

> I'd suggest relying on minimal parser for mli files that doesn't depend on the AST. How doable do you think that is?

I'll give that a try. Maybe I can get rid of the ppxlib AST dependency. I'll also try to fix these merge conflicts...

Thanks again!

jsomers:
The current version doesn't use `Ppxlib_ast` anymore, but rather `Ppxlib`. I've also fixed the merge conflicts. The build and tests seem to run okay, though still with that "exit 1" output I don't understand.

jsomers:
Okay, I removed the `ppxlib` dependency. There was no need for it---we can use OCaml's `Lexer` and `Parser` modules directly. This requires renaming `lexer.mll` because otherwise the OCaml `Lexer` module is shadowed by mdx's own lexer.

Locally at least `dune build` and `dune runtest` now run without error.

jsomers:
Okay, last comment from me: I added an end-to-end test along the lines you suggested. Do you think this is now ready for review?

NathanReb:
Yeah it should be! I'll take a look today!

Can you add the odoc dependency to the opam file so that the CI passes?

NathanReb:
Hmm could you also rebase on top of `master` to ease the review?

I think github isn't suited for merge-based workflows and shows a lot of unrelated changes in the diff.

jsomers:
@NathanReb I rebased against master (and squashed all the commits together) to make this easier to read. I also fixed some of the builds I think...

I'll try to add a few more tests.

Thanks @gpetiot for doing the rebase on that separate pull request. Would you mind closing that one now so that there's only one?

lib/mdx.ml
```diff
@@ -73,31 +78,103 @@ let parse l =
...
+    let t' = Block.eval syntax t in
     if t == t' then x else Block t'
 
 type expect_result =
   | Identical
   | Differs
 
+let remove_whitespace str =
+  let open Astring in
+  List.fold_right (fun s acc ->
```
NathanReb: This can be probably be simplified using `Astring`'s `String.filter Char.is_white`.

Could you also move it to `Util` please? That applies to other such functions you might have defined. The point is to avoid accidentally duplicate their code elsewhere if we ever need them.

lib/mdx.ml
```diff
@@ -73,31 +78,103 @@ let parse l =
...
-  let result = if corrected <> file_contents then Differs else Identical in
+  (* The code blocks in mli syntax are often formatted by a tool like ocp-indent or
+     ocamlformat, and it would be impossible to match this formatting exactly down to
+     every space and newline.
+
+     Instead, for this syntax only, we make the diff whitespace-invariant.
+  *)
+  let has_changed = match syntax with
+    | Syntax.Cram | Syntax.Normal -> corrected <> file_contents
+    | Syntax.Mli -> remove_whitespace corrected <> remove_whitespace file_contents
```
NathanReb: I'm a bit worried about removing all whitespaces for the diff here. It should work now because we only support toplevel blocks but I feel like it's easy to forget about this and break mli support if we had support for other blocks in mli files.

We don't necessarily have to fix this now but if we merge it as is we must open an issue as a reminder that we need to imrove that.

In particular, what I have in mind is tied to a refactoring of the tests logic where we would better split I/O from the generation of the updated blocks. Once we have that we can easily diff individual blocks and keep track of any a change there instead of diffing the whole document as a string. cc @gpetiot @voodoos.

lib/lexer_mdx.mll
```diff
@@ -85,5 +85,6 @@ let token syntax lexbuf =
     match syntax with
     | Syntax.Normal -> text      None lexbuf
     | Syntax.Cram   -> cram_text None lexbuf
+    | Syntax.Mli    -> failwith "No lexer here for mli"
```
NathanReb: What do you think about splitting this into `markdown_token` and `cram_token` so that we get rid of the failwith and properly dispatch between the various parsers in `Mdx` where we have all we need to parse every syntax?

mdx.opam
```diff
@@ -14,7 +14,7 @@ build: [
 ]
 
 depends: [
-  "ocaml" {>= "4.02.3"}
+  "ocaml" {>= "4.02.3" & < "4.10.0"}
```
NathanReb: Can you remove the `< 4.10.0` bound? There's nothing in mdx preventing it from working with 4.10. As soon as odoc gets a 4.10 compatible release we will be able to build there as well!

test/bin/mdx-test/expect/simple-mli/test-case.mli
```diff
@@ -0,0 +1,9 @@
+(** This is a doc comment with some code blocks in it:
+
+    {[# List.map (fun x -> x * x) [(1 + 9); 2; 3]
+- : int list = [100; 4; 9]
```
NathanReb: Looking at the test I'm wondering if it would hurt to have mdx respect the indentation or at least just indent everything to the block start level so that mdx users that don't use ocamlformat or don't have ocp-indent checked in CI can still reasonably use that feature.

That's still probably a lot of people so I'm kind of considering it. How much work do you think that would require?

lib/mdx.ml
```diff
@@ -73,31 +78,103 @@ let parse l =
       | `Block b   -> Block b
     ) l
 
+let parse_mli file_contents =
```
NathanReb: This could probably go `Mli_parser`.

It would be nice to have some unit tests for this btw, maybe for an intermediate step since `Block.t` are a bit heavyweight.

bin/test/main.ml
```diff
@@ -393,6 +393,7 @@ let report_error_in_block block msg =
 let run setup non_deterministic not_verbose syntax silent verbose_findlib
     prelude prelude_str file section root direction force_output output : int =
     try
+      Printexc.record_backtrace false;
```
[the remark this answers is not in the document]
jsomers: It looks like we do still need it, sadly -- at least the test/bin/mdx-test/expect/prelude-file/test-case.md test fails, printing this extra bit:

> Called from file "toplevel/toploop.ml", line 208, characters 17-27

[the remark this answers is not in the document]
NathanReb: Ok, in that case I think it would be nice to disable backtraces by default but to expose a cli flag so that it can be controlled by users if they want to.

What do you think?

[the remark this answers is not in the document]
jsomers: Sounds good, I will add the flag!

lib/mdx.ml
jsomers: Sounds good, done! (Will push shortly.)

lib/mdx.ml
jsomers: This is important enough to our use case here that I think I will aim to leave it in -- and then address later along the lines you suggest.

mdx.opam
jsomers: True! Done

test/bin/mdx-test/expect/simple-mli/test-case.mli
jsomers: I've done this because I agree it's better for those users. It wasn't a huge amount of work, just a lot of fiddly changes to get it right.

lib/mdx.ml
jsomers: :+1: Moved it out and wrote some unit tests

lib/mdx.ml
NathanReb: Sounds good, I'll add an issue once we merge!

lib/lexer_mdx.mll
jsomers: I'm not sure I know what you mean. Could you maybe sketch out what you're thinking this code would look like?

jsomers:
@NathanReb Besides the one open question [above](https://github.com/realworldocaml/mdx/pull/206#discussion_r373564367), this should be ready for your re-review. Thanks again!

lib/lexer_mdx.mll
NathanReb: What would you think if instead of having a:
```ocaml
val token : Syntax.t -> Lexing.lexbuf -> [...] list
```
we had:
```ocaml
val cram_token : Lexing.lexbuf -> [...] list
val markdown_token : Lexing.lexbuf -> [...] list
```
in `Lexer_mdx` and turned `Mdx.parse_lexbuf` into:
```ocaml
let parse_lexbuf file_contents syntax lexbuf =
  match syntax with
  | Mli -> Mli_parser.parse_mli file_contents
  | Normal -> parse (Lexer_mdx.markdown_token lexbuf)
  | Cram -> parse (Lexer_mdx.cram_token lexbuf)
```

So that we get rid of this unhandled `Mli` syntax in `Lexer_mdx`. We can clean it up further later but I'd rather get rid of that failwith since this `token` function is only used in one place, in which syntax is already known.

jsomers: @NathanReb Ah, that's a good idea -- this is what we do now.

.ocamlformat
```diff
@@ -1,2 +1,2 @@
-version=0.13.0
```
NathanReb: This should still be `0.13.0`!

Sorry I added this comment to my previous review but it was still pending somehow :/

lib/mli_parser.ml
```diff
@@ -0,0 +1,179 @@
...
+       4 --
+       5 ----------
+       6 -------]--
+       7 ----------
+       8 ----------
+
+       The code below handles this multiline case, concatenating the included substrings
+       from lines 2-6 ([lines_to_include]). *)
+      let first_line = List.hd list in
+      let last_line = List.rev list |> List.hd in
```
Julow: This can be computed in a safe way by doing this before the match on `lines_to_include`:

```ocaml
let drop_last lst =
  match List.rev lst with
  | [] -> None
  | last :: rev_tl -> Some (List.rev rev_tl, last)

let drop_first_and_last = function
  | [] -> None
  | first :: tl -> Some (first, drop_last tl)
```

```ocaml
  match drop_first_and_last lines_to_include with
  | None-> ""
  | Some (line, None) -> String.sub line start.column (end_.column - start.column)
  | Some (first_line, Some (stripped, last_line)) ->
```

lib/mli_parser.ml
```diff
@@ -0,0 +1,179 @@
...
+(* drop_first_and_last [1; 2; 3; 4] = [2; 3]. *)
+let drop_first_and_last list = List.tl list |> List.rev |> List.tl |> List.rev
+
+let slice file_contents ~(start : Odoc_model.Location_.point)
+    ~(end_ : Odoc_model.Location_.point) =
+  let lines_to_include =
+    String.split_on_char '\n' file_contents
+    |> List.mapi (fun i line -> (succ i, line))
+    |> List.filter (fun (lnum, _) -> start.line <= lnum && lnum <= end_.line)
+    |> List.map snd
```
Julow: This has a running time of `number of lines * number of comments`, that's `O(N^2)`.

What do you think of splitting the file into an array in `parse_mli` and accessing lines by their line number in the array ?
Another benefit of this is the computation of `end` in `parse_mli` will be far easier and more robust.

jsomers: @Julow good call, I've done this.

lib/mli_parser.ml
jsomers: Quite right -- I've taken your suggestion, which is much better.

bin/cli.ml
```diff
@@ -48,6 +48,12 @@ let silent_eval =
     (fun x -> `Silent_eval x)
     Arg.(value & flag & info [ "silent-eval" ] ~doc)
 
+let record_backtrace =
+  let doc = "Print backtraces when evaluating toplevel phrases." in
+  named
+    (fun x -> `Record_backtrace x)
+    Arg.(value & flag & info [ "record-backtrace" ] ~doc)
```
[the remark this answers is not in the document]
jsomers: For context see https://github.com/realworldocaml/mdx/pull/206#issuecomment-575582487 and https://github.com/realworldocaml/mdx/pull/206#discussion_r373542330.

If there is another way to repair the broken test, I am happy to extract this into another pull request. But if I revert these changes I still get the error.

lib/output.ml
```diff
@@ -23,8 +23,11 @@ let dump ppf = function
   | `Output s -> Fmt.pf ppf "`Output %S" s
   | `Ellipsis -> Fmt.pf ppf "`Ellipsis"
 
-let pp ?(pad = 0) ppf = function
-  | `Output s -> Fmt.pf ppf "%a%s\n" pp_pad pad s
+let pp ?(pad = 0) ?syntax ppf = function
+  | `Output s -> (
+      match syntax with
+      | Some Syntax.Mli -> Fmt.pf ppf "%a%s\n" pp_pad pad s
+      | _ -> Fmt.pf ppf "%a%s\n" pp_pad pad s )
```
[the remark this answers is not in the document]
jsomers: Good call. This allows us to clean up some weaving-through of `?syntax` in a number of places.

lib/mdx.ml
```diff
@@ -74,10 +87,24 @@ type expect_result = Identical | Differs
...
+  let items = parse_lexbuf file_contents syntax lexbuf in
   Log.debug (fun l -> l "run @[%a@]" dump items);
   let corrected = f file_contents items in
-  let result = if corrected <> file_contents then Differs else Identical in
+
+  (* The code blocks in mli syntax are often formatted by a tool like ocp-indent or
+     ocamlformat, and it would be impossible to match this formatting exactly down to
+     every space and newline.
+
+     Instead, for this syntax only, we make the diff whitespace-invariant.
```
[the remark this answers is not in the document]
jsomers: That's fine by me. I did this. I added a dependency on `str` but I assume that's okay, since it's part of the stdlib?

bin/cli.ml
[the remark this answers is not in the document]
NathanReb: @Julow I specifically asked for this change. Because one of our new dependencies has the side effect of enabling backtraces, we have to explicitly disable them in the code.

I want this to be configurable, even if the default is to disable backtraces.

lib/mdx.ml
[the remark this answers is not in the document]
NathanReb: That won't solve everything either as
```ocaml
# let x = "a  b"
```
will be considered equal to
```ocaml
# let x = "a    b"
```
If we want to be accurate we need to account for OCaml strings.

bin/test/main.ml
```diff
@@ -351,11 +359,14 @@ let report_error_in_block block msg =
...
-let run setup non_deterministic silent_eval syntax silent verbose_findlib
-    prelude prelude_str file section root force_output output : int =
+let run setup non_deterministic silent_eval record_backtrace syntax silent
+    verbose_findlib prelude prelude_str file section root force_output output :
+    int =
   try
-    run_exn setup non_deterministic silent_eval syntax silent verbose_findlib
-      prelude prelude_str file section root force_output output
+    ( match record_backtrace with
+    | `Record_backtrace bool -> Printexc.record_backtrace bool );
```
NathanReb: W should probably move that bit to `run_exn`!

lib/util.ml
```diff
@@ -121,3 +123,11 @@ module List = struct
     in
     aux l
 end
+
+module Array = struct
+  let slice t ~from ~to_ =
+    let start_index, length =
+      if from = to_ then (from, 1) else (from, to_ - from + 1)
```
Julow: The if is not necessary. If `from = to_`, `to_ - from + 1` is `1`.

bin/cli.ml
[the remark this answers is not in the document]
Julow: > I want this to be configurable, even if the default is to disable backtraces.

Why is disabling them in code not enough ?

lib/mdx.ml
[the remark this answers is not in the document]
Julow: Indeed we'll have a problem with code. Splitting words only works for text. Code should be compared verbatim.

lib/block.ml
```diff
@@ -117,16 +120,38 @@ let dump ppf ({ file; line; section; labels; contents; value; _ } as b) =
...
+    match syntax with
+    | Some Syntax.Cram -> Fmt.fmt "  %s"
+    | Some Syntax.Mli ->
+        let i = ref (t.column + 2) in
+        let s = ref Fmt.string in
+        while !i > 0 do
+          s := Fmt.prefix (Fmt.unit " ") !s;
+          i := !i - 1
+        done;
+        !s
```
Julow: Is this the equivalent of `Fmt.fmt "%s%s" (String.make (t.column + 2) ' ')` ?

bin/cli.ml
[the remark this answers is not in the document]
NathanReb: Because in some cases you might want to be able to show the backtraces in a toplevel code fragment.

I think it makes sense to add an escape hatch rather than turning them off with no proper way to re-enable them.

Backtraces being off is a good default, I'm just thinking about the other cases here.

NathanReb:
> I found a loop between Mdx and OCamlformat
I thought OCaml format wasn't formatting the doc comments by default. Does that apply even with doc comment formatting off? If so it is indeed quite the issue, although I'm working on a dune feature to properly format files that are being promoted.

Do you confirm that ocamlformat won't mess with the content of the block and that if we get the rest right it should work fine?

bin/cli.ml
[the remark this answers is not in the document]
Julow: I think this shouldn't be included in this PR. My idea is that we should keep the previous behavior (backtraces disabled ?) and open an issue to talk about how to handle them correctly.

Julow:
OCamlformat doesn't format docstrings by default for now, but that's a common used option that we consider turning on by default.

Maybe a fix to

lib/block.ml
jsomers: No :/. if I write:

```ocaml
let pp_lines syntax t =
  let pp =
    match syntax with
    | Some Syntax.Cram -> Fmt.fmt "  %s"
    | Some Syntax.Mli -> Fmt.fmt "%s%s" (String.make (t.column + 2) ' ')
    | _ -> Fmt.string
  in
  Fmt.(list ~sep:(unit "\n") pp)
```
I get the error:
```
File "lib/block.ml", line 131, characters 40-72:
Error: This expression has type string but an expression was expected of type
         Format.formatter
```

Julow: Yes, sorry, the `ppf` argument comes first.
Also, padding with space characters is something `Format` does natively:
```ocaml
fun ppf -> Fmt.fmt "%*s%s" ppf (t.column + 2) ""
```

jsomers:
> Indeed we'll have a problem with code. Splitting words only works for text. Code should be compared verbatim.

> I found a loop between Mdx and OCamlformat

@Julow, I wonder if we might kill two birds with one stone here by keeping the comparison verbatim, except being agnostic about boundary whitespace, i.e., the arrows in `{[<- ... ->]}`. I think that whitespace is the source of most of the problems, and it has little relevance to the code/output.

> Also giving this to Mdx again removes the closing of the comment:

```diff
 (** {[
       1 + 2 
-    ]} *)
+    ]}
```
This is a separate bug that I am currently investigating. Something weird is going on with the calculation of the "remainder" `Text` node at the end of the mli. In tests, strangely, if I make this portion long enough, it is added as a Text node, and if not, it is dropped. I'll pursue it more tomorrow.

Julow:
> @Julow, I wonder if we might kill two birds with one stone here by keeping the comparison verbatim, except being agnostic about boundary whitespace, i.e., the arrows in {[<- ... ->]}. I think that whitespace is the source of most of the problems, and it has little relevance to the code/output.

I think the comparison is not relevant, it is not used when `--force-output` is passed, which should be the typical use case with Dune.
The most important thing to do is to not change the input if possible (keeping the indentation and position of `{[`, `]}`). If this bigger problem is fixed, the comparison function can be much simpler.

lib/block.ml
jsomers: Apologies, but I declare defeat. I have never used the `Fmt` library before and find it hard to understand. I have struggled for many hours with `Fmt` inside mdx. Unless I'm given the exact diff to apply here that will compile, I won't be able to improve this particular bit of code.

jsomers:
@Julow I've fixed the bug where we eliminated the trailing `*)` and there should no longer be contention with ocamlformat on single-line blocks. Now, too, the comparison function is back to just being `String.equal` instead of splitting to words.

I don't think we should continue to block this pull request on fixing contention with external tooling, especially if that is not a default behavior of ocamlformat, and if there are changes coming to dune /mdx to integrate these formatters directly. The cost of fixing these issues inside this pull request is higher than fixing them later (e.g. because of keeping up with merge conflicts, and because it is harder to track the diffs here).

lib/block.ml
Julow: This diff works for me:
```dif
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -128,14 +128,7 @@ let pp_lines syntax t =
   let pp =
     match syntax with
     | Some Syntax.Cram -> Fmt.fmt "  %s"
-    | Some Syntax.Mli ->
-        let i = ref (t.column + 2) in
-        let s = ref Fmt.string in
-        while !i > 0 do
-          s := Fmt.prefix (Fmt.unit " ") !s;
-          i := !i - 1
-        done;
-        !s
+    | Some Syntax.Mli -> fun ppf -> Fmt.fmt "%*s%s" ppf (t.column + 2) ""
     | _ -> Fmt.string
   in
   Fmt.(list ~sep:(unit "\n") pp)
```

test/bin/mdx-test/expect/simple-mli/test-case.mli
```diff
@@ -0,0 +1,20 @@
+(** This is a doc comment with some code blocks in it:
+
+    {[
+      # List.map (fun x -> x * x) [(1 + 9); 2; 3]
+      - : int list = [100; 4; 9]
+    ]}
+
+    {[List.map (fun x -> x * x) [1; 2; 3]]}
```
Julow: Could you add a multi line code block that is not toplevel ? Currently the `pp_lines` function is not tested.

NathanReb:
Wonderful! Thanks for working so hard this. Let's merge!

test/bin/mdx-test/expect/simple-mli/test-case.mli
jsomers: Now that this has merged, I'll address this and your other pending comment in another pull request. Thanks!

jsomers: https://github.com/realworldocaml/mdx/pull/242

lib/block.ml
jsomers: what a dope i am! i get it now. see https://github.com/realworldocaml/mdx/pull/242

PR #207, opened 2020-01-10
File label for `sh` 

### LINKED ISSUE #134: Add implementation of 'file' label for sh blocks
Currently, sh blocks are not using the label 'file' if this one was added.
We should add it's implementation for promotion purposes

voodoos:
Fixes #134

If the label file is present `mdx-test` will include the whole files in the block.
Parts are not supported and CRAM test will not run on this block.

This PR also include some refactoring of the `test_block` function.

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -294,44 +294,6 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     let active =
       active t && Block.version_enabled t && (not (Block.skip t))
     in
-    (*
-    match active, non_deterministic, Block.mode t, Block.value t with
-    | _, _, _, Error _ -> Block.pp ?syntax ppf t
-    | true, _, _, Raw -> 
-      match Block.part t with
-      | None -> 
-        match Block.file t with
-        ...
-    | false, _, _, _ -> Block.pp ?syntax ppf t
-    | true, false, `Non_det `Command, _ -> Block.pp ?syntax ppf t
-    | true, false, `Non_det `Output, Cram { tests; _ } ->
-    | true, false, `Non_det `Output, Toplevel tests ->
-    | true, _, _, OCaml ->
-      match Block.file t with
-      ...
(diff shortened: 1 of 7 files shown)


bin/test/main.ml
```diff
@@ -294,62 +294,63 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
       active t && Block.version_enabled t && (not (Block.skip t))
     in
-    match active, non_deterministic, Block.mode t, Block.value t with
+    let print_block () = Block.pp ?syntax ppf t in
+    match active, 
+          non_deterministic, 
+          Block.file t, 
+          Block.mode t, 
+          Block.value t 
+    with
```
Julow: This match tries to do too much things.

What do you think of:
- Moving the last 2 cases to functions
  The patterns on `Block.file t` can be moved to a nested match
- Do the `active` check with an `if`
  The first `Error _` case is not necessary, it is handled again in the 2 last cases

So this monster-match now only do the non-det checks.

bin/test/main.ml
```diff
@@ -294,62 +294,63 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
-      run_cram_tests ?syntax t ?root ppf temp_file pad tests
-    (* Top-level tests. *)
-    | true, _, _, Toplevel tests ->
-      assert (syntax <> Some Cram);
-      match Block.file t with
-      | Some ml_file ->
-        update_file_or_block ?root ppf file ml_file t direction
-      | None ->
+        ) tests     
+
```
NathanReb: I'm not a huge fan of empty lines in the middle of functions. Could you please remove it?

bin/test/main.ml
NathanReb: We've discussed this offline with @voodoos and decided to focus on the feature and keep the deeper refactoring for later, there's so much that's wrong about this function I'd prefer to keep a clear separation between refactoring and feature addition here.

PR #208, opened 2020-01-15
Allow to set --non-deterministic through an env var

NathanReb:
This is especially helpful if you're invoking mdx through dune rules you wrote or generated yourself. It will also prove useful with the mdx stanza which won't run non-deterministic tests by default.


samoht:
LGTM

PR #212, opened 2020-01-23
New `part-begin` / `part-end` delimiters

### LINKED ISSUE #165: Improve part delimiters
I think there are a few issues with how `mdx` delimits parts:
1. They are not clearly identified as being interpreted by `mdx` and that can be confusing. `[@@@mdx.part ...]` would be better
2. They aren't actual delimiters and its not very user friendly. Separate begin and end markers for a given part would be better than having to introduce `[@@@part 2]` to mark the end of the part 1 even though no part 2 is ever used anywhere.
3. The fact that they are attributes can be slightly confusing as those usually have meaning. One could think they are interpreted by a PPX and expanded into something. Is there any reason why this isn't done using comments instead?

voodoos:
This PR add support for part delimiters defined in comments as a future replacement of the `[@@@part ...]` attributes. Fixes #165 

## Behavior
- Use a single line comment `(* $MDX part-begin=p1 *)` to declare the beginning of a part named `p1
`
- Use `(* $MDX part-end *)` to mark the end of a part
- Every `part-begin` must be followed by a `part-end`

## Known issue
Effort have been made to keep the behavior of legacy attributes unchanged, according to the test suite. However behavior of the `to-ml` direction do differ, but because support for this direction should be removed in the future i chose not to handle these cases and modified the test.

Dropping support for the `to-ml` direction and the old attribute-parts should lead to a nice simplification of the  code which need a lot of ad-hoc behavior to handle these.

## Todo before merging
- [x] Test the changes on some real use-cases such as RWO
- [x] Add more information to the error messages
- [x] Add test cases for the different error cases
- [ ] Add warnings suggesting the use of these new part delimiters before deprecating the old ones ?


NathanReb:
I tried it on RWO and retro-compat works like a charm!

One note though on error messages when you mess up the delimiters:
```
$ dune runtest book/parsing-with-ocamllex-and-menhir/
   ocaml-mdx alias book/parsing-with-ocamllex-and-menhir/runtest (exit 1)
(cd _build/default/book/parsing-with-ocamllex-and-menhir && ../../../install/default/bin/ocaml-mdx test --prelude=prelude.ml --direction=to-md README.md)
Error in the OCaml code block in README.md at line 429:
Found a `part-end` label without a matching `part-begin`.
```
As discussed offline it would be nice to add the name of the file that's being incuded in the `.md` and eventually add a line number in such a case.

Also as discussed offline, the current parts name are only alphanum chars, it would be nice to add `-` and `_`!

gpetiot:
> Effort have been made to keep the behavior of legacy attributes unchanged, according to the test suite. However behavior of the `to-ml` direction do differ, but because support for this direction should be removed in the future i chose not to handle these cases and modified the test.
> 
> Dropping support for the `to-ml` direction and the old attribute-parts should lead to a nice simplification of the code which need a lot of ad-hoc behavior to handle these.

The `to-ml` synchronization has been removed in #214

NathanReb:
Is this ready for review? Except for the conflicts I mean.

voodoos:
I am working right-now on improving the error messaging

lib/top/part.ml
```diff
@@ -40,39 +40,99 @@ let trim_empty_rev l =
...
+      (match part with
+      | Some p ->
+        let msg = Printf.sprintf "File ended before part %s." p in
+        Error (msg, nline)
+      | None -> Ok [make_part ~is_begin_end_part:true part_lines])
     | line ->
-      match Re.exec_opt part_statement_re line with
-      | None -> parse_parts input make_part (line :: lines)
-      | Some groups ->
+      match parse_line line, part with
```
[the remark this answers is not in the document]
voodoos: Agreed, done in b372a77

lib/part.ml
```diff
@@ -0,0 +1,197 @@
...
+
+let rec replace_or_append part_name body = function
+  | p :: tl when String.equal (Part.name p) part_name ->
+    { p with body } :: tl
+  | p :: tl ->
+    p :: replace_or_append part_name body tl
+  | [] ->
+    [{ name = part_name; sep_indent = ""; body }]
+
+let replace file ~part ~lines =
```
NathanReb: Is this still called somewhere? If we removed the `to-ml` direction this is probably deadcode. We can take care of it later but I'm writing it down so I will remember!

voodoos: Oh, and if the `sep_indent` thing is also only used by the `to-ml` direction it will be a very nice thing to cut^^

PR #213, opened 2020-01-27
Fix a bug with sync of ml files in sub directories

NathanReb:
I discovered it when trying to upgrade the mdx verison used by RWO. The fix comes with a regression test!

The bug was never released so I did not update the changelog!


NathanReb:
cc @voodoos

voodoos:
Well catched ! I also agree with your correction.

PR #217, opened 2020-01-30
New binary that extract dependencies from `.md` files

voodoos:
## Brief
Add a new binary that will be used by the future dune mdx stanza to analyze what are the dependencies of a given `md` file.

## Usage : 
`mdx deps file1.md file2.md`

will output in a non-formatted fashion:

```
file1.md: file:firstFileDep.ml dir:firstDirDep/ file:secondFileDep.txt
file2.md: ...
```

## Limitations
Only `.md` files can be used as input and labels `version` and `non-deterministic` are ignored.


bin/deps.ml
```diff
@@ -0,0 +1,40 @@
...
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+let run files =
+  List.iter (fun f ->
+    let doc = Mdx.parse_file
+      Mdx.Normal (* todo infer syntax, works only with md for now *)
```
gpetiot: This should be now possible with #222

bin/deps.ml
```diff
@@ -0,0 +1,40 @@
+(*
+ * Copyright (c) 2018 Ulysse Gérard <ulysse@gtarides.com>
```
Julow:
```suggestion
 * Copyright (c) 2020 Ulysse Gérard <ulysse@gtarides.com>
```

voodoos: 😅

Julow: The same typo in `dep.ml`, `dep.mli`, `deps.mli`

lib/dep.ml
```diff
@@ -0,0 +1,43 @@
...
+type t = File of string | Dir of string
+
+let of_block block =
+  let open Block in
+  match directory block, file block, skip block with
+  | Some d, None, false -> Some (Dir d)
+  | None, Some f, false -> Some (File f)
+  | None, None, false -> None
+  | _, _, true -> None
+  | _ -> assert false
```
Julow: The missing case is `Some d, Some f, false`. It can happen because these values are from labels that can be set together (and they work correctly).

Test case:

> \```ocaml dir=lib,file=block.mli
> \```

If you really need to return one value, the file should do. Note that it's relative to the value of `dir`.

bin/deps.ml
gpetiot: I reverted the changes in #222 so you can modify `Mdx.parse_file` to make it call `Syntax.infer` if no syntax is provided.

bin/deps.ml
```diff
@@ -0,0 +1,44 @@
...
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+let run files =
+  List.iter (fun file ->
+    let syntax = match Mdx.Syntax.infer ~file with
+    | Some s -> s
+    | None -> Mdx.Normal
```
Julow: Shouldn't this command take `--syntax` ? The caller may know better about this than our `Syntax.infer`.
Also, I think we should exit with an error in this case, instead of guessing.

bin/deps.ml
```diff
@@ -0,0 +1,44 @@
...
+open Cmdliner
+
+let arg_files =
+  let doc = "The files of which the dependencies will be listed." in
+  let docv = "FILES" in
+  Arg.(non_empty & pos_all file [] & info [] ~doc ~docv)
+
+let cmd =
+  let doc = "List the dependencies of the input files." in
+  Term.(pure run $ arg_files),
```
Julow: For consistency, we should take `Cli.setup` (eg. for the option `--verbose`). This requires to setup an logger

bin/deps.ml
```diff
@@ -0,0 +1,44 @@
...
+    Printf.printf "%s: %s\n" file deps) files;
+  0
+
+
+open Cmdliner
+
+let arg_files =
+  let doc = "The files of which the dependencies will be listed." in
+  let docv = "FILES" in
+  Arg.(non_empty & pos_all file [] & info [] ~doc ~docv)
```
Julow: We consistently use a trick in this project, which is to wrap `Arg`s with `` named (fun x -> `Files x) ``.
So that the `run` function takes `` (`Files files) `` as argument, to prevent accidental swap of the arguments.

test/bin/misc-test-cases/mdx-deps/deps.md
```diff
@@ -0,0 +1,17 @@
+Toto
+
+```ocaml file=tikitaka.ml
+```
+  |}
+  and lines2 = Mdx.of_string Mdx.Normal
+  {|
```
Julow: What is this?

NathanReb:
> otherwise we can merge :)

Nope we can't. This is meant to be used by dune in the mdx stanza so as long as I haven't tested it, please do not merge.

bin/deps.ml
```diff
@@ -0,0 +1,40 @@
...
+    | Some s, _
+    | None, Some s -> s
+    | None, None -> Printf.eprintf
+      "Fatal error: could not infer syntax from filename %s, use the --syntax \
+      option to specify a syntax.\n" file; exit 1
+    in
+    let doc = Mdx.parse_file
+      syntax
+      file
+    in
```
gpetiot: This should fit one a single line.

lib/document.ml
```diff
@@ -0,0 +1,36 @@
...
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+ type syntax = Syntax.t =
+  | Normal
```
gpetiot: This line is not correctly indented.

lib/document.ml
```diff
@@ -0,0 +1,36 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+ type syntax = Syntax.t =
```
gpetiot: This line is not correctly indented.

lib/dep.mli
```diff
@@ -0,0 +1,25 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+ type t = File of string | Dir of string
```
gpetiot: The indentation is wrong.

bin/deps.ml
```diff
@@ -0,0 +1,40 @@
...
+    let deps = Mdx.Dep.of_lines doc in
+    let deps = List.map Mdx.Dep.to_string deps in
+    let deps = String.concat " " deps in
+    Printf.printf "%s\n" deps;
```
Julow: Now that we can only pass one file at a time, I think it would be better to separate the outputs with newlines instead of spaces.
Unlike spaces, newlines are very uncommon in file names and are expected to break every tools that print or read file names.

Alternatively, since this is meant to be consumed by dune only, we could use a sane format (eg. sexp or json).

NathanReb:
Ok I think we have a statisfying prototype. The only thing I'd like to solve is format, as @Julow suggested we need something that can properly encode file names and that will be easy to parse on dune's side.

We could probably output something like `((file "<filename>") (dir "<dirname>") ...)`, using `Printf.printf "%S"` to escape the filenames properly. That's easy to maintain on our side and should be parsed properly by dune's sexp parser.

The ideal solution would imply dune providing an encoder so that we can easily use it and be sure that whatever version of dune we use it will always know how to parse the output. I'll check with the dune team.

@voodoos @Julow what do you think?

voodoos:
> We could probably output something like `((file "<filename>") (dir "<dirname>") ...)`, using `Printf.printf "%S"` to escape the filenames properly. That's easy to maintain on our side and should be parsed properly by dune's sexp parser.

Using sexpressions seems fine to me.

NathanReb:
I made some adjustments, added a changelog entry and a disclaimer in the command's doc.

Let me know what you think @Julow @voodoos

Once the format is approved on the dune PR I think we can merge and cut the release!

lib/util.ml
```diff
@@ -69,16 +70,41 @@ module Option = struct
...
+        | Atom str ->
+          Buffer.add_string buf (string_of_int (String.length str));
+          Buffer.add_string buf ":";
+          Buffer.add_string buf str
+        | List e ->
+          Buffer.add_char buf '(';
+          ignore (List.map loop e);
+          Buffer.add_char buf ')'
+      in
+      ignore (loop sexp)
```
Julow: This is not the sexp format.
Lists are separated by whitespaces, there is no length prefix and use quotes and backslashes to escape `"abc\"def"`.

lib/util.mli
```diff
@@ -58,3 +58,17 @@ module String : sig
   (** [english_conjonction ["a"; "b"; "c"]] returns ["a, b and c"].
       @raise Invalid_argument when called on the empty list. *)
 end
+
+module Sexp : sig
```
Julow: Would depending on `sexplib` be bad ?
Getting sexp right is more than a dozen of lines.

NathanReb:
This is not sexp but canonical sexp, they are both easier to print and parse. It is not our own weird format but a well defined one and the prefered one on the dune side for machine to machine communication.

Depending on sexplib means depending on base which will hurt compatibility.

Julow:
I missed that, sorry. I didn't know this format.

lib/util.ml
```diff
@@ -69,16 +70,41 @@ module Option = struct
...
-        | Some x -> Some x
-        | None -> aux t
-    in
-    aux l
+module Sexp = struct
+  type t =
+    | Atom of string
+    | List of t list
+
+  let rec equal t t' =
```
voodoos: Isn't this function behaving exactly the same as `(=)` ?

Otherwise your changes look good to me.

NathanReb: It should behave the same but I think it's best practice to redefine them. `Stdlib.(=)` is an external C function. I can't remember why it's best to define your own though, @Julow might know more about that!

Julow: `=` is a bit slower indeed. The main reason to avoid it is when there is polymorphism (user should be able to give its equal function), or when there is values that cannot be compared (eg. functions).
In this case, `=` is fine.

NathanReb: I stole this one from `stdune` to be honest. Should I remove it then or is it fine to keep it?

Julow: I think it's fine to keep it too.
It would be better to import the whole module from stdune to ensures that it's working correctly but that would bring a handful of other modules with it.
I'm Ok with the current code.

NathanReb:
Ok, let's merge!

PR #218, opened 2020-01-31
Enforce the naming consistency of `silent_eval`

gpetiot:
A small simplification for `bin/test/main.ml`

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -41,9 +41,9 @@ let section =
   named (fun x -> `Section x)
     Arg.(value & opt (some string) None & info ["section"; "s"] ~doc ~docv:"PAT")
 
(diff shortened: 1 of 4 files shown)


bin/cli.ml
```diff
@@ -41,9 +41,9 @@ let section =
   named (fun x -> `Section x)
     Arg.(value & opt (some string) None & info ["section"; "s"] ~doc ~docv:"PAT")
 
-let not_verbose =
+let verbose =
```
Julow: I think it should still be called `not_verbose`, to be consistent with the document and the value it generates

gpetiot:
I guess it makes more sense to call it `silent_eval` everywhere so we don't lose the meaning of the option.

PR #221, opened 2020-01-31
Check the labels during parsing

gpetiot:
The idea is to untangle the label processing (construction, checking, getters), and make the check during parsing. The whole code looks more sane.

One of the tests now fails: `test/bin/mdx-test/expect/labels/test-case.md` as the error is now showed in stderr instead of copied in the block. (previously it resulted as an Error block, but it can't be the case anymore as it would require to keep the erroneous values of the labels attached to this block, and it's not possible with the now typed labels). Are you okay with this new behavior?


lib/block.ml
```diff
@@ -141,165 +119,43 @@ let pp ?syntax ppf b =
...
+let skip t = List.exists (function Label.Skip -> true | _ -> false) t.labels
 
-let environment t = match get_label t "env" with
-  | None
-  | Some None
-  | Some (Some (`Eq, "default")) -> "default"
-  | Some (Some (`Eq, s)) -> s
-  | Some (Some _) -> Fmt.failwith "invalid `env` label value"
+let environment t =
+  List.fold_left
```
NathanReb: How about implementing all those functions on top of `get_label`?

Eventually adding a `get_label_or ~default` variant to be used by mode and env.

lib/label.ml
```diff
@@ -0,0 +1,140 @@
...
+let requires_value ~label ~value f =
+  match value with
+  | Some (op, v) -> f op v
+  | None -> Fmt.failwith "label `%s` requires a value" label
+
+let requires_eq_value ~label ~value f =
+  match value with
+  | Some (`Eq, v) -> f v
+  | Some (_, _) ->
+    Fmt.failwith "label `%s` requires assignment using the `=` operator" label
```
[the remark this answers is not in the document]
NathanReb: Note that it will also eventually make it easier to report all mdx parsing errors at once if we want it!

gpetiot:
> I suspect that this is just an intermediate step before some further improvements on `Block`. If that's the case then don't put too much effort into fixing my comments there.

Indeed it is, as I think the checking of labels can be done independently from the rest of block data, and it makes a smaller PR, but I will try to address all of your comments.

lib/label.ml
```diff
@@ -0,0 +1,140 @@
...
+  | None -> Fmt.failwith "label `%s` requires a value" label
+
+let of_string s =
+  let label, value =
```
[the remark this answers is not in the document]
gpetiot: Done

lib/label.ml
```diff
@@ -0,0 +1,140 @@
...
+  | Env of string
+  | Skip
+  | Non_det of [`Output | `Command]
+  | Version of [`Eq | `Neq | `Le | `Lt | `Ge | `Gt] * Ocaml_version.t
+  | Require_package of string
+  | Set of string * string
+  | Unset of string
+
+let pp_relation ppf = function
+  | `Eq -> Fmt.string ppf "="
```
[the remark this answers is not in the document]
gpetiot: Done

lib/block.ml
```diff
@@ -17,6 +17,8 @@
 open Astring
 open Result
 
+module List = Compat.List
```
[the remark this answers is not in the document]
gpetiot: I'm in favor of this, but yes probably in a separate PR.

lib/label.ml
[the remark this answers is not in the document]
gpetiot: Good idea, done.

lib/block.ml
gpetiot: I improved a bit these functions, but I'm not sure we could use the same base for `get_label` and `get_label_or` without making their usage too complex

lib/block.ml
```diff
@@ -141,165 +119,43 @@ let pp ?syntax ppf b =
...
-  | `Any   -> Fmt.string ppf "*"
-  | `None   -> Fmt.string ppf "<none>"
-  | `Some v -> dump_string ppf v
-
-let match_label l p = match p, l with
-  | `Any   , Some _ -> true
-  | `None  , None   -> true
-  | `Some p, Some (_, l) -> String.equal p l
-  | _ -> false
+let get_label t f = List.fold_left f None t.labels
```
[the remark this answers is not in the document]
gpetiot: I changed the definition of `find_map`, its easier to use now but the definition is less trivial.

lib/label.ml
```diff
@@ -0,0 +1,202 @@
...
+
+  let raw_parse_1_char op op_char s =
+    match String.index_opt s op_char with
+    | Some i ->
+      let before = String.sub s 0 i in
+      let after = String.sub s (i+1) (String.length s - i - 1) in
+      (before, Some (op, after))
+    | None -> (s, None)
+
+  let raw_parse_2_chars op op_char_1 op_char_2 s =
```
gpetiot: These 2 functions' purpose is to avoid using String.cut and regexps

lib/label.ml
```diff
@@ -0,0 +1,202 @@
...
+    | _, None -> y
+    | x -> x
+
+  let raw_parse s =
+    raw_parse_2_chars Neq '<' '>' s |||
+    raw_parse_2_chars Ge '>' '=' s |||
+    raw_parse_1_char Gt '>' s |||
+    raw_parse_2_chars Le '<' '=' s |||
+    raw_parse_1_char Lt '<' s |||
+    raw_parse_1_char Eq '=' s
```
Julow: I think we should use a regex here. This parsing function looks very hard to maintain and is not right:
`dir=a<>b` is interpreted as `"dir=a", Some (Neq, "b")`

lib/label.ml
```diff
@@ -0,0 +1,202 @@
...
+            \ %S and %S."
+            v label "command" "output"
+        in
+        Error (`Msg msg)
+      | Some _ ->
+        let msg =
+          Format.sprintf
+            "label `%s` requires assignment using the `=` operator" label
+        in
+        Error (`Msg msg)
```
Julow: You could make a `requires_eq` function and implement `requires_eq_value` using `requires_eq` and `requires_value`.

lib/label.ml
```diff
@@ -0,0 +1,202 @@
...
+    | [h] -> single_of_string h
+    | h :: t ->
+      let f acc s =
+        match acc, of_string s with
+        | Ok labels, Ok label -> Ok (label :: labels)
+        | Error msgs, Ok _ -> Error msgs
+        | Ok _, Error msg -> Error [msg]
+        | Error msgs, Error msg -> Error (msg :: msgs)
+      in
+      match List.fold_left f (single_of_string h) t with
```
Julow: I think `Ok []` would work as a starting value and you can drop the `single_of_string` logic.
There is no need to ensure that the list is not empty, it is a valid return value.

lib/label.ml
```diff
@@ -0,0 +1,202 @@
...
+let of_string = function
+  | "" -> Ok []
+  | s ->
+    let single_of_string x =
+      match of_string x with
+      | Ok label -> Ok [label]
+      | Error msg -> Error [msg]
+    in
+    match String.split_on_char ',' s with
+    | [] -> Error [`Msg "split_on_char should not return an empty list"]
```
Julow: Are you afraid that `split_on_char` may return `[]` for a string that is not `""` ?
I think this is not necessary and that the result of `split_on_char` can be given unchecked to the algorithm below, it's not necessary to check for the empty string above.

lib/lexer.mll
```diff
@@ -59,7 +71,7 @@ and cram_text section = parse
   | "<-- non-deterministic" ws* (("command"|"output") as choice) eol
```
Julow: If the user type something else than "command" or "output", the label will be ignored.
Also, I think we should let the `Label` module decide what are the right value.
I suggest exposing a function in `Label` that interpret a label from the "parsed label" representation (eg. `Label.interpret "non-deterministic" (Some (Eq, choice))`).
This will help extending this syntax in the future.

test/lib/test_label.ml
```diff
@@ -0,0 +1,25 @@
...
+  [ make_test ~input:"" ~expected:("", None)
+  ; make_test ~input:"foo=" ~expected:("foo", Some (Eq, ""))
+  ; make_test ~input:"foo<>bar" ~expected:("foo", Some (Neq, "bar"))
+  ; make_test ~input:"<bar" ~expected:("", Some (Lt, "bar"))
+  ; make_test ~input:"foo<=bar" ~expected:("foo", Some (Le, "bar"))
+  ; make_test ~input:"foo>bar" ~expected:("foo", Some (Gt, "bar"))
+  ; make_test ~input:">=" ~expected:("", Some (Ge, ""))
+  ]
+
+let suite = ("Label", test_raw_parse)
```
Julow: I think it would be nice to test that labels are interpreted correctly too (eg. testing `Label.of_string`)

lib/label.ml
gpetiot: Good idea, done.

lib/label.ml
gpetiot: It's better indeed.

lib/label.ml
gpetiot: At first I was afraid of returning `Ok [""]` instead of `Ok []` when the input is empty, but it's enough to filter out empty strings after splitting the input string indeed.

lib/lexer.mll
gpetiot: Good idea, done.

lib/label.ml
gpetiot: Done

test/lib/test_label.ml
gpetiot: done

lib/label.ml
```diff
@@ -0,0 +1,195 @@
...
+  let f acc s =
+    let label, value = Relation.raw_parse s in
+    match acc, interpret label value with
+    | Ok labels, Ok label -> Ok (label :: labels)
+    | Error msgs, Ok _ -> Error msgs
+    | Ok _, Error msg -> Error [msg]
+    | Error msgs, Error msg -> Error (msg :: msgs)
+  in
+  let not_empty = function "" -> false | _ -> true in
+  let split = String.split_on_char ',' s |> List.filter not_empty in
```
Julow: Filtering allows leading and trailing comas and double comas `",dir=a,,"`. That's not necessarily bad

lib/label.ml
```diff
@@ -0,0 +1,195 @@
...
+            "%S is not a valid value for label `%s`. Valid values are <none>,\
+            \ %S and %S."
+            v label "command" "output"
+        in
+        Error (`Msg msg)
+      | Some _ ->
+        let msg =
+          Format.sprintf
+            "label `%s` requires assignment using the `=` operator." label
+        in
```
Julow: You should use `requires_eq` in the `Some` case.

gpetiot: Good catch, done.

gpetiot:
@NathanReb do you want to review one more time or is it good to go?

test/lib/test_label.ml
```diff
@@ -0,0 +1,56 @@
...
+  ; make_test ~label:"version" ~value:None
+      ~expected:(Error (`Msg "Label `version` requires a value."))
+  ; make_test ~label:"non-deterministic" ~value:(Some (Eq, "bar"))
+      ~expected:(Error (`Msg "\"bar\" is not a valid value for label `non-deterministic`. Valid values are <none>, \"command\" and \"output\"."))
+  ; make_test ~label:"dir" ~value:(Some (Lt, "bar"))
+      ~expected:(Error (`Msg "Label `dir` requires assignment using the `=` operator."))
+  ; make_test ~label:"set-foo" ~value:(Some (Eq, "bar"))
+      ~expected:(Ok (Set ("foo", "bar")))
+  ]
+
```
NathanReb: let's add tests for `of_string` there as well just to avoid breaking it in the future. That would encourage us to add regression tests in the future.

lib/label.ml
```diff
@@ -0,0 +1,191 @@
...
+          match Ocaml_version.of_string v with
+          | Ok v -> Ok (Version (op, v))
+          | Error (`Msg e) ->
+            let msg = Format.sprintf "Invalid `version` label value: %s." e in
+            Error (`Msg msg) ) )
+  (* non-deterministic accepts any value *)
+  | "non-deterministic" -> (
+      match value with
+      | Some (op, value) ->
+        requires_eq ~label ~op ~value (function
```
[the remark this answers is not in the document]
gpetiot: I'm reverting e4ad2c2

lib/label.ml
gpetiot: It was not allowed before, I added unit tests for that.

gpetiot:
> It looks much better with the regexp!
> 
> I have a few minor comments but then it's good to go!

I think I addressed everything

lib/label.ml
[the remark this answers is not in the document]
Julow: I think that is a mistake.
- The logic for the error "requires assignment" is duplicated
- The code interpreting that label is not on the same level of abstractions as the others
  (`requires_* ... (fun x -> interpret only the value)`, you have to care about relations and errors)
- `requires_eq` is needed to implement `requires_eq_value` (and it's still here)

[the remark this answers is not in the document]
NathanReb: Non-det is a unique case here so I think it's okay if it differs.

I think the op logic is just `match op with Eq -> Ok .. | _ -> Error ...` so it's ok to duplicate it if it makes the code clearer.

`requires_eq` is not required by `requires_eq_value` it's just an indirection used to share the code between `requires_eq_value` and the non-det case. It can be inlined in `requires_eq_value`.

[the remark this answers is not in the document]
NathanReb: The error formatting code could be shared, we could extract a `label -> value -> (_, string) result` for the wrong value error and a `label -> (_, string) result` for the non-eq op error.

That way we ensure consistent error messages while still clarifying the control flow.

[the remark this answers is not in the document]
NathanReb: cc @gpetiot, what do you think?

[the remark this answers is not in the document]
gpetiot: I thing we should not over-engineer this function (I think it's getting out of hand :D), only to not repeat an error message, I'd better keep it simple

PR #222, opened 2020-02-01
Infer syntax kind when --syntax is not set, and add 'markdown' as an alias to 'normal'

### LINKED ISSUE #220: Infer syntax based on file extension
It would be convenient not to have to set `--syntax` if it can be easily guessed from the extension:
- `.t` is cram
- `.md` is markdown
- `.mli` is well, mli

That will also be necessary if we want to allow multiple files to be passed to mdx commands, it be `test`, `deps` or even `pp`.

gpetiot:
Fix #220

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -12,7 +12,7 @@ let non_deterministic =
 
 let syntax =
   let parse = function
-    | "normal" -> `Ok Mdx.Normal
+    | "markdown" | "normal" -> `Ok Mdx.Normal
     | "cram" -> `Ok Mdx.Cram
     | s -> `Error (Format.sprintf "unrecognized syntax %S" s)
   in
@@ -23,7 +23,9 @@ let syntax =
        | Mdx.Cram -> "cram")
   in
   let syntax = parse, print in
-  let doc = "Which syntax to use. Either 'normal' or 'cram'." in
+  let doc =
+    "Which syntax to use. Either 'markdown' (also 'normal') or 'cram'."
+  in
   named (fun x -> `Syntax x)
     Arg.(value & opt (some syntax) None & info ["syntax"] ~doc ~docv:"SYNTAX")
 
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -11,18 +11,12 @@ let non_deterministic =
     Arg.(value & flag & info ["non-deterministic"; "n"] ~env ~doc)
 
 let syntax =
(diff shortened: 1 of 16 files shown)


lib/syntax.ml
```diff
@@ -1,3 +1,20 @@
...
+
+let pp fs = function
+  | Normal -> Fmt.string fs "normal"
+  | Cram -> Fmt.string fs "cram"
+
+let equal x y = match (x, y) with
+  | Normal, Normal
+  | Cram, Cram -> true
+  | _ -> false
+
```
voodoos: Isn't this the standard behavior of OCaml's `=` operator ?

Julow: Anyway, if this is going to stay, the `false` case should specify all the constructors: `| (Normal | Cram), _ -> false`.
This ensures a compilation error if a constructor is added.

lib/mdx.ml
```diff
@@ -78,7 +79,18 @@ type syntax = Syntax.t =
   | Cram
 
 let parse_lexbuf syntax l = parse (Lexer.token syntax l)
-let parse_file syntax f = parse (Lexer.token syntax (snd (Misc.init f)))
+
+let parse_file ?syntax f =
```
Julow: This argument doesn't need to be optional.
It is never called without. Also, it adds responsibilities to the function, adds dead code and requires a fallback case (`| None -> Normal`). I think this should be handled at a higher level.

gpetiot: This is done for https://github.com/realworldocaml/mdx/pull/217/files#r374035482

Julow: I think it's the responsibility of that code to call infer and to define the fallback case.

gpetiot: Yes we can do that instead

bin/cli.ml
```diff
@@ -12,7 +12,7 @@ let non_deterministic =
 
 let syntax =
   let parse = function
-    | "normal" -> `Ok Mdx.Normal
+    | "markdown" | "normal" -> `Ok Mdx.Normal
     | "cram" -> `Ok Mdx.Cram
```
Julow: What do you think of moving this to `Syntax.of_string : string -> t option` ?
Because you improved and exposed that module, I think it makes sense to also expose that code.

gpetiot: It's a good idea, done.

PR #223, opened 2020-02-06
Add a type for headers instead of handling strings

gpetiot:
This is another stone on the path to checking the consistency of block type, block value, header and labels during parsing. Rebased on top of #221


lib/block.mli
```diff
@@ -14,6 +14,16 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
+(** Code blocks headers. *)
+
+module Header : sig
+  type t = Shell | OCaml | Other of string
```
Julow: The `OCaml` constructor is also in the `value` type below, which is more important, I think.
This is not really an issue but can be slightly confusing.

gpetiot:
> Is it worth it ? The new constructors are used only in `Block.eval`, nowhere else.
> I think `eval` was the right place to match on the header string, it was its main job, no need to abstract it.

I got more commits on top of this PR working towards typed blocks, checking the labels and so on, this PR is just an intermediate step, but it's likely that some of it will be replaced/improved in the next PR. I will close this one until the whole rework is over.

PR #224, opened 2020-02-07
Do not unset INSIDE_DUNE by default

NathanReb:
This can mess with dune pretty bad, in particular in dune's test suite. It can still be set on a block by block basis if it's actually needed.

I tested that on RWO and there's only one block where I had to set it. That block could also be preserved by simply adding an ellipsis `...`.

I still need to see if it fixes the issues I had in the dune's test suite.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,10 @@
 - Add support for OCaml 4.10 (#204, @kit-ty-kate)
 - Infer syntax kind when `--syntax` is not set, and add 'markdown' as an alias to 'normal' (#222, @gpetiot)
 
+#### Changed
+
+- Do not unset `INSIDE_DUNE` when executing shell commands by default (#224, @NathanReb)
+
 #### Fixed
 
 - Fix a bug that could cause `ocaml-mdx test` to crash on some `include` in toplevel code blocks
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -63,7 +63,6 @@ let with_dir root f =
       raise e
 
 let get_env blacklist =
-  let blacklist = "INSIDE_DUNE"::blacklist in
   let env = Array.to_list (Unix.environment ()) in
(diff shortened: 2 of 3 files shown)


samoht:
Can you check that it doesn't break https://github.com/realworldocaml/mdx/pull/132 ?

clecat:
Well, the flag was unset to allow blocks using dune to properly generate the expected outcome:
When running mdx tests from dune for instance, dune sets this flags, which will make every following summoning of dune more quiet, generating less output.
Therefore, when mdx blocks calls dune, the flag is already set, generating a truncated output.

I suppose removing it would not be harmfull, except that it might break unit tests, but there might be some side-effects

NathanReb:
I'm not sure I understand what you mean by checking that it doesn't break #132. Is there a particular revdeps or behaviour you want to test?

Our test suite passes and so does all RWO tests but one, which I fixed by explicitly adding the `unset-INSIDE_DUNE` label although the new output didn't seem completely off.
It is likely that some dune commands in shell blocks might change but I'm not sure it's much of an issue. I'm happy to check our revdeps and see what happens there if that's the concern.

Also I believe that `INSIDE_DUNE` being set will in most cases result in a more stable output and a more mdx/testing friendly behaviour. It can still be disabled explicitly unset if one wants it but I believe making this the default behaviour is not the best, especially now that we have a dune integration.

We could eventually add a label called `dune-cmd` or something similar that would set the environment so that dune commands behave in the most friendly way but I'd rather wait until that prove necessary to do it.

avsm:
I'm happy to modify that one RWO example if this simplifies mdx...

clecat:
I agree with you, it should not be a problem to remove it from the default list

samoht:
> I'm not sure I understand what you mean by checking that it doesn't break #132.

What I mean is that you should check which issue #132 was fixing before undoing it, to understand better the impact of your changes. It's clearer now (thanks @clecat) and I'm fine to make that change.

PR #225, opened 2020-02-11
Fix prelude loading

NathanReb:
WIP


gpetiot:
The problem is that the default environment is `"default"`, it behaves as expected with:

```diff
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -255,7 +255,7 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
           let env, f = Mdx.Prelude.env_and_file p in
           let eval () = eval_raw Block.empty ?root c ~line:0 [f] in
           match env with
-          | None   -> eval ()
+          | None   -> Mdx_top.in_env "default" eval
           | Some e -> Mdx_top.in_env e eval
         ) fs
     | fs, [] ->
@@ -263,7 +263,7 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
           let env, f = Mdx.Prelude.env_and_file p in
           let eval () = eval_raw Block.empty ?root c ~line:0 (Mdx.Util.File.read_lines f) in
           match env with
-          | None   -> eval ()
+          | None   -> Mdx_top.in_env "default" eval
           | Some e -> Mdx_top.in_env e eval
```

or put the default value in `Prelude.env_and_file` or in `Mdx_top.in_env`

NathanReb:
Hmm I think I see, I'll take a look!

NathanReb:
Using `"default"` seems bad as a user could totally want to name an env `"default"` and that would result in a strange behaviour from their perspective.

We could eventually use `"mdx-default-env"` or something with which it's harder to unintentionally collide if it turns out the other alternative is not satisfying.

NathanReb:
Thaks for looking into it!

gpetiot:
Fixes #226 

@NathanReb what do you think?

test/bin/mdx-test/expect/env-and-prelude/test-case.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+```ocaml
+# x + y
+- : int = 11
+# z
+- : int = 0
+```
+
+```ocaml env=a,version<4.08
```
NathanReb: Maybe we can get rid of the two blocks with with version constraints and use an ellipse instead, what do you think?
```
# z
...
Error: Unbound value z
```

test/bin/mdx-test/expect/env-and-prelude/test-case.md
```diff
@@ -0,0 +1,69 @@
...
+
+```ocaml env=b,version>=4.08
+# x + y
+- : int = 33
+# z
+Line 1, characters 1-2:
+Error: Unbound value z
+```
+
+Changing the order of the options on the command line results in weird behaviour.
```
NathanReb: Have you tried to reproduce this?

Might be worth extracting a separate test case for that. It doesn't matter much if we don't fix it right away but it's important to at least keep track of it!

gpetiot: Yes I tried it and I don't have different results when moving the arguments.

NathanReb: Ok, in that case I suggest we remove that last bit from the file!

gpetiot: done

PR #228, opened 2020-02-12
Check the blocks during parsing

### LINKED ISSUE #216: Report inconsistent blocks at parsing time
Not all labels or combination of labels are allowed for every block in mdx. This not consistenlty handled at the moment. In some cases, `ocaml-mdx test` will fail and report these errors, in some others the labels will simply be ignored silently.

It would be nice to always report those errors at parsing time to both make it clearer to users what is allowed and what is not and to centralize those concerns in a single place in the code. Piece of code consuming blocks wouldn't have to ensure they are valid then.

gpetiot:
Fix #216

new checks done when building blocks at parsing time:
- `env` should only be set for toplevel and ocaml blocks
- `non-deterministic` should not be set for for include blocks
- `part` requires a `file` label, and only for OCaml files

Further improvements: (in another PR?)
- ~It can be further simplified if we manage to merge the block types and block values, but we would need a `Raw` bock type~ done
- ~move all labels in the block type records and remove the `labels` field, so far the block-specific labels are already moved to the records~ done

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -252,22 +252,15 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
     match syntax with Some syntax -> Some syntax | None -> Syntax.infer ~file
   in
   let c = Mdx_top.init ~verbose:(not silent_eval) ~silent ~verbose_findlib () in
-  let section =
-    match section with None -> None | Some p -> Some (Re.Perl.compile_pat p)
-  in
-  let active b =
-    match (section, Block.section b) with
-    | None, _ -> true
-    | Some re, None -> Re.execp re ""
-    | Some re, Some s -> Re.execp re (snd s)
-  in
   eval_prelude ?root c prelude prelude_str;
 
   let test_block ~ppf ~temp_file t =
-    let active = active t && Block.version_enabled t && not (Block.skip t) in
     let print_block () = Block.pp ?syntax ppf t in
-    match
-      (active, non_deterministic, Block.file t, Block.mode t, Block.value t)
+    match Block.is_active ?section t,
+          non_deterministic,
+          Block.file t,
+          Block.mode t,
+          Block.value t
     with
     (* Print errors *)
     | _, _, _, _, Error _
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -341,7 +341,9 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
 let report_error_in_block block msg =
   let kind =
     match block.Block.value with
-    | Raw _ | Error _ | Include _ -> ""
+    | Raw _ | Error _ -> ""
+    | Include { file_kind = Fk_ocaml _; _ } -> "OCaml file include "
+    | Include { file_kind = Fk_other _; _ } -> "file include "
     | OCaml _ -> "OCaml "
     | Cram _ -> "cram "
     | Toplevel _ -> "toplevel "
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -229,6 +229,13 @@ let update_file_or_block ?root ppf md_file ml_file block part =
 
 exception Test_block_failure of Block.t * string
 
+let with_non_det ~command ~output ~det non_deterministic = function
+  (* the command is non-deterministic so skip everything *)
+  | Some Label.Nd_command when not non_deterministic -> command ()
+  (* its output is non-deterministic; run it but keep the old output. *)
+  | Some Label.Nd_output when not non_deterministic -> output ()
+  | _ -> det ()
+
 let eval_prelude ?root top prelude prelude_str =
   let aux to_lines p =
     let env, f = Mdx.Prelude.env_and_file p in
@@ -268,32 +275,27 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
               (* By construction, there is no part for non-OCaml blocks *)
               let new_content = read_part file_included None in
               update_block_content ppf t new_content )
-      | OCaml { non_det; _ } -> (
-          match non_det with
-          (* the command is non-deterministic so skip everything *)
-          | Some Nd_command when not non_deterministic -> print_block ()
-          | _ ->
-              assert (syntax <> Some Cram);
-              eval_raw ~block:t ?root c ~line:t.line t.contents;
-              Block.pp ppf t )
-      | Cram { tests; pad; non_det } -> (
-          match non_det with
-          (* the command is non-deterministic so skip everything *)
-          | Some Nd_command when not non_deterministic -> print_block ()
-          (* its output is non-deterministic; run it but keep the old output. *)
-          | Some Nd_output when not non_deterministic ->
+      | OCaml { non_det; _ } ->
+          let det () =
+            assert (syntax <> Some Cram);
+            eval_raw ~block:t ?root c ~line:t.line t.contents;
+            Block.pp ppf t
+          in
+          with_non_det non_deterministic non_det ~command:print_block
+            ~output:det ~det
+      | Cram { tests; pad; non_det } ->
+          with_non_det non_deterministic non_det ~command:print_block
+            ~output:(fun () ->
               print_block ();
               let blacklist = Block.unset_variables t in
               List.iter
                 (fun t -> ignore (run_test ?root blacklist temp_file t))
-                tests
-          | _ -> run_cram_tests ?syntax t ?root ppf temp_file pad tests )
-      | Toplevel { phrases = tests; non_det; _ } -> (
-          match non_det with
-          (* the command is non-deterministic so skip everything *)
-          | Some Nd_command when not non_deterministic -> print_block ()
-          (* its output is non-deterministic; run it but keep the old output. *)
-          | Some Nd_output when not non_deterministic ->
+                tests)
+            ~det:(fun () ->
+              run_cram_tests ?syntax t ?root ppf temp_file pad tests)
+      | Toplevel { phrases = tests; non_det; _ } ->
+          with_non_det non_deterministic non_det ~command:print_block
+            ~output:(fun () ->
               assert (syntax <> Some Cram);
               print_block ();
               List.iter
@@ -304,10 +306,10 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
                       let output = List.map (fun l -> `Output l) e in
                       if Output.equal test.output output then ()
                       else err_eval ~cmd:test.command e)
-                tests
-          | _ ->
+                tests)
+            ~det:(fun () ->
               assert (syntax <> Some Cram);
-              run_toplevel_tests ?root c ppf tests t )
+              run_toplevel_tests ?root c ppf tests t)
     else print_block ()
   in
   let gen_corrected file_contents items =
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -257,47 +257,42 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
   let test_block ~ppf ~temp_file t =
     let print_block () = Block.pp ?syntax ppf t in
     if Block.is_active ?section t then
-      match t with
-      | Include b -> (
-          match b.value with
-          | OCaml | Toplevel _ ->
+      match Block.value t with
+      (* Print errors *)
+      | Error _ -> print_block ()
+      | Raw -> print_block ()
+      | Include { file_included; part_included } -> (
+          match Block.header t with
+          | Some Block.Header.OCaml ->
             assert (syntax <> Some Cram);
-            update_file_or_block ?root ppf file b.file_included t
-              b.part_included
-          | Cram _ | Raw ->
-            let new_content = (read_part b.file_included b.part_included) in
-            update_block_content ppf t new_content
-          | Error _ -> print_block () )
-      | _ ->
-        match Block.value t with
-        (* Print errors *)
-        | Error _ -> print_block ()
-        | Raw -> print_block ()
-        | OCaml -> (
-            match Block.non_det t with
-            (* the command is non-deterministic so skip everything *)
-            | Some Nd_command when not non_deterministic -> print_block ()
-            | _ ->
-              assert (syntax <> Some Cram);
-              eval_raw
-                ~block:t ?root c ~line:(Block.line t) (Block.contents t);
-              Block.pp ppf t
-          )
-        | Cram  { tests; pad } -> (
-            match Block.non_det t with
-            (* the command is non-deterministic so skip everything *)
-            | Some Nd_command when not non_deterministic -> print_block ()
-            (* its output is non-deterministic; run it but keep the old output. *)
-            | Some Nd_output when not non_deterministic ->
-              print_block ();
-              let blacklist = Block.unset_variables t in
-              List.iter
-                (fun t -> ignore (run_test ?root blacklist temp_file t))
-                tests
-            | _ -> run_cram_tests ?syntax t ?root ppf temp_file pad tests
-          )
-        | Toplevel tests ->
-          match Block.non_det t with
+            update_file_or_block ?root ppf file file_included t part_included
+          | _ ->
+            let new_content = (read_part file_included part_included) in
+            update_block_content ppf t new_content )
+      | OCaml { non_det; _ } -> (
+          match non_det with
+          (* the command is non-deterministic so skip everything *)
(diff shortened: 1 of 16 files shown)


lib/block.mli
```diff
@@ -66,7 +118,19 @@ val pp_line_directive: (string * int) Fmt.t
...
+val section: t -> section option
+(** [section t] is [t]'s section. *)
+
+val contents: t -> string list
+
+val value: t -> value
+(** [value t] is [t]'s value. *)
+
+val mode: t -> [`Command | `Output] option
 (** [mode t] is [t]'s mode. *)
```
Julow: `` `Non_det of .. `` was the mode, ``[`Command | `Output]`` was only the argument of that mode.
It's true that we never needed the "mode" concept. This function should be renamed `non_det` or similarly.

lib/block.mli
```diff
@@ -30,26 +40,68 @@ type section = int * string
...
 (** The type for supported code blocks. *)
-type t = {
+type ocaml_block = {
+  line    : int;
+  file    : string;
+  section : section option;
+  labels  : Label.t list;
+  header  : Header.t option;
+  contents: string list;
+  value   : value;
```
Julow: All these values are common to all types of blocks. And they make sense on their own (as the "parsed" block), except maybe `value`.

lib/block.mli
```diff
@@ -30,26 +40,68 @@ type section = int * string
...
+  part_included : string option;
+  (** [part_included] is the part of the file to synchronize with.
+      If lines is not specified synchronize the whole file. *)
+}
+
+type t =
+  | Toplevel of ocaml_block
+  | Shell of shell_block
+  | Include of include_block
+  | OCaml of ocaml_block
```
Julow: What is the advantage of branching here rather than with the `value` field, as before (it's still here btw) ?
What does the `value` field mean now and can it be set to inconsistent values ?

I think we should not work around the `value` type and the way it works, but instead improve it. 
For example:
```diff
-type cram_value = { pad: int; tests: Cram.t list }
+type cram_value = {
+  pad: int;
+  tests: Cram.t list;
+  non_det : [`Output | `Command] option;
+}
+
+type ocaml_value = {
+  non_det : [`Output | `Command] option;
+  env     : string;
+  (** [env] is the name given to the environment where tests are run. *)
+}
+
+type include_value = {
+  file_included : string;
+  (** [file_included] is the name of the file to synchronize with. *)
+  part_included : string option;
+  (** [part_included] is the part of the file to synchronize with.
+      If lines is not specified synchronize the whole file. *)
+}
 
 (** The type for block values. *)
 type value =
   | Raw
-  | OCaml
+  | OCaml of ocaml_value
   | Error of string list
   | Cram of cram_value
   | Toplevel of Toplevel.t list
+  | Include of include_value
```

lib/lexer.mll
```diff
@@ -52,7 +58,7 @@ and cram_text section = parse
         newline lexbuf;
         `Section section :: cram_text (Some section) lexbuf }
   | "  " ([^'\n']* as first_line) eol
-      { let header = Syntax.cram_default_header in
+      { let header = Some Block.Header.Shell in
```
Julow: It feels weird to have hardcoded headers interpreted internally by `mk`. What about having `Block.mk_shell` that would skip checking the header ?
If the header doesn't make sense in this case, it shouldn't be in `Block.t`.

lib/block.mli
gpetiot: A partial reason for this is that the `value` is only computed when the block is evaluated, and not when the block is built (the value is only set to `Raw`), I will have a look if that's possible to move this evaluation during the parsing.

lib/block.mli
gpetiot: Done

lib/block.mli
NathanReb: I think we shouldn't duplicate every single field of the main type but instead have the shared part in a record and the block type specific parts as record arguments of a variant type, as @Julow  suggest although value probably isn't the right name here.

The main reason for this change is that in the final type we can get rid of the unproper `Label.t list` and have fields for each valid label instead.

NathanReb: I do believe we should move `eval` at parsing time indeed!

Julow: If `eval` has to be a separate step, we should let the previous type `t` as it was, as the "parsed blocks". `eval` should return a richer type instead.
Example:
- `value` should only be informations available a parsing, it should not allows dummy values until `eval` is called.
  `t` may not even need a `value` field
- `eval` should return something like the new `type t` you made, without the "parsed block" bits
  Example:
  ```
  type evaluated = OCaml of { env : Env.t } | Toplevel of Toplevel.t list | Include of { file_included : ... } | ...
  ```
  We can pass `t`s and `evaluated`s around when we need without constraints on "did it pass through `eval`" or "is it a valid block"

NathanReb: Yeah to clarify I think we should have two different types, `raw` which can hold inconsistent values and should be fairly similar to current type `t`, and `t` which is the clean type ie a type that doesn't allow inconsistent value.

Rough parsing returns `raw` values which are then turned into `t`. That step is currently called `eval` although it doesn't evaluate anything and should probably be renamed.

Everywhere else in the code, we should use `t` and never `raw`. `raw` is just an intermediate representation.

lib/block.mli
gpetiot: done

lib/block.mli
gpetiot: I improved this, the evaluation is now done during parsing, but I didn't find a better name for the `value` field.

lib/lexer.mll
gpetiot: I moved the header field to blocks where it can't be guessed, but I'm not convinced we need a `mk_shell` function, as `mk` performs a lot of checks and I think it's a clear interface as it is with a single constructor function.

lib/block.ml
```diff
@@ -243,33 +242,129 @@ let line_directive = Fmt.to_to_string pp_line_directive
...
+  let set_variables =
+    List.filter_map
+      (function Label.Set (v, x) -> Some (v, x) | _ -> None)
+      labels
+  in
+  let unset_variables =
+    List.filter_map (function Label.Unset x -> Some x | _ -> None) labels
+  in
+  let open Util.Result.Infix in
+  ( match get_label (function File x -> Some x | _ -> None) labels with
```
NathanReb: How about binding a `file` variable and rewriting this as:
```suggestion
  let file = get_label (function File x -> Some x | _ -> None) labels in
  ( match file with
```

test/bin/mdx-test/failure/in-toplevel/test-case.md.expected
```diff
@@ -1,2 +1,2 @@
-Error in the toplevel code block in test-case.md at line 4:
```
NathanReb: Do you know why this was removed? Is there a good reason for it?

Same goes for the other occurences!

lib/lexer.mll
NathanReb: The header is just a syntax marker for the markdown interpreters that doesn't have a semantic value for MDX. It is used, in addition with other things, such as the syntax of the block content (`$` markers for cram blocks, `#` markers for toplevel blocks, etc...) to infer the type of MDX block we're dealing with. I think it's indeed `mk`'s job to interpret it and we should stick to a minimal parsing here.

bin/test/main.ml
```diff
@@ -234,95 +250,65 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
-        | Toplevel tests ->
-            assert (syntax <> Some Cram);
-            run_toplevel_tests ?root c ppf tests t
-        | Raw | _ -> print_block () )
+    if Block.is_active ?section t then
+      match Block.value t with
+      (* Print errors *)
+      | Error _ -> print_block ()
+      | Raw _ -> print_block ()
+      | Include { file_included; part_included; header } -> (
```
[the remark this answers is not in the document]
gpetiot: Do you mean only for Include blocks or for all block types?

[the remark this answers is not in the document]
NathanReb: Just for file includes here, since the header is still part of the parent record type, we have all the information needed to print it so it's okay to remove it from the `Include` record arg.

As a matter of fact, since parts are only meaning full for OCaml files, how about having a record type that don't allow construction of forbiden values by merging `part` and `header` into a single field that could be `Ocaml of part option | Other`?

bin/test/main.ml
```diff
@@ -234,95 +250,65 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
...
+      | OCaml { non_det; _ } -> (
+          match non_det with
+          (* the command is non-deterministic so skip everything *)
+          | Some Nd_command when not non_deterministic -> print_block ()
+          | _ ->
+              assert (syntax <> Some Cram);
+              eval_raw ~block:t ?root c ~line:t.line t.contents;
+              Block.pp ppf t )
+      | Cram { tests; pad; non_det } -> (
+          match non_det with
```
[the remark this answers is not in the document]
gpetiot: done

lib/block.ml
```diff
@@ -120,52 +163,32 @@ let pp ?syntax ppf b =
...
+let skip t = t.skip
 
 let environment t =
-  get_label_or
-    (function Label.Env e -> Some e | _ -> None)
-    ~default:"default" t
+  match t.value with
+  | OCaml b -> b.env
+  | Toplevel b -> b.env
+  | Cram _ | Error _ | Include _ | Raw _ -> "default"
```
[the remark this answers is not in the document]
gpetiot: done

lib/block.mli
```diff
@@ -33,23 +74,36 @@ type t = {
   line : int;
   file : string;
   section : section option;
+  dir : string option;
```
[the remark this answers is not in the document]
gpetiot: I think it can be done in a subsequent PR

bin/test/main.ml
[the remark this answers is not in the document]
gpetiot: That's something that could be explored in a subsequent PR, I think the diff is already big enough

lib/block.ml
```diff
@@ -14,67 +14,109 @@
...
-    header = None;
-    contents = [];
-    value = Raw;
-  }
-
 let dump_string ppf s = Fmt.pf ppf "%S" s
 
 let dump_section = Fmt.(Dump.pair int string)
 
+let header t =
```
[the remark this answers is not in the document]
gpetiot: This function can be removed but the header field can still be accessed from the value record, we can find a way to improve that in a subsequent PR maybe

test/bin/mdx-test/failure/in-toplevel/test-case.md.expected
gpetiot: This error message depends on the block value, which was previously built during evaluation:
```ocaml
let eval t =
  match t.header with
  | Some ("sh" | "bash") ->
      let value = cram t.contents in
      { t with value }
  | Some "ocaml" -> (
      match guess_ocaml_kind t with
      | `OCaml `Code -> { t with value = OCaml }
      | `OCaml `Toplevel ->
          let value = toplevel ~file:t.file ~line:t.line t.contents in
          { t with value }
      | `Other -> Fmt.failwith "Dead code. todo: remove" )
  | _ -> t
```
which decided it was a toplevel block, but now we see there is a `file` label so we decide it is an Include block, we can do something with the header of include blocks maybe

NathanReb: Could you please include the kind of MDX block (not the header) that the error comes from when it's available?

I think it will help users figure out the origin of such issues, in particular they might have misunderstood how MDX would interpret the block.

lib/block.ml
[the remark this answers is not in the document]
NathanReb: If the function can be removed I'd rather remove it now. We can think whether it's worth hiding the rest later!

lib/block.mli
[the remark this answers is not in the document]
NathanReb: I'd like to know what the current behaviour is before deciding whether this can be left out for now or not.

I'm happy to take a look myself if you prefer!

bin/test/main.ml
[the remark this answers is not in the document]
NathanReb: To be honest I think this falls exactly into the scope of this PR, ie improving the block representation so that errors are dealt with early and we don't have to care about inconsistent values after the parsing is done.

lib/block.mli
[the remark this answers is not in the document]
gpetiot: I've had a quick look and this flag is also used for include blocks (see `update_file_or_block` of bin/test/main)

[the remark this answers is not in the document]
NathanReb: Ok, let's leave it as it is then. Thanks!

test/bin/mdx-test/failure/in-toplevel/test-case.md.expected
NathanReb: To clarif, I think the error should probably say something like `Error in the file include code block ...` or even eventually `Error in the OCaml file include code block ...`

bin/test/main.ml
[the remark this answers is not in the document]
gpetiot: done

test/bin/mdx-test/failure/in-toplevel/test-case.md.expected
gpetiot: done

gpetiot:
Is the commit history easy to read or do you want me to squash everything? (not sure to we can reach a middle ground as a lot of commits are interdependent and can't be moved around)

lib/block.mli
```diff
@@ -33,23 +83,36 @@ type t = {
...
+  required_packages : string list;
   labels : Label.t list;
-  header : string option;
   contents : string list;
+  skip : bool;
+  version_enabled : bool;
+      (** Whether the current OCaml version complies with the block's version. *)
+  set_variables : (string * string) list;
+  unset_variables : string list;
   value : value;
```
Julow: Sorry to bring the debate again but the `value` doesn't need to be in this record.

Some part of the program doesn't need the "value" to be computed (for example `Document` or the `dep` command).
Given the implementation of `Block.mk`, it should be easy to compute it afterwards (from `t`), avoiding unnecessary error checks when they are not necessary (it's only used in `bin/test/main.ml`).

Also, the `value` representation could evolve and doesn't need the restriction of being of being tied to the representation of a "parsed block" or the restriction of making sense in every part of the program.
I think it should be moved to its own module (for example `Evaluated_block`).

gpetiot: Most of what is currently in the value type is not really a result of evaluation, maybe I need to rename it first. The result of evaluation is only the `tests` of `cram_value` and the `phrases` of `toplevel_value`.
Maybe we can try to split those from the rest.
But I'm not sure we can trim more things from the checks of `Block.mk` as the label checks sound like a good thing even if the block doesn't need to be evaluated. Do you agree?

Julow: Some label checks doesn't need to be done if we don't need `value`.
What I mean is that `value` is something that will change as we change `bin/test/main` but the rest of `Block.t` have to make sense for every users of it (every other commands).

NathanReb: I strongly disagree here, the whole point of this PR is not to process invalid mdx document and to unify the error handling. I agree `value` isn't the best suited name but so was `eval`.

Nothing is evaluated, `value` is just the cleaned up block. We don't ever want to deal with improper blocks outside of this module.

Of course the `dep` command need the blocks to be valid. We don't want to add a file or directory dependency if we shouldn't.

lib/top/mdx_top.mli
```diff
@@ -29,4 +29,4 @@ val eval : t -> string list -> (string list, string list) result
 (** [eval t p] evaluates the toplevel phrase [p] (possibly spawning on
     mulitple lines) with the configuration value [t]. *)
 
-val in_env : string -> (unit -> unit) -> unit
+val in_env : string -> (unit -> 'a) -> 'a
```
NathanReb: What is that change for?

gpetiot: That is because I cleaned the "execution in environment" feature, it previously was done at a high level no matter the kind of block, and now it is only done for the kind of blocks that have an environment label, and for the Toplevel blocks I need to match on the result of the execution:
```ocaml
                  match
                    Mdx_top.in_env env (fun () ->
                        eval_test ~block:t ?root c test)
                  with
                  | Ok _ -> ()
                  | Error e ->
                      let output = List.map (fun l -> `Output l) e in
                      if Output.equal test.output output then ()
                      else err_eval ~cmd:test.command e
```

PR #230, opened 2020-02-19
Fix a bug with non-deterministic label printing

NathanReb:
When changing the way we handle the labels we introduced a bug where the user would use the default non-det label:
```
``sh non-deterministic
```
and mdx would discard this information, resolve it to the default `output` value and reprint it as:
```
``sh non-deterministic=output
```
ultimately making the default value unusable since the files would always differ.

We have to be more careful about this kind of breakage in the future, in particular when we break an existing test case we need to take the time to understand whether it is acceptable or not.

### DIFF
--- a/lib/label.ml
+++ b/lib/label.ml
@@ -67,6 +67,12 @@ module Relation = struct
       with Not_found -> (s, None)
 end
 
+type non_det =
+  | Nd_output
+  | Nd_command
+
+let default_non_det = Nd_output
+
 type t =
   | Dir of string
   | Source_tree of string
@@ -74,7 +80,7 @@ type t =
   | Part of string
   | Env of string
   | Skip
-  | Non_det of [`Output | `Command]
+  | Non_det of non_det option
   | Version of Relation.t * Ocaml_version.t
   | Require_package of string
   | Set of string * string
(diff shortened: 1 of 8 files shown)


gpetiot:
`non-deterministic` and `non-deterministic=output` are interpreted the same (and were as well before my changes), I'm not seeing this as a regression, rather as a normalization of the way this label is handled.

NathanReb:
This is a breaking change as when run with dune the `.corrected` file differs whereas they would be similar if using `1.5.0`. It is therefore a regression.

NathanReb:
you can reproduce the bug by checking out my first commit and running `dune runtest`

lib/label.ml
```diff
@@ -67,14 +67,20 @@ module Relation = struct
       with Not_found -> (s, None)
 end
 
+type non_det_mode =
```
Julow: I think this should be `non_det`. The mode is a concept from `Block` and is represented by an other type.

test/bin/mdx-test/expect/non-det-default-preserved/test-case.md
```diff
@@ -0,0 +1,6 @@
+When using the `non-deterministic` label in default mode
+(i.e. without passing a value), mdx should preserve that in the output.
+
+```sh non-deterministic
+$ echo "paf"
```
Julow: `echo $RANDOM` ?

PR #231, opened 2020-02-21
Format ml and dune files

NathanReb:
This finally enables ocamlformat for mdx!

There a couple places that we don't want to format, namely:
- `lib/compat.ml`, `lib/migrate_ast.ml` and `lib/top/compat_top.ml` as they use CPPO syntax. They are skipped thanks to a couple `.ocamlformat-ignore` files.
- The end-to-end test cases which sometime contains ml files. These ones are skipped by adding a `.ocamlformat` with `disable=true` at the root of the test case dir.


.ocamlformat
```diff
@@ -0,0 +1,2 @@
+version=0.13.0
+profile=conventional
```
emillon: you can even leave this line out as it's now the default

lib/label.ml
```diff
@@ -43,33 +43,36 @@ module Relation = struct
     | "<=" -> Le
     | "<" -> Lt
     | "=" -> Eq
-    | _ -> assert false (* can not happen, filtered by the regexp *)
+    | _ -> assert false
```
emillon: a bit unfortunate but it should do the right thing if you put the comment before assert

lib/library.ml
```diff
@@ -16,47 +16,50 @@
...
-  | [""] | [""; _] | [_; ""] -> invalid ()
-  | [base_name] -> Ok {base_name; sub_lib = None}
-  | base_name::sl -> Ok {base_name; sub_lib = Some (String.concat "." sl)}
-  | [] -> assert false (* String.cuts invariant *)
+  | [ "" ] | [ ""; _ ] | [ _; "" ] -> invalid ()
+  | [ base_name ] -> Ok { base_name; sub_lib = None }
+  | base_name :: sl -> Ok { base_name; sub_lib = Some (String.concat "." sl) }
+  | [] -> assert false
+
+(* String.cuts invariant *)
```
emillon: same here the comment can be attached if you move it before assert

.ocamlformat
NathanReb: Doesn't it make more sense to explicitly declare it? If the default changes this won't break.

I don't really mind, in particular since we set the version so it's unlikely that anything would break anyway!

lib/label.ml
NathanReb: Good catch, I didn't really review the whole diff, I'll fix that!

.ocamlformat
gpetiot: Agreed it doesn't hurt to explicitly set it to `conventional` (I do the same), I think we can also set `parse-docstrings=true` as it is not the default yet.

NathanReb:
There is a weird failure with `4.02` :/

NathanReb:
I pushed a few changes in separate commits, let me know what you think and let's merge!

PR #232, opened 2020-02-21
Simplify --syntax test cases so that they don't randomly fail

NathanReb:
This test case randomly fails quite often in the CI or locally. I'm not sure why since I can't reproduce it but I'm willing to try anything to fix it and that's a start!

### DIFF
--- a/test/bin/mdx-test/misc/syntax/test-case-cram.ext
+++ b/test/bin/mdx-test/misc/syntax/test-case-cram.ext
@@ -1,11 +1,4 @@
 The syntax of this file is `cram`, it can't be inferred by its file extension, so `--syntax=cram` is required.
 
-  $ cat <<EOF > foo \
-  > hello\
-  > world\
-  > EOF
-  $ cat foo
+  $ echo "hello"
   hello
-  world
-  $ echo foo
-  foo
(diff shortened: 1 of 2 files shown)


NathanReb:
Interesting https://ci.ocamllabs.io/github/realworldocaml/mdx/commit/9f816042b37ab166388aee2c4c7135e6c6e351e0/variant/4.06.

NathanReb:
I removed the newline, it's not what's being tested here anyway

NathanReb:
Ok after giving a better look at the test case it doesn't need to do such things at all so let's just replace the shell block by a simple echo!

NathanReb:
I'm still a bit puzzled as to why they weren't working though

Julow:
I fail to reproduce.

NathanReb:
Can we agree that the test case can be simplified to this and merge?

Julow:
I don't think we should have merged this. This was testing that Mdx supports multiline commands with cram syntax and pass them to the shell as we would expect.
The heredoc is an important and tricky use case.

NathanReb:
No, this was not the point of that specific test. The point of the test was to ensure that the `--syntax` option was working correctly.

See https://github.com/realworldocaml/mdx/commit/56dcec39ebb00b5f9de7d31c96f1791be27b68a1#diff-b2ba33c462caee5b3eedec8cd7315ff6.

Julow:
True, there is an other test for heredocs.

Issue #135, opened 2019-07-09
Combining skip with other labels


NathanReb:
It would be really convenient to be able to add the `skip` label to an `ocaml` or `sh` snippet without having to remove other labels.

For example if I have the following snippet in my `.md` file:
```markdown
```sh dir=./some/dir
$ some-command
Some output
``
```

I'd like to be able to skip it without having to remove the `dir` label, which can be useful for temporarily disable it while debugging for instance. In other word I'd like to be able to write:
```markdown
```sh skip dir=./some/dir
$ some-command
Some output
``
```
and have it just skip that snippet's execution.

Currently what happens is that I get the following diff if I place `skip` before `dir`:
```diff
  ```sh skip dir=./some/dir
+ >> 
+ "skip dir" is not a valid label or prefix. Valid labels are "dir", "source-tree", "file", "part", "env", "skip", "non-deterministic" and "version" and valid prefixes are "set-".
  $ some-command
```

and the following error if I place it after:
```
$ ocaml-mdx test --prelude=prelude.ml --direction=to-ml README.md
ocaml-mdx-test: internal error, uncaught exception:
                (Sys_error
                  "./some/dir skip: No such file or directory")
```

Do you think that would  be possible and would make sense? If not I guess we could at least improve the error reporting a bit to make it clear only one those labels can be used at a time!

I'm of course happy to contribute a fix.

Let me know what you think!

NathanReb:
@clecat pointed that it's just a syntax error and that labels must be separated by a comma, no space like the following:
```
```sh skip,dir=./some/dir
$ some-command
Some output
``
```

Specifying that syntax in some kind of `mdx` manual or to extend the current README would be very helpful as I couldn't find documentation for it although I admit I haven't looked for a very long time!

gpetiot:
The doc was improved by #234

PR #234, opened 2020-03-10
HTML comments can carry block labels 

### LINKED ISSUE #205: Alternative label declaration in markdown comments
Currenlty mdx labels follow the opening of the block like this:
```
```ocaml env=main,...
```

The issue is that not all markdown processing tools will ignore those, for instance pandoc doesn't. I also know @CraigFe has encountered issues of its own with mdx labels.

The idea would be to provide an alternative using markdown comments as defined in [CommonMark](https://spec.commonmark.org/0.29/). It would look something like:
```
<!-- $MDX env=main ... -->
```ocaml
```

That would ultimately completely replace the current form in the long run so that markdown files meant to be consumed by mdx can also be fed to any standard markdown processing tool.

gpetiot:
Fix #205 

I kept the syntax of the `,` separating each label.
But when printing block I chose to only print the labels according to this HTML comment syntax, otherwise we would have to remember what was the original syntax (for each set of label) and keep it somewhere in the block type, which is not very pretty. Hence the huge diff in the tests.

Changelog to be written later, depending on what we agree to keep/change from the current behavior.

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -75,6 +75,7 @@ type t = {
   source_trees : string list;
   required_packages : string list;
   labels : Label.t list;
+  legacy_labels : bool;
   contents : string list;
   skip : bool;
   version_enabled : bool;
@@ -127,9 +128,13 @@ let pp_contents ?syntax ppf t = Fmt.pf ppf "%a\n" (pp_lines syntax) t.contents
 let pp_footer ?syntax ppf () =
   match syntax with Some Syntax.Cram -> () | _ -> Fmt.string ppf "```\n"
 
-let pp_labels ppf = function
+let pp_legacy_labels ppf = function
   | [] -> ()
   | l -> Fmt.pf ppf " %a" Fmt.(list ~sep:(unit ",") Label.pp) l
+
+let pp_labels ppf = function
+  | [] -> ()
+  | l -> Fmt.pf ppf "<!-- $MDX %a -->\n" Fmt.(list ~sep:(unit ",") Label.pp) l
 
 let pp_header ?syntax ppf t =
   match syntax with
@@ -143,9 +148,14 @@ let pp_header ?syntax ppf t =
           Fmt.pf ppf "<-- non-deterministic command\n"
       | _ -> failwith "cannot happen: checked during parsing" )
   | _ ->
-      Fmt.pf ppf "```%a%a\n"
-        Fmt.(option Header.pp)
-        (header t) pp_labels t.labels
+      if t.legacy_labels then
+        Fmt.pf ppf "```%a%a\n"
+          Fmt.(option Header.pp)
+          (header t) pp_legacy_labels t.labels
+      else
+        Fmt.pf ppf "%a```%a\n" pp_labels t.labels
+          Fmt.(option Header.pp)
+          (header t)
 
 let pp ?syntax ppf b =
   pp_header ?syntax ppf b;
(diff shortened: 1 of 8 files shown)


lib/lexer.mll
```diff
@@ -28,6 +29,18 @@ rule text section = parse
             let msg = String.concat ~sep:" " msgs in
             failwith msg
         in
+        let labels_cmt =
+          match label_cmt with
+          | Some label_cmt -> (
+            match Label.of_string label_cmt with
```
voodoos: Maybe we can factorize this `match` ?

lib/lexer.mll
```diff
@@ -18,7 +18,8 @@ rule text section = parse
       { let section = (String.length n, str) in
         newline lexbuf;
         `Section section :: text (Some section) lexbuf }
-  | "```" ([^' ' '\n']* as h) ws* ([^'\n']* as l) eol
+  | ( "<!--" ws* "$MDX" ws* ([^' ' '\n']* as label_cmt) ws* "-->" ws* eol )?
+      "```" ([^' ' '\n']* as h) ws* ([^'\n']* as l) eol
```
voodoos: Should we allow for one or more blank lines between the labels and the comment ?

lib/lexer.mll
gpetiot: Good catch, done.

lib/lexer.mll
gpetiot: Good idea, done.

NathanReb:
Changing the default printing of blocks is a breaking change so we can't get that in. Eventually we will deprecate the inline labels but in the meantime both should be supported and printed as they were parsed.

lib/lexer.mll
NathanReb: Is there a good reason to allow for blank lines? If not I think we should enforce sticking the labels above the block.

If we don't enforce that we have to remember what lives between the labels and the block to reprint it as it was.

gpetiot: That's a good point, I reverted the change, now we can have 0 or 1 linebreak (so no blank lines)

gpetiot:
> Changing the default printing of blocks is a breaking change so we can't get that in. Eventually we will deprecate the inline labels but in the meantime both should be supported and printed as they were parsed.

I now keep the old labels (called legacy_labels in the code)

lib/lexer.mll
NathanReb: Will 0 line break work? I'm guessing it should but that really depends on how external markdown processing tools work. There's no arm in allowing it anyway but we should document the following form:
```
<!-- $MDX <labels> -->
``<header>
<code>
``
```

So that we can forbid it in the future if it turns out to be an issue!

gpetiot: I updated the doc, I also added a precision about labels being separated by a comma, which should be enough to close #135 if you agree

NathanReb: Yep, let's close #135.

That doesn't mean we won't have to start writing proper documentation at some point though!

lib/block.ml
```diff
@@ -75,6 +75,7 @@ type t = {
   source_trees : string list;
   required_packages : string list;
   labels : Label.t list;
+  legacy_labels : Label.t list;
```
NathanReb: How about making this a boolean and not allowing to mix both for the same block?

test/bin/mdx-test/expect/envs/test-case.md
```diff
@@ -60,7 +60,8 @@ type t2 = t1
 type t2 = t1
 ```
 
-```ocaml version<4.08,env=e2
+<!-- $MDX version<4.08 -->
```
NathanReb: Can you leave this test untouched and add a new test that ensure the new labels are supported and properly printed back?

gpetiot: done

lib/block.ml
gpetiot: done

PR #235, opened 2020-03-11
Add a more precise ellipsis accepted anywhere in the output

### LINKED ISSUE #40: More precises ellipsis
This kind of diff is a bit annoying:

```
-|  (cd _build/default && /home/yminsky/.opam/fresh-4.06.1/bin/ocamlc.opt -w -40 -g -bin-annot -I .freq.eobjs -I /home/yminsky/.opam/fresh-4.06.1/lib/base -I /home/yminsky/.opam/fresh-4.06.1/lib/base/caml -I /home/yminsky/.opam/fresh-4.06.1/lib/base/shadow_stdlib -I /home/yminsky/.opam/fresh-4.06.1/lib/sexplib0 -I /home/yminsky/.opam/fresh-4.06.1/lib/stdio -no-alias-deps -o .freq.eobjs/counter.cmo -c -impl counter.ml)
+|  (cd _build/default && /Users/thomas/git/rwo/book/_opam/bin/ocamlc.opt -w @a-4-29-40-41-42-44-45-48-58-59-60-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -bin-annot -I .freq.eobjs -I /Users/thomas/git/rwo/book/_opam/lib/base -I /Users/thomas/git/rwo/book/_opam/lib/base/caml -I /Users/thomas/git/rwo/book/_opam/lib/base/shadow_stdlib -I /Users/thomas/git/rwo/book/_opam/lib/sexplib0 -I /Users/thomas/git/rwo/book/_opam/lib/stdio -intf-suffix .ml -no-alias-deps -opaque -o .freq.eobjs/counter.cmo -c -impl counter.ml)
```

It would be nice if `mdx` could support some way to skip some words (e.g using `<path>` or some similar keywords or notation) instead of the sometimes too generic `...` ellipsis.

gpetiot:
Fix #40

### DIFF
--- a/lib/output.ml
+++ b/lib/output.ml
@@ -17,72 +17,121 @@
 open Misc
 open Compat
 
-type t = [ `Output of string | `Ellipsis ]
+let ellipsis = "..."
 
-let dump ppf = function
-  | `Output s -> Fmt.pf ppf "`Output %S" s
-  | `Ellipsis -> Fmt.pf ppf "`Ellipsis"
-
-let pp ?(pad = 0) ppf = function
-  | `Output s -> Fmt.pf ppf "%a%s\n" pp_pad pad s
-  | `Ellipsis -> Fmt.pf ppf "%a...\n" pp_pad pad
-
-let equals_sub l r start length =
-  let stop = start + length in
-  let rec loop i = i = stop || (l.[i] = r.[i] && loop (succ i)) in
-  loop start
-
-let ellipsis_equal = function
-  | `Output l, `Output r ->
-      let length_r = String.length r in
-      let length_l = String.length l in
-      length_r > 3
-      && r.[length_r - 3] = '.'
-      && r.[length_r - 2] = '.'
-      && r.[length_r - 1] = '.'
-      && length_l > length_r - 4
-      && equals_sub l r 0 (length_r - 3)
-  | _, _ -> false
-
-let equal a b =
-  let rec aux x y =
-    match (x, y) with
-    | [], [] | [ `Ellipsis ], _ | _, [ `Ellipsis ] -> true
-    | (`Ellipsis :: a as x), (_ :: b as y) | (_ :: b as y), (`Ellipsis :: a as x)
-      ->
-        aux x b
-        || (* n+ matches: skip y's head *)
-        aux a y
-        (* 0  match  : skip x's head *)
-    | a :: b, h :: t -> (a = h || ellipsis_equal (a, h)) && aux b t
-    | _ -> false
+(* From jbuilder's stdlib *)
+let ansi_color_strip str =
+  let len = String.length str in
+  let buf = Buffer.create len in
+  let rec loop i =
+    if i = len then Buffer.contents buf
+    else
+      match str.[i] with
+      | '\027' -> skip (i + 1)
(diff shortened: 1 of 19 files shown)


gpetiot:
I added unit tests for the exported functions, I'm not sure we should export more to add tests

lib/output.ml
```diff
@@ -17,34 +17,45 @@
 open Misc
 open Compat
 
-type t = [ `Output of string | `Ellipsis ]
+type sub = S of string | Sub_ellipsis
+
+type t = [ `Output of sub list | `Ellipsis ]
```
NathanReb, line 22: What would you think about having a third variant for raw lines so that we can use `String.equal` for those instead of going through a regexp?

test/lib/test_output.ml
```diff
@@ -0,0 +1,50 @@
...
+      let actual = of_string line in
+      Alcotest.(check (list Testable.sub)) test_name expected actual
+    in
+    (test_name, `Quick, test_fun)
+  in
+  [
+    make_test ~line:"" ~expected:[ S "" ];
+    make_test ~line:"foo" ~expected:[ S "foo" ];
+    make_test ~line:".." ~expected:[ S ".." ];
+    make_test ~line:"..." ~expected:[ S ""; Sub_ellipsis; S "" ];
```
NathanReb, line 16: Any particular reason why this is not simply parsing to `Sub_ellipsis`? I know it's a bit of a corner case since that will actually be recognized as an `Ellipsis` and not an `Output` but I'm more generally suprised by the leading `S ""` whenever the line starts with an ellipsis.

lib/output.mli
```diff
@@ -16,15 +16,19 @@
...
+type t = [ `Output of sub list | `Ellipsis ]
 (** The type for test outputs. Ellipisis ([...]) allow to skip zero,
-   one or many lines while comparing the outputs with {!equal}. *)
+    one or many lines while comparing the outputs with {!equal}.
+    Sub_ellipsis ([<...>]) allow to skip zero, one or more characters or words
+    in e line, while comparing the outputs ith {!equal}. *)
 
-val equal : t list -> t list -> bool
-(** [equal x y] is true iff [x] and [y] are equivalent, modulo
+val equal : t list -> ref:t list -> bool
```
NathanReb, line 27: I'm a bit surprised this compares two `t list` rather than a `string list` to a `t list`, do you happen to know why?

gpetiot: it's because the arguments are swapped when calling the function recursively, I don't know how to implement it differently, so I added new functions operating on `string list` and calling the initial functions

lib/output.ml
gpetiot: I added extra cases in the matches function instead

test/lib/test_output.ml
gpetiot: I added more cases when parsing the string, it's fixed now

gpetiot:
I added more tests and improved the equality tests, this is ready to be reviewed now.

gpetiot:
@NathanReb @pitag-ha what do you think of this one?

gpetiot:
We will work on something else that works better with the dune promote workflow instead!

PR #237, opened 2020-03-12
Add labels to explicitly set block type

### LINKED ISSUE #164: Explicit ocaml block types
I'm a bit confused by all the possible type of ocaml blocks mdx interprets, toplevel, code, code extracted from a file and I'm not sure how it behaves in each cases.

From a user perspective it can be a bit puzzling. I think a first step is to add proper documentation. I'll try to take care of that first part in the next few days but I think it might be worth re-thinking the how these blocks are specified and eventually let the user be explicit about it.

That would probably break mdx's API so we need to think carefully about how to handle all this but I think in the end the `mdx`'s code base would benefit from it as we could get rid of all the "guessing" logic!

gpetiot:
Fix #164 
This adds an intermediary step checking what is guessed vs what is enforced with this new label. Some blocks (ocaml, cram, toplevel) can be forced into a raw block (when it may be invalid on purpose, but the markdown header is set for highlighting)

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -16,6 +16,7 @@
 
 open Result
 open Compat
+open Util.Result.Infix
 
 module Header = struct
   type t = Shell of [ `Sh | `Bash ] | OCaml | Other of string
@@ -287,13 +288,19 @@ let executable_contents ~syntax b =
 
 let version_enabled version =
   let open Util.Result.Infix in
-  Ocaml_version.of_string Sys.ocaml_version >>= fun curr_version ->
+  Ocaml_version.of_string Sys.ocaml_version >>| fun curr_version ->
   match version with
   | Some (op, v) ->
-      Ok (Label.Relation.compare op (Ocaml_version.compare curr_version v) 0)
-  | None -> Ok true
+      Label.Relation.compare op (Ocaml_version.compare curr_version v) 0
+  | None -> true
 
 let get_label f (labels : Label.t list) = Util.List.find_map f labels
+
+let label_not_allowed ~label ~kind =
+  Util.Result.errorf "`%s` label is not allowed for %s blocks." label kind
+
+let label_required ~label ~kind =
+  Util.Result.errorf "`%s` label is required for %s blocks." label kind
 
 let check_not_set msg = function
   | Some _ -> Util.Result.errorf msg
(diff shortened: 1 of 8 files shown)


lib/block.ml
```diff
@@ -365,28 +368,42 @@ let mk ~line ~file ~column ~section ~labels ~legacy_labels ~header ~contents
...
-          Ok (Include { file_included; file_kind }) )
+          Ok (`Include { file_included; file_kind }) )
   | None -> (
       check_not_set "`part` label requires a `file` label." part >>= fun () ->
       match header with
       | Some Header.Shell ->
           check_no_errors errors >>= fun () ->
           check_not_set "`env` label cannot be used with a `shell` header." env
-          >>= fun () -> Ok (Cram { non_det })
+          >>= fun () -> Ok (`Cram { non_det })
```
NathanReb: Any particular reason why we're using polymorphic variants again? If the name collide I think we should just scope them in a submodule or use different names for the kind alone and for the block themselves.

What do you think?

lib/label.ml
```diff
@@ -74,6 +74,8 @@ type non_det = Nd_output | Nd_command
 
 let default_non_det = Nd_output
 
+type block_kind = Raw | OCaml | Cram | Toplevel | Include
```
NathanReb: My impression is that Raw is unecessary as it has no value to mdx, it simply won't be interpreted. I think the skip label already covers that case.

lib/block.ml
```diff
@@ -365,28 +368,42 @@ let mk ~line ~file ~column ~section ~labels ~legacy_labels ~header ~contents
...
+          | `Code -> Ok (`OCaml { env; non_det; errors })
           | `Toplevel ->
               check_no_errors errors >>= fun () ->
-              Ok (Toplevel { env; non_det }) )
-      | _ -> check_no_errors errors >>= fun () -> Ok (Raw { header }) ) )
+              Ok (`Toplevel { env; non_det }) )
+      | _ -> check_no_errors errors >>= fun () -> Ok (`Raw { header }) ) )
+  >>= fun guess ->
+  ( match (guess, block_kind) with
+  | `Include b, (None | Some Include) -> Ok (Include b)
```
NathanReb: I'm not sure that's how it should work. I think the explicit block kind should prevail no matter what.

If no block kind is specified, then we try to infer what it is based on the header, the labels, the syntax etc.

One of the main reason why those are useful is that they allow for clear error reporting and easy to document behaviour.

gpetiot: I wrote it this way because we have to check the block labels whether the block kind is specified or not. Let's say I set the block kind to `toplevel` I have to check all the labels corresponding to the record fields are set (to fill the record), and that none of the incompatible labels are set. That's why I check the labels first, and then check that what is guessed from them is consistent with the block kind if it exists.

NathanReb: I haven't taken a deep enough look but that seems dubious to me. I feel like the `guess` is already the result of the inference. That means that the error will always be the one from inference. Also you if the inferred block differs from the explicit one, the error is always the same. I feel like that's not really the intended behaviour.

Could you add tests to make sure this behaves properly?

lib/label.ml
gpetiot: I didn't think of this, if using `skip` fixes #164 I guess this PR does not have any purpose for now until we make the block type label mandatory

lib/block.ml
gpetiot: Fixed.

lib/block.ml
gpetiot: I redesigned the whole code, I still have to add some tests.

gpetiot:
> E.g. if you set the block to toplevel but use regular ocaml syntax (no leading `#` and proper line continuation) you get an approriate error rather than a silent change of behaviour.

I rewrote the whole PR and added more precise error messages (and some tests about this particular toplevel/ocaml error)

test/bin/mdx-test/expect/empty/test-case.md
```diff
@@ -7,7 +7,7 @@ Testing empty blocks
 ```
 
 <!-- $MDX env=x -->
-```foo
+```ocaml
```
gpetiot, line 10: To avoid having a `env is not allowed for raw blocks` message, which is not the purpose of the test.

lib/block.ml
```diff
@@ -295,105 +296,195 @@ let version_enabled version =
...
+    get_label (function Block_kind x -> Some x | _ -> None) labels
+  in
+  let config = get_label_config labels in
+  ( match block_kind with
+  | Some OCaml -> mk_ocaml ~config ~header ~contents ~errors
+  | Some Cram -> mk_cram ~config ~header ~errors
+  | Some Toplevel -> mk_toplevel ~config ~header ~contents ~errors
+  | Some Include -> mk_include ~config ~header ~errors
+  | None ->
+      mk_ocaml ~config ~header ~contents ~errors
```
NathanReb, line 465: I think this should fallback to the old block type inference we used to have. That should ensure we don't break the old behaviour + will perform better and allow better error messages than sequentially trying to interpret it as any of the block types.

In particular if we change the order here it will change the error message and that's highly confusing to users.

lib/block.ml
```diff
@@ -295,105 +296,195 @@ let version_enabled version =
...
-  | None -> Ok ()
+let label_not_allowed ~label ~kind =
+  Util.Result.errorf "`%s` label is not allowed for %s blocks." label kind
+
+let label_required ~label ~kind =
+  Util.Result.errorf "`%s` label is required for %s blocks." label kind
+
+let fmt_header = Fmt.using (Format.asprintf "`%a`" Header.pp) Fmt.string
+
+let header_not_allowed ~header ~kind =
```
NathanReb, line 307: I wonder if we should really forbid any headers. When the block type is set explicitly, the header shouldn't matter. When the block type isn't set explicitly, we use it to determine the block type so this will never happen.

What do you think?

lib/block.ml
```diff
@@ -295,105 +296,195 @@ let version_enabled version =
...
-  let open Util.Result.Infix in
-  ( match file_inc with
-  | Some file_included -> (
-      check_not_set
-        "`non-deterministic` label cannot be used with a `file` label." non_det
-      >>= fun () ->
-      check_not_set "`env` label cannot be used with a `file` label." env
-      >>= fun () ->
-      check_no_errors errors >>= fun () ->
+type label_config = {
```
NathanReb, line 323: I think using a record here makes much more sense than what we use to have but `label_config` is a bit confusing. How about a simpler `labels` or eventually `block_config`?

lib/block.ml
gpetiot: I think it's better for consistency to forbid headers incompatible for blocks of type T, when said headers are incompatible with T. This way, when we remove the block type constraint the behavior of the block would be the same.

NathanReb: The block header doesn't have an actual semantic value, it's just used for styling and in our case, to infer the block type when that's the only way. They're not incompatible, they just mess with our inference logic. If we set their type explicitly, mdx shouldn't try to reason about them as they are irrelevant to mdx. I feel like it's a bit intrusive.

Imagine I want to apply a custom style to some OCaml blocks and to do so I set the block header to `custom-ocaml` and explicitly set the mdx block type to ocaml. With the current behaviour, mdx will fail to parse the block, telling me I can't use this header for this block type. Even though this scenario is a bit of a fiction and I don't of any such cases, I think it's possible as some md interpreters just use the block header to set a class in the generated html div.
In such a case it would be incredibly frustrating to have it refused by mdx for no substantial reason since the block header prevents in no way mdx from interpreting the block.

lib/block.ml
gpetiot: Agreed, done.

lib/block.ml
gpetiot: I see, good point.

gpetiot: Agreed, it makes more sense, I will simplify this part of the code.

lib/block.ml
```diff
@@ -304,45 +311,108 @@ let check_no_errors = function
...
+          ) )
+  | { file_inc = None; _ } -> label_required ~label:"file" ~kind
+  | { non_det = Some _; _ } ->
+      label_not_allowed ~label:"non-deterministic" ~kind
+  | { env = Some _; _ } -> label_not_allowed ~label:"env" ~kind
+
+let infer_block ~config ~header ~contents ~errors =
+  match config with
+  | { file_inc = Some file_included; non_det; env; part; _ } -> (
       check_not_set
```
NathanReb, line 416: would it make sense to use `mk_include` here or is the error reporting different?

NathanReb: Same remark for the other cases, i.e. once we have infered the type of block, can't use the corresponding `mk_<block-type>` function?

PR #238, opened 2020-03-12
Parse mdx-error blocks to carry errors of OCaml blocks

### LINKED ISSUE #219: Print errors in raw ocaml blocks rather than failing
When the evaluation of a raw OCaml block fails in mdx, the execution stops immediately and the error is reported on stderr.

This does not integrate nicely with the rest of the `mdx test` workflow. I think there's more value in reporting other errors as well. I also believe that since the error will be displayed next to the code and the text it will make more sense to the user and that without any further effort on our side to properly display it.

In addition it would bring raw ocaml blocks behaviour a bit closer to ocaml toplevel blocks which I believe makes sense since they already have very few differences and are meant to be mixed and work together well.

Finally, this could eventually be seen as a feature if one wishes to purposefully write erroneous code to showcase the associated compiler error.

gpetiot:
Fix #219
I retained the `mdx-error` block solution, that is easily extensible to toplevel blocks (#179) if we choose to

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -47,7 +47,7 @@ type cram_value = { non_det : Label.non_det option }
 type ocaml_value = {
   env : Env.t;
   non_det : Label.non_det option;
-  errors : string list;
+  errors : Output.t list;
 }
 
 type toplevel_value = { env : Env.t; non_det : Label.non_det option }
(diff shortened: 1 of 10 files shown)


lib/block.ml
```diff
@@ -125,6 +129,14 @@ let pp_lines syntax =
 
 let pp_contents ?syntax ppf t = Fmt.pf ppf "%a\n" (pp_lines syntax) t.contents
 
+let pp_errors ppf t =
+  match t.value with
+  | OCaml { errors; _ } when List.length errors > 0 ->
+      Fmt.string ppf "```\n";
```
NathanReb: Is there any particular reason why `pp_errors` should be closing the above block?

gpetiot: No particular reason, just that I called `pp_errors` before `pp_footer` but that's not very pretty indeed, will fix that (and squash)

PR #240, opened 2020-03-16
Fix toplevel parsing when phrases contain tabs 

### LINKED ISSUE #179: mdx swallows code on syntax errors, in some cases

gpetiot:
fix #179 
The phrases were not correctly split into Command/Output because of the tabs, there is no point in having a parsing error here. Other errors are correctly printed after the toplevel command they are attached to, but for that we need to correctly split the Command/Output lines.

The change of indentation is obvious when promoting the new block (no warning needs to be emitted).

### DIFF
--- a/lib/lexer_top.mll
+++ b/lib/lexer_top.mll
@@ -12,7 +12,7 @@ rule token = parse
  | _ as c        { failwith (Printf.sprintf "unexpected character '%c'. Did you forget a space after the '#' at the start of the line?" c) }
 
 and phrase acc buf = parse
-  | ("\n"* as nl) "\n  "
+  | ("\n"* as nl) "\n" ws+
       { Lexing.new_line lexbuf;
         let nl = Compat.List.init (String.length nl) (fun _ -> "") in
         phrase (nl @ Buffer.contents buf :: acc) (Buffer.create 8) lexbuf }
--- a/lib/lexer_top.mll
+++ b/lib/lexer_top.mll
@@ -12,7 +12,7 @@ rule token = parse
  | _ as c        { failwith (Printf.sprintf "unexpected character '%c'. Did you forget a space after the '#' at the start of the line?" c) }
 
 and phrase acc buf = parse
-  | ("\n"* as nl) "\n" ws+
+  | ("\n"* as nl) "\n" ("  " | "\t")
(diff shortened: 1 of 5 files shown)


lib/lexer_top.mll
```diff
@@ -12,7 +12,7 @@ rule token = parse
  | _ as c        { failwith (Printf.sprintf "unexpected character '%c'. Did you forget a space after the '#' at the start of the line?" c) }
 
 and phrase acc buf = parse
-  | ("\n"* as nl) "\n  "
+  | ("\n"* as nl) "\n" ("  " | ws ws?)
```
Julow: Shouldn't this be `("  " | "\t")` ?

gpetiot: I wrote it like this because before 4.08 we need the parser to consume exactly 2 spaces to preserve the indentation of the rest of the line, and the `ws ws?` case is when there is a mix of tabs and spaces so we are not able to preserve the indentation in the same way, so we can consume all spaces in this case.

Julow: `ws ws?` can also match a single space character.
There is a rendering problem in my recommendation above, there should be 2 spaces in the first case:
```
("  " | "\t")
```

gpetiot: Looks like a single tabulation is enough indeed, done.

PR #241, opened 2020-03-17
Keep locations from parsing instead of recomputing the lines, providing better error messages

gpetiot:
The idea was to remove the line computation that was taking place in `toplevel.ml` and avoid passing the pair (filename, line) from function to function, we have more precise locations from the parser.

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -18,6 +18,7 @@ open Mdx
 open Compat
 open Result
 open Astring
+open Migrate_ast
 open Mdx.Util.Result.Infix
 
 let src = Logs.Src.create "cram.test"
@@ -84,14 +85,12 @@ let run_test ?root blacklist temp_file t =
   match snd (Unix.waitpid [] pid) with WEXITED n -> n | _ -> 255
 
 let root_dir ?root ?block () =
-  match block with
-  | Some t -> (
-      match Mdx.Block.directory t with
-      | Some d -> (
-          match root with
-          | Some r -> Some (r / d)
-          | None -> Some (Filename.dirname t.file / d) )
-      | None -> root )
+  match (block : Block.t option) with
+  | Some { dir = None; _ } -> root
+  | Some { dir = Some d; loc = { loc_start = { pos_fname; _ }; _ }; _ } -> (
+      match root with
+      | Some r -> Some (r / d)
+      | None -> Some (Filename.dirname pos_fname / d) )
   | None -> root
 
 let resolve_root file dir root =
@@ -123,11 +122,10 @@ let run_cram_tests ?syntax t ?root ppf temp_file pad tests =
     tests;
   Block.pp_footer ?syntax ppf t
 
-let eval_test ?block ?root c test =
-  Log.debug (fun l ->
-      l "eval_test %a" Fmt.(Dump.list (Fmt.fmt "%S")) (Toplevel.command test));
+let eval_test ?block ?root c cmd =
+  Log.debug (fun l -> l "eval_test %a" Fmt.(Dump.list (Fmt.fmt "%S")) cmd);
   let root = root_dir ?root ?block () in
-  with_dir root (fun () -> Mdx_top.eval c (Toplevel.command test))
+  with_dir root (fun () -> Mdx_top.eval c cmd)
 
 let err_eval ~cmd lines =
   Fmt.epr "Got an error while evaluating:\n---\n%a\n---\n%a\n%!"
@@ -137,13 +135,10 @@ let err_eval ~cmd lines =
     lines;
   exit 1
 
-let eval_raw ?block ?root c ~line lines =
-  let test =
-    Toplevel.{ vpad = 0; hpad = 0; line; command = lines; output = [] }
-  in
-  match eval_test ?block ?root c test with
+let eval_raw ?block ?root c cmd =
+  match eval_test ?block ?root c cmd with
   | Ok _ -> ()
-  | Error e -> err_eval ~cmd:lines e
+  | Error e -> err_eval ~cmd e
 
 let split_lines lines =
   let aux acc s =
@@ -153,17 +148,14 @@ let split_lines lines =
   in
   List.fold_left aux [] (List.rev lines)
 
-let eval_ocaml ~block ?syntax ?root c ppf ~line lines errors =
-  let test =
-    Toplevel.{ vpad = 0; hpad = 0; line; command = lines; output = [] }
-  in
+let eval_ocaml ~block ?syntax ?root c ppf cmd errors =
   let update ~errors = function
     | { Block.value = OCaml v; _ } as b ->
         { b with value = OCaml { v with errors } }
     (* [eval_ocaml] only called on OCaml blocks *)
     | _ -> assert false
   in
-  match eval_test ?root ~block c test with
+  match eval_test ?root ~block c cmd with
   | Ok _ -> Block.pp ?syntax ppf (update ~errors:[] block)
   | Error lines ->
       let errors =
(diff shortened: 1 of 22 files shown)


test/bin/mdx-test/failure/in-toplevel/test-case.md.expected
```diff
@@ -1,2 +1,2 @@
-Error in the OCaml file include code block in test-case.md at line 4:
+File "test-case.md", lines 3-4, characters 36-0: Error in the OCaml file include code block
```
gpetiot: Weird character count `0` in the message, no idea where it comes from, since all the locs come from the parser.

gpetiot: Removed the columns from the pretty-printer

bin/test/main.ml
```diff
@@ -136,10 +135,9 @@ let err_eval ~cmd lines =
     lines;
   exit 1
 
-let eval_raw ?block ?root c ~line lines =
-  let test =
-    Toplevel.{ vpad = 0; hpad = 0; line; command = lines; output = [] }
-  in
+let eval_raw ?block ?root c lines =
+  let command = Location.mknoloc lines in
```
NathanReb, line 139: I wonder if we could improve this so we get a loc at least for the prelude files. The prelude_str we can't do much about it unfortunately but I think we'll probably drop support for those in the future so I don't think it matters that much!

From a quick look `eval_raw` is only used in `eval_prelude` so I think we could make it so that `Mdx.Util.File.read_lines` could gather locations as well. If it's too widely used maybe we can extract an `Mdx.Util.File.read_lines_with_locations`. The idea is that we'd build proper locations for the prelude files and no loc for the prelude_str.

What do you think?

test/bin/mdx-test/failure/in-toplevel/test-case.md.expected
```diff
@@ -1,2 +1,2 @@
-Error in the OCaml file include code block in test-case.md at line 4:
+File "test-case.md", lines 3-4: Error in the OCaml file include code block
```
NathanReb, line 1: I had a quick look at the test file here which looks like this:
````markdown
Testing error message

```ocaml file=not_found.ml
# 1 + 1;
```
````

It seems like the closing line isn't counted as part of the block, location-wise. Do you know if this is somehow expected? Do you we could fix it so that the loc also includes that last line?

File: bin/test/main.ml (hunk 2 above)
Comment (gpetiot) [reply bin/test/main.ml]: That's a good idea, I modified the `read_lines` function to return a loc (the computation is a bit ad-hoc, but didn't find a way to do it another way) and added a few unit tests.

File: test/bin/mdx-test/failure/in-toplevel/test-case.md.expected (hunk 3 above)
Comment (gpetiot) [reply test/bin/mdx-test/failure/in-toplevel/test-case.md.expected]: I fixed this, but it also push back the locations in the directive for mdx rule, not sure if this is very important.

File: test/bin/mdx-pp/expect/spaces/test_case.ml.expected [hunk 4]
```diff
@@ -1,4 +1,4 @@
-#5 "test-case.md"
+#11 "test-case.md"
```
NathanReb, line 1: Is this expected?

test/bin/mdx-pp/expect/section-option/test_case.ml.expected
```diff
@@ -1,6 +1,6 @@
-#7 "test-case.md"
+#8 "test-case.md"
```
NathanReb, line 1: I think we offset all locations by 1 line in mdx-pp now.

test/bin/mdx-pp/expect/spaces/test_case.ml.expected
gpetiot: No it's not, but it looks like the whole situation is a mess, the locations are guessed and modified everywhere!

test/bin/mdx-pp/expect/spaces/test_case.ml.expected
```diff
@@ -1,4 +1,4 @@
-#5 "test-case.md"
+#7 "test-case.md"
```
gpetiot, line 1: Looks like a fix as the code starts line 7

test/bin/mdx-pp/expect/section-option/test_case.ml.expected
gpetiot: fixed

test/bin/mdx-pp/expect/spaces/test_case.ml.expected
gpetiot: fixed (but the fix is ugly)

gpetiot:
I think it's okay to not have a changelog entry for this change, the PR is ready for review again!

NathanReb:
> I think it's okay to not have a changelog entry for this change, the PR is ready for review again!

We probably should as it improves the error messages as well, users will be happy to know about it!

lib/toplevel.mli
```diff
@@ -19,8 +19,7 @@
 type t = {
   vpad : int;
   hpad : int;
-  line : int;
-  command : string list;
+  command : string list Location.loc;
```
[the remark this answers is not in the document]
gpetiot: The location is computed in lexer_top, do you think it would be better to not update the location updates in the lexer?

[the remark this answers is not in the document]
Julow: Having a location in the lexer make sense. Since the `loc_end` field is ignored anyway, I think it should be ignored earlier to avoid computing it in other places. Keeping it in the lexer is fine.

lib/util.ml
```diff
@@ -53,14 +53,33 @@ module File = struct
...
+      let loc_end =
+        Lexing.
+          {
+            pos_fname = file;
+            pos_lnum = !pos_lnum;
+            pos_bol = !pos_bol;
+            pos_cnum = !pos_cnum;
+          }
+      in
+      Location.mkloc (List.rev !r) { loc_start; loc_end; loc_ghost = false }
```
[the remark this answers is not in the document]
gpetiot: done

bin/test/main.ml
```diff
@@ -271,7 +267,7 @@ let preludes ~prelude ~prelude_str =
   in
   match (prelude, prelude_str) with
   | [], [] -> []
-  | [], fs -> List.map (aux (fun x -> [ x ])) fs
+  | [], fs -> List.map (aux (fun x -> Location.mknoloc [ x ])) fs
```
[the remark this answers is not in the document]
gpetiot: done

lib/toplevel.mli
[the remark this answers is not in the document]
gpetiot: I simplified pp_line_directive and its calls, it's difficult to do more as we need the locations from the read_lines function as well in other places.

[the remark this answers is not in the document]
Julow: What I meant is that we should have only a `Lexing.position` in `Toplevel.t`. The `loc_end` field should never be computed.

lib/util.ml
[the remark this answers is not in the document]
Julow: I think this function shouldn't not be called from `read_lines` but from where it's useful instead.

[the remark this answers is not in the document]
gpetiot: done

lib/toplevel.mli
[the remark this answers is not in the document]
gpetiot: The whole location is computed in the lexer anyway so we won't gain anything by cutting this field, but I managed to remove a lot of unused parameters here and there that makes the code easier to read.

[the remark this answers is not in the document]
Julow: This was useful to remove computing it in `read_lines`. I see you removed that so I have less strong opinions about it.
Note that the full location is still not used and still makes a little harder to create a `Toplevel.t`.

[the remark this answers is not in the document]
gpetiot: done

lib/mli_parser.mli
```diff
@@ -4,9 +4,5 @@ module Code_block : sig
   type t = { location : Odoc_model.Location_.span; contents : string }
 end
 
-val docstring_code_blocks : string -> Code_block.t list
```
NathanReb, line 7: Do you know why this was exposed?

NathanReb (approved):
Looks good!

lib/mli_parser.mli
gpetiot: It was exposed since its creating in https://github.com/realworldocaml/mdx/pull/206 but was never used out of this module

PR #243, opened 2020-03-21
Adopt new dune rule syntax

ComposMin:
Updating mdx rule command to use the syntax required in newer dune versions (tested with dune 2.4.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,8 @@ To test that the code blocks of `file.md` stay consistent, one can use
 dune's `diff?` stanza:
 
 ```
-(alias
-  (name runtest)
+(rule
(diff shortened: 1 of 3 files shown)


NathanReb:
Thanks for looking into this!

`ocaml-mdx rule` will soon be deprecated and we don't intend in maintaining it anymore. We're not very satistifed with the workflow it offers and decided to work on an integration with dune instead. If you're using dune 2, you should use the `mdx` stanza instead, see the documentation here: https://dune.readthedocs.io/en/latest/dune-files.html#mdx-since-2-4.

We should update the README to mention it, I'll take care of that!

PR #244, opened 2020-04-03
Mdx_top.init can be parameterized by dirs, packages and predicates

gpetiot:
@NathanReb let me know if this is what you had in mind, before I go further (if yes, I still have to re-architecture the lib to make the whole test function customizable by the dirs).

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -15,257 +15,7 @@
  *)
 
 open Mdx
-open Compat
-open Result
-open Astring
 open Migrate_ast
-open Mdx.Util.Result.Infix
-
-let src = Logs.Src.create "cram.test"
-
-module Log = (val Logs.src_log src : Logs.LOG)
-
-let ( / ) = Filename.concat
-
-(* From jbuilder's stdlib *)
-let ansi_color_strip str =
-  let len = String.length str in
-  let buf = Buffer.create len in
-  let rec loop i =
-    if i = len then Buffer.contents buf
-    else
-      match str.[i] with
-      | '\027' -> skip (i + 1)
-      | c ->
-          Buffer.add_char buf c;
-          loop (i + 1)
-  and skip i =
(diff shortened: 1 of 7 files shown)


bin/test/main.ml
```diff
@@ -284,7 +284,12 @@ let run_exn (`Setup ()) (`Non_deterministic non_deterministic)
   let syntax =
     match syntax with Some syntax -> Some syntax | None -> Syntax.infer ~file
   in
-  let c = Mdx_top.init ~verbose:(not silent_eval) ~silent ~verbose_findlib () in
+  let c =
+    Mdx_top.init ~verbose:(not silent_eval) ~silent ~verbose_findlib ~dirs:[]
+      ~packages:
+        [ "unix"; "findlib.top"; "findlib.internal"; "compiler-libs.toplevel" ]
```
NathanReb, line 290: It might be worth extracting this list of packages and the list of directives into some variable somewhere so that it can be reused easily between the mdx binary and the future `mdx.dune_lib` or whatever we'll name it.

gpetiot: Good idea, done.

gpetiot:
I defined a new library `mdx.test`, let me know if that's enough for dune.

gpetiot:
I rebased it on master, @voodoos if there is everything you need for the mdx stanza for dune I think we should merge this one, as it's a bit painful to rebase when there is any change in bin/test/main.ml (which happens often)

voodoos:
I think everything I need is here, as Nathan said, it is mostly the `~dir` parameter which will be useful.

@NathanReb, do you see anything missing ?

NathanReb (approved):
Sure you're right, let's merge this, we can always make changes later if it needs some adjustment. It is already a good change on its own to move the logic from the `bin/test/main.ml` to the library.

Thanks for working on this and sorry the rebase hassle!

PR #248, opened 2020-04-14
Preserve cram block headers

### LINKED ISSUE #247: Support ```bash code fences
MDX currently coerces code fences with language `bash` to have language `sh`. This happens even if `<!-- $MDX skip -->` is applied to the fence:

````diff
 <!-- $MDX skip -->
-```bash
+```sh

 ```
````

It seems fine to me to just preserve the choice that exists in the source file. Minimally, `mdx` should not force a choice when the block is skipped.

gpetiot:
fix #247


gpetiot:
Subsumed by #249

PR #249, opened 2020-04-14
Preseve choice of `bash` versus `sh` in code fences

CraigFe:
This resolves https://github.com/realworldocaml/mdx/issues/247.

- store shell language choice in block headers;
- track header information on cram blocks (if it exists).

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -18,16 +18,18 @@ open Result
 open Compat
 
 module Header = struct
-  type t = Shell | OCaml | Other of string
+  type t = Shell of [ `Sh | `Bash ] | OCaml | Other of string
 
   let pp ppf = function
-    | Shell -> Fmt.string ppf "sh"
+    | Shell `Sh -> Fmt.string ppf "sh"
+    | Shell `Bash -> Fmt.string ppf "bash"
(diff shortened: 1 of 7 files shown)


lib/block.ml
```diff
@@ -42,7 +44,7 @@ end
 
 type section = int * string
 
-type cram_value = { non_det : Label.non_det option }
+type cram_value = { header : Header.t option; non_det : Label.non_det option }
```
gpetiot, line 47: This type can be more precise, as only `Sh` and `Bash` variants are used

CraigFe: True; done.

CraigFe:
> Could you add a block in the test to make sure it behaves appropriately with  as well, as you pointed out in your Issue? Given the implementation it might not make much sense to test it now but it will help prevent breakage if we ever change the way we deal with skipped blocks!

Good idea; done.

Interestingly, MDX seems to insert a blank line in any empty code fence (regardless of language). That could perhaps be considered a separate issue, although it doesn't bother me much.

CraigFe:
Done in https://github.com/realworldocaml/mdx/issues/252.

PR #250, opened 2020-04-14
Generalize ocaml delimiters that are used for parts

### LINKED ISSUE #246: More descriptive error messages for invalid `$MDX` comments
My first attempt at using the `(* $MDX *)` syntax failed in an odd way:

```ocaml
(* $MDX part-begin=main*)
let x = ()
(* $MDX part-end *)
```

which gives the error message:

```text
Error in the OCaml file include code block in test.md at line 5:
In file ./test.ml, line 3: There is no part to end.
```

Turns out you need a space between the label and the comment terminator:

```diff
-(* $MDX part-begin=main*)
+(* $MDX part-begin=main *)
```

I think the unspaced form should probably be supported, but that isn't the main problem here. More generally, MDX will silently ignore comments with labels that it doesn't know about:

```ocaml
(* $MDX foo *)
```

I suggest emitting error messages along the lines of:
- "line n: could not parse `$MDX` comment"
- "line n: unrecognised label in `$MDX` comment: `foo`"

gpetiot:
First draft to fix #246

### DIFF
--- a/lib/ocaml_delimiter.ml
+++ b/lib/ocaml_delimiter.ml
@@ -71,26 +71,29 @@ module Regexp = struct
            opt (str ";;");
            ws;
          ]
+
+  let matches regexp line =
+    match Re.exec_opt regexp line with
+    | Some g ->
+        let indent = Re.Group.get g 1 in
+        let name = Re.Group.get g 2 in
+        let payload = Re.Group.get g 3 in
+        Some (indent, name, payload)
+    | None -> None
 end
 
 let parse line =
   let error () =
     Util.Result.errorf "'%s' is not a valid ocaml delimiter for mdx" line
   in
-  match Re.exec_opt Regexp.attribute line with
-  | Some g -> (
-      let indent = Re.Group.get g 1 in
-      let name = Re.Group.get g 2 in
-      let payload = Re.Group.get g 3 in
+  match Regexp.matches Regexp.attribute line with
+  | Some (indent, name, payload) -> (
       match name with
       | "part" -> Ok (Some (Part_begin (Attr, { indent; payload })))
       | _ -> Ok None )
   | None -> (
-      match Re.exec_opt Regexp.cmt_assign line with
-      | Some g -> (
-          let indent = Re.Group.get g 1 in
-          let name = Re.Group.get g 2 in
-          let payload = Re.Group.get g 3 in
+      match Regexp.matches Regexp.cmt_assign line with
+      | Some (indent, name, payload) -> (
           match name with
           | "part-begin" -> Ok (Some (Part_begin (Cmt, { indent; payload })))
           | _ -> error () )
--- a/lib/ocaml_delimiter.ml
+++ b/lib/ocaml_delimiter.ml
@@ -1,0 +1,100 @@
+(*
+ * Copyright (c) 2018 Thomas Gazagnaire <thomas@gazagnaire.org>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+open Result
+
+type part_begin = { indent : string; payload : string }
+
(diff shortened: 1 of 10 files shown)


lib/ocaml_delimiter.ml
```diff
@@ -0,0 +1,100 @@
...
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *)
+
+open Result
+
+type part_begin = { indent : string; payload : string }
+
+type t = Part_begin of [ `Cmt | `Attr ] * part_begin | Part_end
```
NathanReb, line 21: I think we should declare a type for `Cmt | Attr`, it would also make it a good place to mention that `Attr` is legacy and that we'll drop support for it in `2.x`

lib/ocaml_delimiter.ml
```diff
@@ -0,0 +1,100 @@
...
+  match Re.exec_opt Regexp.attribute line with
+  | Some g -> (
+      let indent = Re.Group.get g 1 in
+      let name = Re.Group.get g 2 in
+      let payload = Re.Group.get g 3 in
+      match name with
+      | "part" -> Ok (Some (Part_begin (`Attr, { indent; payload })))
+      | _ -> error () )
+  | None -> (
+      match Re.exec_opt Regexp.cmt_assign line with
```
NathanReb, line 87: I think it might make sense to extract each of these pattern-matching into a function to clarify the nested pattern-match here a bit.

lib/ocaml_delimiter.ml
```diff
@@ -0,0 +1,100 @@
...
+    Util.Result.errorf "'%s' is not a valid ocaml delimiter for mdx" line
+  in
+  match Re.exec_opt Regexp.attribute line with
+  | Some g -> (
+      let indent = Re.Group.get g 1 in
+      let name = Re.Group.get g 2 in
+      let payload = Re.Group.get g 3 in
+      match name with
+      | "part" -> Ok (Some (Part_begin (`Attr, { indent; payload })))
+      | _ -> error () )
```
NathanReb, line 85: I think this will trigger a warning on a regular attribute and that's not something we want. I think it's okay if we don't have perfect error reporting for attributes anyway since we want to encourage users to move to to the comment form.

gpetiot: Good catch, I fixed it.

lib/ocaml_delimiter.ml
gpetiot: Done.

lib/ocaml_delimiter.ml
gpetiot: Done

lib/ocaml_delimiter.ml
```diff
@@ -0,0 +1,105 @@
...
+        let payload = Re.Group.get g 3 in
+        Some (indent, name, payload)
+    | None -> None
+end
+
+let parse line =
+  let error () =
+    Util.Result.errorf "'%s' is not a valid ocaml delimiter for mdx" line
+  in
+  match Regexp.matches Regexp.attribute line with
```
NathanReb, line 89: Hmm I had something a bit more straightforward in mind.

For instance the parsing of the attribute is pretty simple, it's either a delimiter attribute or not, there can be no error here so we could extract a `parse_attr : string -> t option`.
For the comments it's a bit different but I think it makes sense to have a single function
``parse_cmt : string -> (t option, [`Msg of string]) result`` so that the part that determines whether it's an MDX comment and returns the indent, the name and the payload is shared.

I think this could result in a more robust parser.

From a highlevel point of view it would look like:
```ocaml
match parse_attr line with
| Some _ as delimiter-> Ok delimiter
| None -> parse_cmt line
```

`parse_cmt` would return `None` if we're not looking at an MDX comment and an error if it does start with `(* $MDX` but the pair `name` and `payload` don't fit together or don't make sense to MDX. Does that make sense?

What I have in mind is something like:
```ocaml
let parse_cmt line =
  match Re.exec_opt Regexp.cmt_simple with
  | Some g ->
    match Re.Group.get g 1 with
    |  "part-end" -> Ok (Some Part_end)
    | s when String.starts_with "part-begin=" -> Ok (Some (Part_begin (Cmt, { indent; payload = ...})))
    | _ -> error ()
  | None -> Ok None
```

I took some shortcut here but you get the idea. What would you think about such a solution?
I think it simplifies this bit quite a lot and in the future `parse` will just become `parse_cmt`.

test/lib/test_ocaml_delimiter.ml
```diff
@@ -0,0 +1,33 @@
...
+       silently ignore them and not print a warning here. *)
+    make_test ~line:"(* $MDX part-begin=bar     " ~expected:(Ok None);
+    make_test ~line:"   (* $MDX part-end   *)   " ~expected:(Ok (Some Part_end));
+    make_test ~line:"    [@@@part \"foobar\"]    "
+      ~expected:
+        (Ok (Some (Part_begin (Attr, { indent = "    "; payload = "foobar" }))));
+    make_test ~line:"(* $MDX foo *)"
+      ~expected:
+        (Error (`Msg "'(* $MDX foo *)' is not a valid ocaml delimiter for mdx"));
+    make_test ~line:"[@@@foo \"bar\"]" ~expected:(Ok None);
```
NathanReb, line 30: Might be good to add a test case for `part-end=...`, I feel like this is a confusion that might happen quite often, we might even want to have a specific error for this one.

lib/ocaml_delimiter.ml
gpetiot: It's better indeed, done.

test/lib/test_ocaml_delimiter.ml
gpetiot: Good idea, done.

PR #251, opened 2020-04-15
Deprecate 'mdx rule' starting from 1.7.0

gpetiot:


CHANGES.md
```diff
@@ -12,6 +12,8 @@
 
 #### Deprecated
 
+- The command 'mdx rule' is deprecated and will be removed in 1.8.0 (#251, @gpetiot)
```
NathanReb, line 15: Let's say we'll remove it in `2.0.0`. I'd rather try to follow semver here!

bin/rule.ml
```diff
@@ -266,7 +269,10 @@ let locks =
     Arg.(value & opt (list ~sep:',' string) [] & info [ "locks" ] ~doc ~docv)
 
 let cmd =
-  let doc = "Produce dune rules to synchronize markdown and OCaml files." in
+  let doc =
+    "Produce dune rules to synchronize markdown and OCaml files. mdx rule is \
```
NathanReb, line 273:
```suggestion
    "Produce dune rules to synchronize markdown and OCaml files. $(b, ocaml-mdx rule) is \
```

NathanReb: I think it'll look cleaner like this. Also users might get confused since they invoke it as `ocaml-mdx` now!

gpetiot:
A `Deprecated` module sounds like a good idea, I added this and also changed the version numbers.

NathanReb:
It looks good but can you add a pointer to the mdx stanza in the deprecation warning? I think we should always point users to the replacement when we deprecate a feature.

PR #253, opened 2020-04-20
Avoid adding newlines to empty blocks

### LINKED ISSUE #252: Avoid adding newlines to empty fences
MDX currently inserts newlines into empty code fences under certain circumstances. Interestingly, this doesn't seem to affect unlabelled shell blocks:

````diff
 ```sh
 ```

 ```foo
+
 ```

 <!-- $MDX env=x -->
 ```foo
+
 ```

 <!-- $MDX skip -->
 ```
+
 ```
````

### PR DESCRIPTION
**Pull request by gpetiot**
Fix #252

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -140,14 +140,13 @@ let lstrip string =
   Astring.String.with_index_range string ~first:hpad
 
 let pp_contents ?syntax ppf t =
-  match syntax with
-  | Some Syntax.Mli ->
-      if List.length t.contents = 1 then
-        Fmt.pf ppf "%s" (String.concat "\n" t.contents)
-      else Fmt.pf ppf "\n%a" (pp_lines syntax t) (List.map lstrip t.contents)
-  | Some Cram | Some Normal | None ->
-      if List.length t.contents > 0 then
-        Fmt.pf ppf "%a\n" (pp_lines syntax t) t.contents
+  match (syntax, t.contents) with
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/block.ml [hunk 1]
```diff
@@ -146,7 +146,8 @@ let pp_contents ?syntax ppf t =
         Fmt.pf ppf "%s" (String.concat "\n" t.contents)
       else Fmt.pf ppf "\n%a" (pp_lines syntax t) (List.map lstrip t.contents)
   | Some Cram | Some Normal | None ->
-      Fmt.pf ppf "%a\n" (pp_lines syntax t) t.contents
+      if List.length t.contents > 0 then
```
NathanReb, line 149: How about replacing the `List.length` in this function by pattern matching. Since we only care about the list being a singleton or being non-empty we don't need to walk the entire list.

I'm thinking something like:
```ocaml
let pp_contents ?syntax ppf t =
  match syntax, t.contents with
  | Some Syntax.Mli, [ _ ] ->
      Fmt.pf ppf "%s" (String.concat "\n" t.contents)
  | Some Syntax.Mli, _ ->
      Fmt.pf ppf "\n%a" (pp_lines syntax t) (List.map lstrip t.contents)
  | (Some Cram | Some Normal | None), [] -> ()
  | (Some Cram | Some Normal | None), _ ->
      Fmt.pf ppf "%a\n" (pp_lines syntax t) t.contents
```

gpetiot: It's better indeed, done.

Issue #255, opened 2020-04-21
The README is out of date


shonfeder:
It looks like, at least, the tag syntax and the dune stanza's in the README are out of date. 

It would probably be helpful to put the tutorial material under test, by preprocessing the dune stanzas into the markdown and running the README through mdx itself itself (or something).

(btw, I've had a fantastic time making use of this library! :) )

gpetiot:
Hi, the tag syntax presented in the README (the HTML comments) is the "new" syntax, it's the only one advised in the README. The "old" syntax (in the block header) is still supported for now for legacy users, but may disappear in future versions (and is still used in the tests).

(I haven't followed the recent developments for the dune stanza, it may be out of date indeed)

shonfeder:
Oh interesting. Something must be amiss with my installation then. Perhaps the new tag syntax isn't released? Here's a minimal example:

With 

```sh
$ ocaml-mdx --version
1.6.0
```

In `min-example.md`, I have

    ```sh set-FOO=foo
    $ echo $FOO
    ```

    
    ```sh
    $ echo $BAR
    ```

And the first first sets the env var as expected but the second doesn't:

```sh
$ ocaml-mdx test ./min-example.md && diff min-example.md min-example.md.corrected 
2a3
> foo
7a9
> 
```

shonfeder:
Ah yeah, that's it. I pinned to this repo (instead of installing from opam), and it's now working as documented:

```sh
$ ocaml-mdx test ./min-example.md && diff min-example.md min-example.md.corrected 
2a3
> foo
7a9
> bar
```

NathanReb:
Yeah it's not released yet.

We should be able to cut a release soon though as quite a lot of features and fixes have been merged since the last one. We'll also work on the documentation as I think not all features are documented atm. In particular we should also mention the dune stanza in the README!

PR #257, opened 2020-04-21
Support underscores in enviroment variables

### LINKED ISSUE #254: Environment variable setting breaks on variables with underscore
This should be a minimal reproducing example:

    ```sh set-SOME_VAR=foo
    $ echo 
    ```

Fails with

```
invalid code block: Label `set-SOME_VAR=foo` requires a value.
```

shonfeder:
Closes #254

The last two commits are just tidying up which I assumed would be useful given

1. The explanation in https://github.com/realworldocaml/mdx/issues/255#issuecomment-617164205 that the HTML comment-based notation is now the preferred way to set flags
2. The presence of an `.ocamlformat` file.

Let me know if you'd rather I drop those commits :)

### DIFF
--- a/test/bin/mdx-test/expect/environment-variable-set/test-case.md
+++ b/test/bin/mdx-test/expect/environment-variable-set/test-case.md
@@ -38,3 +38,11 @@ And the variable stays available in subsequent blocks.
   $ echo $BOO
   far
 ```
+
+Environment variables can contain underscores
+
+<!-- $MDX set-TWO_WORDS=success -->
+```sh
+  $ echo $TWO_WORDS
+  success
+```
--- a/lib/label.ml
+++ b/lib/label.ml
@@ -51,7 +51,7 @@ module Relation = struct
     @@ seq
          [
            bos;
-           group (rep (alt [ alnum; char '-' ]));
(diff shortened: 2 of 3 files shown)


shonfeder:
It looks like the .ocamlformat results actually conflict with the lining checks: https://ci.ocamllabs.io/github/realworldocaml/mdx/commit/1c793133b8761095aaf21eb366fa07e60913fbfa/variant/Lint%20formatting

I assume this means I should drop the formatting changes, but I'll wait for guidance :)

test/bin/mdx-test/expect/environment-variable-set/test-case.md
```diff
@@ -21,6 +21,14 @@ And the variable stays available in subsequent blocks.
   - : unit = ()
 ```
 
+Environment variables can contain underscores
+
+```ocaml set-TWO_WORDS=success
```
NathanReb: End to end tests are great but we're also trying to improve the unit testing situation in MDX, would you mind making those unit tests in https://github.com/realworldocaml/mdx/blob/master/test/lib/test_label.ml.

You can probably leave one of those two e2e tests, they definitely don't hurt!

PR #259, opened 2020-05-05
Preserve indentation of included file

### LINKED ISSUE #258: Import blocks messes with indentation
It seems that when importing part of an ocaml file, MDX sometimes messes with the indentation.
Consider the following the code:
```ocaml
let _add_with_let_in t ~key ~data =
  let i = hash_bucket key in
  let replace = bucket_has_key t i key in
  let filtered_bucket =
    if replace then
      List.filter t.buckets.(i) ~f:(fun (key',_) -> Poly.(key' <> key))
    else
      t.buckets.(i)
  in
  (* $MDX part-begin=add-with-let-in *)
  let () = t.buckets.(i) <- (key, data) :: filtered_bucket in
  if not replace then t.length <- t.length + 1
  (* $MDX part-end *)
```

When importing it to a markdown block I get the following:
````markdown
```ocaml file=examples/correct/dictionary/dictionary.ml,part=add-with-let-in
let () = t.buckets.(i) <- (key, data) :: filtered_bucket in
  if not replace then t.length <- t.length + 1
```
````

Note how for the first line there is no indentation whatsoever but for the second line with the if statement, the indent level from the original file is respected.

It would be good if we had a consistent behaviour wrt to such code fragments. That is either preserve the original indentation and remove indentation on all lines according to the indentation of MDX delimiter. In this example I think acceptable outcomes would be either:
````markdown
```ocaml file=examples/correct/dictionary/dictionary.ml,part=add-with-let-in
  let () = t.buckets.(i) <- (key, data) :: filtered_bucket in
  if not replace then t.length <- t.length + 1
```
````
or:
````markdown
```ocaml file=examples/correct/dictionary/dictionary.ml,part=add-with-let-in
let () = t.buckets.(i) <- (key, data) :: filtered_bucket in
if not replace then t.length <- t.length + 1
```
````

### PR DESCRIPTION
**Pull request by gpetiot**
Fix #258 
I chose to keep the indentation as is. Removing it completely is a bad solution (as all lines don't necessarily have the same indentation level). The solution of decreasing the indentation of all lines by the minimum indentation among all lines (not necessary the first) is more complicated.

Also keeping it as-is allows to have a better understanding of how the code is related to other code blocks (maybe more obvious if the indentation is printed as a special character or something)

### DIFF
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -230,7 +230,8 @@ let read_part file part =
         file
   | Some lines ->
       let contents = String.concat ~sep:"\n" lines in
-      String.trim contents ~drop:(function '\n' -> true | _ -> false)
+      String.drop contents ~rev:true ~sat:Char.Ascii.is_white
+      |> String.drop ~sat:(function '\n' -> true | _ -> false)
 
 let write_parts ~force_output file parts =
   let output_file = file ^ ".corrected" in
--- a/bin/test/main.ml
+++ b/bin/test/main.ml
@@ -230,7 +230,7 @@ let read_part file part =
         file
   | Some lines ->
       let contents = String.concat ~sep:"\n" lines in
-      String.trim contents
+      String.trim contents ~drop:(function '\n' -> true | _ -> false)
 
 let write_parts ~force_output file parts =
   let output_file = file ^ ".corrected" in
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,8 @@
 
 - Avoid adding newlines to empty blocks (#253, @gpetiot)
 
+- Preserve the indentation of included files (#259, @gpetiot)
+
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: bin/test/main.ml [hunk 1]
```diff
@@ -230,7 +230,7 @@ let read_part file part =
         file
   | Some lines ->
       let contents = String.concat ~sep:"\n" lines in
-      String.trim contents
+      String.trim contents ~drop:(function '\n' -> true | _ -> false)
```
NathanReb, line 233: Won't that leave trailing whitespaces when it used not to?
Could we do something like a `drop ~rev:true` followed by a `drop ~sat:(function '\n' -> true | _ false` to preserve that behaviour?

gpetiot: It should be fixed now, I modified the test to check this.

PR #261, opened 2020-05-18
Support OCaml 4.11

kit-ty-kate:
Currently mdx is not compatible with OCaml 4.11. This PR brings the OCaml 4.11 support.

One thing I'm not totally sure about is the `ext_uid = Uid.mk ~current_unit:"mdx"`. The name must be unique and since I'm not an mdx developper I don't know where this function might be used. If the function is not used outside of mdx, maybe it's worth using `Uid.internal_not_actually_unique`.
Here is the PR in which the change was added: https://github.com/ocaml/ocaml/pull/8934


kit-ty-kate:
Ah, apparently this requires odoc to be released first. The OCaml 4.11 support has been merged a bit ago with https://github.com/ocaml/odoc/issues/403 cc @jonludlam

jonludlam:
Thanks @kit-ty-kate - we'll figure out the best way to do this ASAP.

kit-ty-kate:
All green now, the remaining error is unrelated:
```
[ERROR] No solution for mdx: The following dependencies couldn't be met:
          - mdx -> cmdliner >= 1.0.3 -> ocaml >= 4.03.0
              base of this switch (use `--unlock-base' to force)
```

jonludlam:
It'd be handy to have this as odoc now requires mdx for tests, and this is breaking our 4.11 travis build.

Issue #264, opened 2020-06-10
Properly report errors in mdx deps


NathanReb:
`ocaml-mdx deps` seems to let some errors slip through as uncaught exceptions.

I think that for clarity it would be better if used `result` or at least our own exceptions so that we can catch mdx related errors and report them ourself rather than leak those uncaught exceptions.

As an example, when labels are invalide, one gets the following:
```
ocaml-mdx book/foreign-function-interface/README.md.mdx.deps (exit 125)
(cd _build/default/book/foreign-function-interface && ../../../install/default/bin/ocaml-mdx deps README.md) > _build/default/book/foreign-function-interface/README.md.mdx.deps
ocaml-mdx: internal error, uncaught exception:
           Failure("File \"README.md\", line 230, character 371: invalid code block: `non-deterministic` label cannot be used with a `file` label.")
```

gpetiot:
Closed with #267

PR #267, opened 2020-06-10
Properly report parsing errors

gpetiot:
First step of #264

Use a Result type instead of exception for parsing errors, so far I only did the work for mdx and mli parsing, cram and toplevel parsing should be modified as well but the change is more involved.

### DIFF
--- a/bin/deps.ml
+++ b/bin/deps.ml
@@ -22,7 +22,7 @@ let run (`Setup ()) (`Syntax syntax) (`File file) =
     | Some s, _ | None, Some s -> s
     | None, None ->
         Printf.eprintf
-          "Fatal error: could not infer syntax from filename %s, use the \
+          "[mdx] Fatal error: could not infer syntax from filename %s, use the \
            --syntax option to specify a syntax.\n"
           file;
         exit 1
--- a/bin/deps.ml
+++ b/bin/deps.ml
@@ -14,6 +14,8 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
+open Mdx.Util.Result.Infix
+
 let run (`Setup ()) (`Syntax syntax) (`File file) =
   let syntax =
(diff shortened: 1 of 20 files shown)


bin/deps.ml
```diff
@@ -25,11 +25,15 @@ let run (`Setup ()) (`Syntax syntax) (`File file) =
           file;
         exit 1
   in
-  let doc = Mdx.parse_file syntax file in
-  let deps = Mdx.Dep.of_lines doc in
-  let deps = List.map Mdx.Dep.to_sexp deps in
-  Printf.printf "%s" (Mdx.Util.Sexp.Canonical.to_string (List deps));
-  0
+  match Mdx.parse_file syntax file with
```
NathanReb, line 28: It might be nice to extract a function here that takes care of matching over the result, reporting the error and exiting with the right status so that we don't have to wrap everything in a pattern matching and think about this every time.

gpetiot: done

gpetiot:
Can we bump the requirements to ocaml > 4.02 so we can use `Format.kasprintf` ?

gpetiot:
This has been rebased and ready to be tested again to see whether it fixes #264

lib/util.ml
```diff
@@ -22,6 +22,13 @@ module Result = struct
     let ( >>= ) r f = match r with Ok x -> f x | Error _ as e -> e
 
     let ( >>| ) r f = match r with Ok x -> Ok (f x) | Error _ as e -> e
+
+    let ( >>! ) r f =
+      match r with
+      | Ok x -> f x
+      | Error (`Msg e) ->
+          Printf.eprintf "Fatal error: %s" e;
```
NathanReb, line 30: How about mentioning MDX here? Since we're moving towards a model where mdx won't be called directly, might be worth mentioning where the error message comes from so that users won't have to figure that out for themselves.

NathanReb: Also probably worth adding a `\n` at the end!

gpetiot: done

PR #268, opened 2020-06-11
Require at least cmdliner 1.0.0

emillon:
Mdx uses `Arg.conv` which has been added in 1.0.0.


emillon:
I'll hunt the history a bit to see what versions needs fixing in opam-repository.

PR #270, opened 2020-06-16
Cram test exception due to empty line: improve msg

pitag-ha:
Before, when there was an empty line between an input and its output, the error message was ``invalid cram block: [`Output "(...)"; (...)]``. Now, it is more informative.

### DIFF
--- a/lib/cram.ml
+++ b/lib/cram.ml
@@ -105,6 +105,17 @@ let of_lines t =
       (pad, aux cmd [] [] t)
   | `Command cmd :: t -> (pad, aux [ cmd ] [] [] t)
   | [] -> (0, [])
+  | `Output line :: _ ->
+      if String.length line > 0 && line.[0] = '$' then
+        failwith
+          "Blocks must start with a command or similar, not with an output \
+           line. To indicate a line as a command, start it with a dollar \
+           followed by a space."
+      else
+        failwith
+          "Blocks must start with a command or similar, not with an output \
+           line. Please, make sure that there's no spare empty line, \
+           particularly between the output and its input."
   | _ -> Fmt.failwith "invalid cram block: %a" Fmt.(Dump.list dump_line) lines
 
 let exit_code t = t.exit_code
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
 
 #### Changed
 
+- Improve error message of cram test exceptions due to empty lines in a block (#270, @pitag-ha)
+
 #### Deprecated
 
 #### Removed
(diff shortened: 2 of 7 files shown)


gpetiot:
Looks good, but could you had some tests to illustrate these cases?

lib/cram.ml
```diff
@@ -105,6 +105,17 @@ let of_lines t =
       (pad, aux cmd [] [] t)
   | `Command cmd :: t -> (pad, aux [ cmd ] [] [] t)
   | [] -> (0, [])
+  | `Output line :: _ ->
+      if String.length line > 0 && line.[0] = '$' then
+        Fmt.failwith
```
CraigFe, line 110:
```suggestion
        failwith
```

lib/cram.ml
```diff
@@ -105,6 +105,17 @@ let of_lines t =
...
+          "Blocks must start with a command or similar, not with an output \
+           line. To indicate a line as a command, start it with a dollar \
+           followed by a space."
+      else
+        Fmt.failwith
```
CraigFe, line 115:
```suggestion
        failwith
```

lib/cram.ml
pitag-ha: Good catch. Thanks

pitag-ha:
@gpetiot : You're totally right. I've just added tests.

lib/cram.ml
```diff
@@ -105,6 +105,17 @@ let of_lines t =
...
+      if String.length line > 0 && line.[0] = '$' then
+        failwith
+          "Blocks must start with a command or similar, not with an output \
+           line. To indicate a line as a command, start it with a dollar \
+           followed by a space."
+      else
+        failwith
+          "Blocks must start with a command or similar, not with an output \
+           line. Is there maybe a spare empty line somewhere, for example \
+           between the output and its input?"
```
gpetiot, line 118: I'm not sure it's common to give hints to the user phrased as a question (some might expect to have to type an answer), how about we rephrase the last sentence as something like: `Please check there is no empty line between the command and its output.` ? (or something similar)

pitag-ha: In Rust, for example, a lot of error messages include a hint of what might have gone wrong phrased as "Did you mean ...?". But it's true: now that I think about it, in OCaml it doesn't seem to be common to have questions in error messages and therefore could lead to misunderstanding. Just as a side-note: also in Ocaml I've come across a few questions in error messages, but always phrased carefully (for example explicitly as a hint), e.g. when using `ppxlib`'s `metaquot` without providing a location in the scope, it says 
"Unbound value loc 
Hint: Did you mean log or lor?"

Anyways, I agree that it's better to rephrase it, so I've done that.

Issue #265, opened 2020-06-10
Breakage in RWO since 1.6


NathanReb:
I ran some tests with our master branch against RWO and it seems that we broke some of the toplevel blocks/envs behaviour from the look of it.

The test branch can be found here: https://github.com/NathanReb/book/tree/test-new-mdx. Some of the tests in the Classes and FFI are broken with master whereas they worked fine with 1.6.

We need to look at this before being able to release a new version of MDX!

NathanReb:
`dune runtest` will work, you can also test individual chapters with `dune runtest book/<chapter>`!

gpetiot:
It seems the semantics of preludes has changed:
- then: `--prelude foo.ml` would apply the prelude for each environment
- now: `--prelude foo.ml` applies the prelude only for the default environment, we need `--prelude bar:foo.ml` to apply the prelude to environment bar

The new preludes stanza does not allow the bar:foo.ml syntax though

gpetiot:
Applying the following diff helps if we don't want to change the semantics again:
```diff
diff --git a/book/classes/dune b/book/classes/dune
index 5cf1868b..7b5520a2 100644
--- a/book/classes/dune
+++ b/book/classes/dune
@@ -4,6 +4,10 @@
   core
   mdx
   ppx_jane)
- (preludes prelude.ml))
+ (preludes
+   (env binary prelude.ml)
+   (env istack prelude.ml)
+   (env iter prelude.ml)
+   (env stack prelude.ml)))
 
 (data_only_dirs examples)
diff --git a/book/command-line-parsing/dune b/book/command-line-parsing/dune
index 5cf1868b..160b1ce6 100644
--- a/book/command-line-parsing/dune
+++ b/book/command-line-parsing/dune
@@ -4,6 +4,7 @@
   core
   mdx
   ppx_jane)
- (preludes prelude.ml))
+ (preludes
+   (env main prelude.ml)))
 
 (data_only_dirs examples)
diff --git a/book/data-serialization/dune b/book/data-serialization/dune
index 5cf1868b..3f6aeb92 100644
--- a/book/data-serialization/dune
+++ b/book/data-serialization/dune
@@ -4,6 +4,16 @@
   core
   mdx
   ppx_jane)
- (preludes prelude.ml))
+ (preludes
+   (env sexp_printer prelude.ml)
+   (env to_from_sexp prelude.ml)
+   (env example_load prelude.ml)
+   (env sexp_opaque prelude.ml)
[...]

PR #271, opened 2020-06-16
Apply unnamed preludes to all environments

gpetiot:
Fix the environment issue of #265 

New behavior:
- `env_and_file "a:f"` associates `f` to the environment named `a`
- `env_and_file " :f"` associates `f` to the default environment (legacy, don't know if this case should be preserved)
- `env_and_file "f"` associates `f` to all environments.


lib/prelude.mli
```diff
@@ -14,7 +14,9 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-val env_and_file : string -> Block.Env.t * string
+val env_and_file : string -> Block.Env.t Util.One_or_all.t * string
```
NathanReb, line 17: I understand `All | One of 'a` is unambiguous when it comes to envs but we could use an option here as well.

If we want to use this type we should probably move it to the `Env` module.

What do you think?

test/bin/mdx-test/expect/env-and-prelude/test-case.opts
```diff
@@ -1 +1 @@
---prelude default.ml --prelude a:a.ml --prelude b:b.ml
+--prelude all.ml --prelude " :default.ml" --prelude a:a.ml --prelude b:b.ml
```
NathanReb, line 1: I realize now that this tests is a bit hard to understand. How about we change it so that each prelude defines a single value that has the same name as the prelude file and we try to evaluate each variables in each blocks so that it's clear which one is available in which env.

To clarify I'm suggesting using this block:
````markdown
```ocaml
# all;;
- : string = "all"
# default;;
# a;;
# b;;
```
````
with all four envs.

I think I wrote this test as a showcase of a bug and looking at it now I'm a bit lost. I'm not sure the `x` part is relevant to what we're testing.

What do you think?

File: lib/prelude.mli (hunk 1 above)
Comment (gpetiot) [reply lib/prelude.mli:17]: I thought of using an option type but I thought it would be too confusing as we want to convey a different information

File: lib/document.ml [hunk 3]
```diff
@@ -33,3 +33,13 @@ let pp ?syntax ppf t =
...
 
 let to_string = Fmt.to_to_string pp
+
+let envs t =
+  List.fold_left
+    (fun acc line ->
+      match line with
+      | Block b ->
+          let env = Block.env b in
+          if List.mem env acc then acc else env :: acc
```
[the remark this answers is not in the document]
gpetiot: done

bin/test/main.ml
```diff
@@ -263,16 +263,15 @@ let with_non_det ~command ~output ~det non_deterministic = function
...
+    (env, to_lines file)
   in
   match (prelude, prelude_str) with
-  | [], [] -> ()
-  | [], fs -> List.iter (aux (fun x -> [ x ])) fs
-  | fs, [] -> List.iter (aux Mdx.Util.File.read_lines) fs
+  | [], [] -> []
+  | [], fs -> List.map (aux (fun x -> [ x ])) fs
+  | fs, [] -> List.map (aux Mdx.Util.File.read_lines) fs
   | _ -> Fmt.failwith "only one of --prelude or --prelude-str shoud be used"
```
gpetiot, line 275: Do you remember why this is forbidden? Couldn't we use a `[Prelude of string | Prelude_str of string] list` to take care of the evaluation order? (if this was the issue)

test/bin/mdx-test/expect/env-and-prelude/test-case.opts
gpetiot: I'm all for it, I made the change

bin/test/main.ml
NathanReb: I don't remember, I'll take a look but to be honest I think we'll just drop support for string preludes in the future.

NathanReb: From a quick look it's always been like this. I don't think it's worth changing it!

gpetiot: Alright thanks for looking at this

PR #276, opened 2020-06-19
Report #require directive errors

### LINKED ISSUE #200: Report #require directive errors
Until we have a better way to handle toplevel fragments we need to report errors from `#require` directive.

Currently what happens is that the directive can fail silently and the error mode is a quite unhelpful `Unbound module X` message.

The way `Topfind` report errors isn't great as it simply prints them to stdout but it seems that we're somehow capturing those.

Hopefully that won't be a problem in the future as we won't rely on those directives to use packages in the toplevel but until then it would prove very helpful to debug issues in RWO which relies on master!

gpetiot:
Fix #200 
I don't like the idea of testing the literal string of the output to decide whether it's an error, but there is no return code or whatever.

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -380,10 +380,12 @@ let cut_into_sentences l =
   in
   aux [] [] l
 
+let errors = ref false
+
 let eval t cmd =
   let buf = Buffer.create 1024 in
   let ppf = Format.formatter_of_buffer buf in
-  let errors = ref false in
+  errors := false;
   let exec_code ~capture phrase =
     let lines = ref [] in
     let capture () =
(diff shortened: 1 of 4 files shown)


gpetiot:
That's a good idea, I did this in the last commit. Although it's not possible to raise exceptions from this and hoping they are caught in our eval function, so I had to resort at making the errors reference global so it is toggled in case the require directive failed.

lib/top/mdx_top.ml
```diff
@@ -591,6 +616,11 @@ let init ~verbose:v ~silent:s ~verbose_findlib () =
   Topfind.don't_load_deeply
     [ "unix"; "findlib.top"; "findlib.internal"; "compiler-libs.toplevel" ];
   Topfind.add_predicates [ "byte"; "toploop" ];
+  (* [require] directive is overloaded to toggle the [errors] reference when
+     an exception is raised. *)
+  Hashtbl.add Toploop.directive_table "require"
```
Julow, line 621:
```suggestion
  Hashtbl.replace Toploop.directive_table "require"
```

This may not be useful, I don't expect a directive to be removed.

PR #277, opened 2020-06-19
Use the csexp library

### LINKED ISSUE #236: Use csexp package from opam
We should use the new https://github.com/diml/csexp instead of having our own local copy to output csexp in `mdx dep`.

gpetiot:
Fix #236


gpetiot:
Good catch, cmdliner 1.0.3 drops the result > 1.5 constraint, let's try this one.

PR #278, opened 2020-06-24
Library csexp.1.2.2 fixes the 4.02.3 compatibility

gpetiot:
That will fix our CI
see: https://github.com/ocaml/opam-repository/pull/16707

### DIFF
--- a/mdx.opam
+++ b/mdx.opam
@@ -22,7 +22,7 @@ depends: [
   "astring"
   "logs"
   "cmdliner" {>= "1.0.3"}
-  "csexp" {>= "1.2.0"}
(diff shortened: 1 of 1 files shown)


gpetiot:
It looks like it's blocked because the package is not available on opam yet (can't satisfy the deps constraints of the package)

NathanReb:
Yeah ocaml-ci probably won't pick it up before a week or so. It doesn't update its internal opam-repo very often.

That's why I'm worried about travis builds going missing.

kit-ty-kate:
> It doesn't update its internal opam-repo very often.

That changed a month ago I believe. It's now updated daily

PR #281, opened 2020-07-27
Add @emillon changelog check action

NathanReb:
This adds an action that checks if the changelog has been updated unless the `no changelog` label is attached.


NathanReb:
Hmm not sure why the action wasn't triggered :/

PR #284, opened 2020-08-25
Allow suppressing throwback of type definitions

### LINKED ISSUE #282: Feature to prevent throwback of type definitions
Right now, in Real World OCaml, every type definition is thrown back to the user, which is no what you want typically, since the type definition itself contains essentially the same information. It would be useful to have some flags to control this behavior, both on a block-by-block basis, and flags to set defaults at the tool invocation level, so we can set broad defaults.

I can think of three modes that would make sense to switch between.

- **Suppress nothing**, which is the current behavior
- **Suppress types**, suppress the restatement of top-level type-definitions only
- **Suppress all signatures**, which would include val and module declarations. This would still let printed values through.

In all of the above, type *errors* should still come through. Note that in the case of type definitions with deriving clauses, like this one:

```ocaml
type t = int * int [@@deriving sexp
```

that one should expect the following in suppress-types mode:

```ocaml
val sexp_of_t : t -> Sexp.t
val t_of_sexp : Sexp.t -> t
```

i.e., the derived value declarations should come through, but not the types.

I propose that module declarations make it through in suppress-types mode, even if those modules contain only types.

Note that something close to the "suppress all" behavior is available by dropping the # syntax. But it would be good to have this control while still keeping the basic syntax.  Indeed, in suppress types mode, it would be useful to have some lines that are type definitions and some that are value declarations in the same block, and have their behavior vary as suggested above.

gpetiot:
Fix #282


lib/top/mdx_top.ml
```diff
@@ -397,13 +397,33 @@ let eval t cmd =
...
+                | _ -> print_phrase ppf phr
+              in
+              List.iter out_sign signs
+          | `Signatures ->
+              let out_sign = function
+                | ( ( Outcometree.Osig_class_type _ | Osig_typext _
+                    | Osig_modtype _ | Osig_type _ ),
+                    _ ) ->
+                    ()
+                | _ -> print_phrase ppf phr
```
Julow, line 423: This causes the double printing. I think this should be done in two steps: filter out items from the tree then call `print_phrase` on the whole result.

Julow (commented):
I think it's worth adding a test case for `--suppress=signatures`. In this example, the module type is suppressed but not the module. I think it's expected but we should add it anyway.

```ocaml
# module type X = sig
    type t
    val x : t
  end
# module X = struct
    type t = int
    let x = 0
  end
module X : sig type t = int val x : t end
```

I have a weird result with `--suppress=signatures`, the result is printed twice and the type is not suppressed:

```ocaml
# #require "ppx_deriving.show"
# type t = int * int [@@deriving show]
type t = int * int
val pp : Format.formatter -> t -> unit = <fun>
val show : t -> string = <fun>
type t = int * int
val pp : Format.formatter -> t -> unit = <fun>
val show : t -> string = <fun>
```

Should this be an option ? I think there is a behavior that works for everyone, we should try to find it.
What about:
Suppress types/module types/class types/exceptions only if this would suppress the whole output, otherwise suppress nothing.

lib/top/mdx_top.ml
gpetiot: Good catch, fixed now

gpetiot:
> Should this be an option ? I think there is a behavior that works for everyone, we should try to find it.
> What about:
> Suppress types/module types/class types/exceptions only if this would suppress the whole output, otherwise suppress nothing.

That's a good idea, @yminsky @NathanReb would this behavior work for you?

PR #286, opened 2020-08-27
Include blocks do not require an empty block anymore

### LINKED ISSUE #285: mdx comments with no code block get ignored
Hi,

I'm using mdx to insert external files in a markdown file.

I'm expecting the following:

    # The code interface
    
    
    
    Rest of markdown

To be promoted to:

    # The code interface
    
    
    ```
    type t
    val empty : t
    ```
    
    Rest of markdown

But it does not work. Instead I need to add an empty code block:

    # The code interface
    
    
    ```
    ```
    
    Rest of markdown

I think it would a nice quality of life improvement not to require this empty block. For example, this is what `ppx_expect` (it adds a `[%expect]` block if not provided).

What do you think?

Thanks!

gpetiot:
Fix #285

### DIFF
--- a/lib/lexer_mdx.mll
+++ b/lib/lexer_mdx.mll
@@ -69,6 +69,19 @@ rule text section = parse
            List.iter (fun _ -> newline lexbuf) errors;
            newline lexbuf);
         `Block block :: text section lexbuf }
+  | "<!--" ws* "$MDX" ws* ([^' ' '\n']* as label_cmt) ws* "-->" ws* eol
+      { let labels = labels label_cmt in
+        let file = lexbuf.Lexing.lex_start_p.Lexing.pos_fname in
+        let column = lexbuf.Lexing.lex_start_p.Lexing.pos_cnum in
+        newline lexbuf;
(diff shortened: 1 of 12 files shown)


lib/lexer_mdx.mll
```diff
@@ -69,6 +69,19 @@ rule text section = parse
            List.iter (fun _ -> newline lexbuf) errors;
            newline lexbuf);
         `Block block :: text section lexbuf }
+  | "<!--" ws* "$MDX" ws* ([^' ' '\n']* as label_cmt) ws* "-->" ws* eol
```
Julow, line 72: It would be great if this regex could be shared between the two rules. I'm not sure if it's possible to do with ocamllex, we should probably split lexing and parsing.

Julow (approved):
Looks good to me.

lib/lexer_mdx.mll
gpetiot: I had a look at this but didn't find a simple way to do it, the long term goal would be to switch to `omd` to parse all that anyway

PR #292, opened 2020-10-19
Handle no such file exception

### LINKED ISSUE #288: Calling test commad on non-existing file results in uncaught internal error
Reproduction:

```
ocaml-mdx test fake-file
                Sys_error("fake-file: No such file or directory")
```

The error is easy to understand, but just for the sake of polish, it would probably be a good idea to catch this kind of common case and provider user-friendly feedback.

gpetiot:
Fix #288

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@
 #### Fixed
 
 - Report `#require` directive errors (#276, @gpetiot)
+- Handle no such file exception: the input file and the values of options `--root` and `--prelude` are checked (#292, @gpetiot)
 
 #### Security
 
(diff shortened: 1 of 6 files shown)


Julow (commented):
There is an other way to implement this, Cmdliner has a `file` conv that tests if the file exist. This way, we'll have the same error message as other tools using cmdliner:
```
  Arg.(... & pos 0 (some file) None & ...)
```
This has the added advantage that the `file` arg is shared between all commands.

Julow:
There is also `non_dir_file` that also checks that it's not a directory.

gpetiot:
> There is also `non_dir_file` that also checks that it's not a directory.

It's better indeed, thanks!

Julow (approved):
This can be done for `prelude` (`non_dir_file`) and `root` (`dir`) too, as we are at it.
Otherwise, looks good to me.

Issue #295, opened 2020-10-20
Support Windows


Julow:
Currently, Mdx on Windows doesn't work well:
- Line endings: https://github.com/realworldocaml/mdx/pull/294

- Cram blocks don't work because Mdx is calling `sh`.

- Calling ocaml-mdx-test doesn't work every time
  argv0 cannot be trusted, also Mdx doesn't control how its installed and shouldn't. PATH is the right abstraction here.
  In Mdx's testsuite, Dune gives us a path containing `/`, Windows' `execve` exits without warning and nothing happen.

- In Mdx's testsuite
  * `dune.inc` files are changed because of line encoding
  * `mli` unit tests are failing because of line encoding
  * `mdx-test/*` tests use POSIX shell features (`echo`, heredoc, `for` loops) and unix tools (`ls`, `seq`)

I'm using Mdx in Odoc's testsuite and I'd like to run tests on Windows.

emillon:
For point 3, this is not what's happening: execvp doesn't fail silently (when it fails, it raises an exception) but it reaches a state where it's a new process so dune doesn't know about it, and its outputs are closed so it cannot print anything. #299 fixes that!

emillon:
For cram-style tests, we expect something like a posix shell so it's ok to rely on the presence of sh and use things like echo, for loops, etc.

Sudha247:
Hello! Bumping this.

For context; [Eio tests run with mdx](https://github.com/ocaml-multicore/eio/tree/main/lib_eio/tests). We recently included a [Select based backend](https://github.com/ocaml-multicore/eio/pull/497) for Windows, enabling Eio to work on Windows. This issue is a blocker for [Eio 1.0](https://github.com/ocaml-multicore/eio/issues/388). It'd be nice to fix it before ideally, so we can be reasonably sure the Windows backend works (and fix issues we might run into). Happy to help with testing or other ways we can.

cc @talex5 @polytypic

emillon:
Hi.
I don't think that this issue is properly scoped at the moment (in the sense that we can't easily close it), but mdx works on windows. There used to be an issue with stray programs, fixed in #299. Cram tests are tricky to do in a portable way but that's not really a problem with mdx itself. md programs should work.

What's not working in the context of eio, precisely?

polytypic:
Let me get back to this next week (vacationing this week) as I’ll try to get the Eio mdx tests working. I suspect most of the issues can be worked around.

PR #299, opened 2020-11-13
Use create_process instead of execvp

emillon:
On Windows, execvp creates a separate process, which can confuse dune when it is waiting for the mdx process. This causes it to miss that the output file has been created. This affects mdx's test suite and possibly users of the mdx stanza. Instead we create a process and immediately wait for it in the parent.

See #295, ocaml/dune#3849.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,7 @@
 - Report `#require` directive errors (#276, @gpetiot)
 - Handle no such file exception: the input file and the values of options `--root` and `--prelude` are checked (#292, @gpetiot)
 - Keep locations from parsing instead of recomputing the lines, providing better error messages (#241, @gpetiot)
+- Use `create_process` instead of `execvp` to call `mdx-test` from `mdx`. This fixes running mdx from dune on Windows (#299, @emillon)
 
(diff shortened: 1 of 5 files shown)


gpetiot (approved):
Looks good to me, maybe add a changelog entry and that's good.
I was wondering if that was something that could be tested in a cram test with `ps` or something similar?

emillon:
It's not intended to be a visible change, so there's no value in a test for this specific detail (on the other hand that fixes a bunch of windows tests). I'll add a changelog entry!

gpetiot (approved):

emillon:
@NathanReb should this have a changelog entry?

emillon:
N.B. there's already an instance of waitpid+exit with code mapping, I'll try to factor that.

NathanReb:
I think it's worth mentioning it in the changelog indeed as I guess it could bite windows users of mdx right?

Either way it's a step toward better windows compat since it affects the main mdx binary so it's worth an entry I'd say!

emillon:
Should be all good now.

gpetiot (approved):

PR #300, opened 2020-11-13
Preview: upgrade to ocamlformat.0.17.0

gpetiot:
Rebased on top of #244


NathanReb:
If I'm not mistaken this will conflict with our omp1 dependency so let's wait for #306 before merging this!

PR #301, opened 2020-11-16
Fix unit tests on Windows

tmattio:
The failure tests in `test/bin/mdx-test` used `system "! ocaml-mdx ..."` to ensure that the exit code is not `0`. This does not work on Windows where `!` is not available. To solve this, we use the `with-accepted-exit-codes` stanza introduced in dune 2.2 and test that the exit code is `1`.

This works on macOS, but as the goal is to fix the unit tests on Windows, I'm setting up a VM to try and run the tests on it. Will update the PR if there are any other failures.

cc @emillon

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,10 +9,6 @@
 
 - Improve error message of cram test exceptions due to empty lines in a block (#270, @pitag-ha)
 
-#### Fixed
-
-- Fix the unit tests on Windows (#301 @tmattio)
-
 #### Deprecated
 
 #### Removed
@@ -22,6 +18,7 @@
 - Report `#require` directive errors (#276, @gpetiot)
 - Handle no such file exception: the input file and the values of options `--root` and `--prelude` are checked (#292, @gpetiot)
 - Keep locations from parsing instead of recomputing the lines, providing better error messages (#241, @gpetiot)
+- Fix the unit tests on Windows (#301 @tmattio)
 
 #### Security
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,7 +18,7 @@
 - Report `#require` directive errors (#276, @gpetiot)
 - Handle no such file exception: the input file and the values of options `--root` and `--prelude` are checked (#292, @gpetiot)
 - Keep locations from parsing instead of recomputing the lines, providing better error messages (#241, @gpetiot)
-- Fix the unit tests on Windows (#301 @tmattio)
+- Use with-accepted-exit-codes in tests instead of system (#301 @tmattio)
 
 #### Security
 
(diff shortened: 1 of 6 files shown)


dune-project
```diff
@@ -1,3 +1,3 @@
-(lang dune 2.0)
+(lang dune 2.2)
```
emillon, line 1: this implies bumping the dependency to dune to `>= 2.2` in `mdx.opam`. I'll defer to @NathanReb to decide if it's OK. On a related note, mdx does not use opam file generation and this would detect this inconsistency. we can set this up if you're interested (in a separate PR)

CHANGES.md
```diff
@@ -9,6 +9,10 @@
 
 - Improve error message of cram test exceptions due to empty lines in a block (#270, @pitag-ha)
 
+#### Fixed
+
+- Fix the unit tests on Windows (#301 @tmattio)
```
emillon, line 14: there's already a fixed section below. I think this should stick to just this work, ie "Use with-accepted-exit-codes in tests instead of system" - Windows is going to span more than a single PR

tmattio: Done - thanks!

dune-project
tmattio: I bumped the constraint in Opam as well, so if we think it's ok to increase the lower bound, this is solved.
I'll send a PR to generate the opam file automatically.

CHANGES.md
```diff
@@ -18,6 +18,7 @@
 - Report `#require` directive errors (#276, @gpetiot)
 - Handle no such file exception: the input file and the values of options `--root` and `--prelude` are checked (#292, @gpetiot)
 - Keep locations from parsing instead of recomputing the lines, providing better error messages (#241, @gpetiot)
+- Use with-accepted-exit-codes in tests instead of system (#301 @tmattio)
```
NathanReb, line 21: I'm sorry that @emillon had you fix this but we only add user facing changes to the changelog and I don't think changes in our test suite belong here.

dune-project
NathanReb: Yeah bumpin minor dune versions is :+1:

NathanReb (commented):
Looks good! Thanks for fixing this!

I just have a minor comment regarding the Changelog. We can merge once that's fixed!

CHANGES.md
tmattio: That makes sense! The Github build will fail without the changelog entry, but if that's ok I'll just remove it

dune-project
tmattio: PR pushed in #302

CHANGES.md
emillon: if you set the `no-changelog-needed` label (or let @NathanReb do it) it'll skip the check.

tmattio: I don't seem to be able to manage labels

NathanReb (approved):
Thanks, merging!

PR #303, opened 2020-11-18
Make dune dependency implicit

emillon:
This way it is added automatically with the right bound when generating the opam file.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -21,8 +21,6 @@
  (depends
   (ocaml
    (>= 4.02.3))
-  (dune
-   (>= 2.0))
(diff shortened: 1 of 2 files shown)


emillon:
(we missed this in #302, cc @tmattio)

emillon:
FYI, the logic is that when the dune dependency is there, it'll be used to generate the opam bound, otherwise the dune dependency is added with a bound that corresponds to dune-lang.

gpetiot (approved):

tmattio:
Good to know, thanks @emillon!

PR #304, opened 2020-11-18
Unset test: set a variable instead of using HOME

emillon:
There are two drawbacks to using `HOME`:

- it is different between systems, so this requires an ellipsis
- in Cygwin, HOME is set by the runtime system when `sh` starts, so it continues to appear even when it has been removed from the environment.

Instead, we can use an explicit variable that we control.

In addition this improves the related language to use "unset variable".


gpetiot (approved):
Looks good to me, thanks!

PR #305, opened 2020-11-18
Add a command that generates test code

voodoos:
This PR is part of the implementation of the new Dune's `mdx` stanza.
The design document can be found [here](https://github.com/ocaml/dune/issues/3955).
The corresponding Dune's PR is [here](https://github.com/ocaml/dune/pull/3956).

This is a rough first attempt to generate a test executable, but it already works for most cases.
The main missing feature is the ability to pass a list of directories containing `cm*` files to the generator. 

I made a subcommand for now, but a separate binary might be a better choice ?

### DIFF
--- a/lib/test/mdx_test.ml
+++ b/lib/test/mdx_test.ml
@@ -268,13 +268,13 @@ let preludes ~prelude ~prelude_str =
 
 let run_exn ~non_deterministic ~silent_eval ~record_backtrace ~syntax ~silent
     ~verbose_findlib ~prelude ~prelude_str ~file ~section ~root ~force_output
-    ~output ~dirs ~packages ~predicates =
+    ~output ~directives ~packages ~predicates =
   Printexc.record_backtrace record_backtrace;
   let syntax =
     match syntax with Some syntax -> Some syntax | None -> Syntax.infer ~file
   in
   let c =
-    Mdx_top.init ~verbose:(not silent_eval) ~silent ~verbose_findlib ~dirs
+    Mdx_top.init ~verbose:(not silent_eval) ~silent ~verbose_findlib ~directives
       ~packages ~predicates ()
   in
   let preludes = preludes ~prelude ~prelude_str in
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -18,6 +18,8 @@
 open Mdx.Compat
 open Compat_top
 
+type directive = Directory of string | Load of string
+
 let redirect ~f =
   let stdout_backup = Unix.dup Unix.stdout in
   let stderr_backup = Unix.dup Unix.stdout in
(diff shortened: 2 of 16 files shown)


lib/test/mdx_test.ml
```diff
@@ -385,3 +385,19 @@ module Predicate = struct
...
+      Package.unix;
+      Package.findlib_top;
+      Package.findlib_internal;
+      Package.compilerlibs_toplevel;
+    ]
+  in
+  let predicates = [ Predicate.byte; Predicate.toploop ] in
+  run_exn ~packages ~predicates ~prelude_str:[] ~non_deterministic:false
+    ~silent_eval:false ~record_backtrace:false ~syntax:None ~silent:false
+    ~verbose_findlib:false ~section:None ~root:None (*todo ?*)
```
gpetiot, line 401: Can you leave a comment more explicit than todo here?

lib/test/mdx_test.ml
```diff
@@ -385,3 +385,19 @@ module Predicate = struct
 
   let toploop = "toploop"
 end
+
+let run_exn_defaults =
```
gpetiot, line 389: I think it would make more sense to keep this function scoped in the `dune-gen` command only, to keep the mdx_test api clean (otherwise we would need to document this function signature and precise that only the `dune-gen` command should call it), what do you think?

gpetiot (commented):
It looks good to me, thanks! I only had a look at this part and not the full dune integration

lib/test/mdx_test.ml
voodoos: I am not sure that this would be easily doable: we want this function to be part of the publicly installed `ocaml-mdx` lib so that is can be used by the independent executable whose code is generated by `dune-gen`.

lib/top/mdx_top.mli
```diff
@@ -22,11 +22,17 @@ open Result
 type t
 (** The type for configuration values. *)
 
+type directive =
+  | Directory of string
+  | Load of string  (** The type for toplevel directives *)
+
+val directories : string list -> directive list
```
gpetiot, line 29: This function is not used anywhere

lib/top/mdx_top.ml
```diff
@@ -18,6 +18,10 @@
 open Mdx.Compat
 open Compat_top
 
+type directive = Directory of string | Load of string
+
+let directories = List.map (fun path -> Directory path)
```
gpetiot, line 23: This function is not used anywhere

lib/test/mdx_test.ml
gpetiot: What I meant is that we could just write this function (in the verbatim string for generated code) in `dune-gen`, since this will only be used by dune-gen, and all the parameters used (packages, predicates) are exported as well as `run_exn` (so just call run_exn instead of run_exn_defaults, and add the missing args in the verbatim string).

Or we could also make most `run_exn` optional to have the same behavior as the `run_exn_defaults` you defined. If we do this that would be nice to document the default values for these args.

voodoos: You are right, I moved it to the generated code.

lib/top/mdx_top.mli
voodoos: In fact it is: in the generated executable. I added it to the generated executable.

gpetiot (approved):
Looks good to me, thanks!

voodoos:
@gpetiot I just realised that I changed the interface of `Mdx_test` :  

https://github.com/realworldocaml/mdx/pull/305/commits/b1a229c99d2d187f4448745104759ce7a4f2be09#diff-5da8e8cd516e18aa55b7f95da3437cee754d51f6d9bdf3439fca45fa3b6b67baL33-R33

Is that ok ?

gpetiot:
It looks fine to me, not sure many people use mdx as a library anyway, I will let @NathanReb merge in case he doesn't agree

NathanReb (approved):
It all looks good to me! Have you had a chance to test it with your dune patch, I'm guessing it all works fine but I'd rather be sure before merging!

> I just realised that I changed the interface of Mdx_test

It's fine, nobody should be using it anyway. The only consumer of the library atm is RWO and I don't think it uses this part of the API at all anyway. In the future we'll probably migrate RWO's code to use OMD instead as it mostly relies on MDX for markdown parsing and to generate the right html.

voodoos:
> It all looks good to me! Have you had a chance to test it with your dune patch, I'm guessing it all works fine but I'd rather be sure before merging!

Yes I did !

NathanReb:
Awesome possum! Let's merge!

PR #312, opened 2020-12-16
Remove deprecated 'rule' command

gpetiot:
Wait for 2.0 before merging this


gpetiot:
What is the situation regarding the `#require` directive? It's weird we didn't have these regressions before rebasing this branch

NathanReb:
> What is the situation regarding the #require directive?

What do you mean by that? If you're asking whether we still want to support them then the answer is definitely yes!

Not sure why those regression showed up but I'll look into it once we have compat with 4.13 merged!

gpetiot:
On last test is KO on some versions (no idea why some are passing but that's not the purpose), `mdx-test-require-local-package`, @NathanReb do you remember the purpose of this test and if it's still makes sense after the rule command is removed?

test/bin/misc-test-cases/mdx-test-require-local-package/dune
```diff
@@ -1,9 +1,9 @@
 (rule
- (targets dune.inc.gen)
+ (alias runtest)
  (deps
-  (package mdx))
```
NathanReb, line 4: Why is this removed? This is required because of the `ocaml-mdx` and `ocaml-mdx-test` binaries trick.

NathanReb:
Have you tried without the `%{bin:...}` change. That's the only thing I can think of here.

It looks like it's looking for the other binary in the wrong place or that it's not built.

NathanReb:
Ah wait, was that the only failing test? In that case that's perfectly normal and it's because it needs the `(package mdx)` dep, as mentioned in the comment above!

gpetiot:
> Have you tried without the `%{bin:...}` change. That's the only thing I can think of here.
> 
> It looks like it's looking for the other binary in the wrong place or that it's not built.

I reverted this, it can be done in another PR anyway.

lib/block.ml
```diff
@@ -215,32 +211,6 @@ let set_variables t = t.set_variables
 
 let unset_variables t = t.unset_variables
 
-let explicit_required_packages t = t.required_packages
-
-let require_re =
-  let open Re in
-  seq [ str "#require \""; group (rep1 any); str "\"" ]
-
-let require_from_line line =
```
Leonidas-from-XIV, line 224: Since I am a bit new to this project I'm a bit confused. I don't understand what removing the rule command has to do with removing support for `require`? Or am I misunderstanding this?

Leonidas-from-XIV (commented):
The code looks fine but I am confused why this is removing so much code.

lib/block.ml
gpetiot: This code was only used to set up the `deps` field of dune rules (as part of the rule command) and it doesn't impact the interpretation of `require` directive in the blocks

Leonidas-from-XIV (approved):

PR #313, opened 2021-01-15
Require dune 2.8 for the tests

NathanReb:
This requires dune 2.8 or higher to run the tests because dune files formatting changed in 2.8 and was causing a test for `mdx rule` to fail.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -21,6 +21,7 @@
  (depends
   (ocaml
    (>= 4.02.3))
+  (dune (or (and :with-test (>= 2.8)) (>= 2.2)))
   ocamlfind
   fmt
   (cppo :build)
--- a/test/bin/mdx-rule/misc/duniverse-mode/dune.gen.expected
+++ b/test/bin/mdx-rule/misc/duniverse-mode/dune.gen.expected
@@ -9,6 +9,11 @@
   prelude.ml)
  (action
   (progn
-   (run ocaml-mdx test --prelude=prelude.ml --prelude-str
-     "#require \"prelude-str\"" %{x})
+   (run
+    ocaml-mdx
+    test
+    --prelude=prelude.ml
+    --prelude-str
+    "#require \"prelude-str\""
+    %{x})
    (diff? %{x} %{x}.corrected))))
--- a/mdx.opam
+++ b/mdx.opam
@@ -21,8 +21,8 @@ license: "ISC"
 homepage: "https://github.com/realworldocaml/mdx"
 bug-reports: "https://github.com/realworldocaml/mdx/issues"
 depends: [
-  "dune" {>= "2.2"}
   "ocaml" {>= "4.02.3"}
+  "dune" {with-test & >= "2.8" | >= "2.2"}
   "ocamlfind"
   "fmt"
   "cppo" {build}


dune-project
```diff
@@ -21,6 +21,7 @@
  (depends
   (ocaml
    (>= 4.02.3))
+  (dune (or (and :with-test (>= 2.8)) (>= 2.2)))
```
emillon, line 24: I think the more idiomatic way to express that is:

```
(dune (>= 2.2))
(dune (and (>= 2.8) :with-test))
```

test/bin/mdx-rule/misc/duniverse-mode/dune.gen.expected
```diff
@@ -9,6 +9,11 @@
   prelude.ml)
  (action
   (progn
-   (run ocaml-mdx test --prelude=prelude.ml --prelude-str
-     "#require \"prelude-str\"" %{x})
+   (run
```
emillon, line 12: I have a couple remarks about that:
- it seems wrong that formatting depends on the version of dune rather of the dune-lang. I thought it was now a pure function of dune-lang
- if there's an easy way to use (enabled_if) to gate that formatting on the dune version, that would be a bit better I think
- if this is a test for mdx rule, it'd be good to add a note somewhere in mdx rule of the things that can be undone once this feature goes away

dune-project
emillon: also, doesn't that split the universe? released version requires dune < 2.8 in CI and unreleased one requires >= 2.8

test/bin/mdx-rule/misc/duniverse-mode/dune.gen.expected
NathanReb: I agree about the formatting issues but as a MDX maintainer I have no control over this and it turns out it does depend on the dune version, not the language version.

Using enabled_if means I need to duplicate the test which frankly feels like unnecessary extra work. mdx-rule is on its way out so we're preserving the test but that's pretty much it.

One alternative is to simply require dune >= 2.8 as this formatting issue might show up for our own dune files, that way we shield ourselves from having to deal with this in the future and having to revert the changes in this PR.

dune-project
NathanReb: Duplicating atoms feels extremely odd to me and even though it will work, I feel like it relies a bit too much on the implementation details of opam for me to feel comfortable expressing the dependency this way.

NathanReb: Only the tests are impacted, released version is co-installable with all dune version >= 2.2 and so would be the next one!

test/bin/mdx-rule/misc/duniverse-mode/dune.gen.expected
emillon: I think it depending on the dune version and not dune-lang would be a dune bug. I'll ask the maintainers.

dune-project
emillon: you can find the "duplicated" pattern in several places in opam-repository: `git grep -e 'ocaml".*with-test.*>' -e 'ocaml".*>.*with-test'` (vchan, ppx_bitstring, ocamlformat, conjury)

@kit-ty-kate what is the preferred way to do that? (have different bounds for a dependency depending on `:with-test`)

test/bin/mdx-rule/misc/duniverse-mode/dune.gen.expected
nojb: It would be a bug indeed, but in a small experiment I cannot reproduce the problem locally. Rerproducing the issue in a small sample would be helpful.

PR #314, opened 2021-01-15
Format "expected" dune file

emillon:
This makes sure that we do not depend on the actual formatting of the dune file (it has changed in dune 2.8).


emillon:
As suggested by @NathanReb, this is a way to be compatible with dune 2.8. (alternative to #313)

NathanReb (approved):
Looks great, thanks for taking care of this!

PR #315, opened 2021-02-03
Use release csexp.1.3.2 instead of vendoring csexp

gpetiot:
The goal is to be sure we use the same `csexp` lib as ocamlformat-rpc (the incompatibilities due to the `result` lib should have disappeared by now?)


NathanReb (approved):
Looks good, thanks for taking care of this! I think we planned on doing it a while ago and simply forgot!

PR #316, opened 2021-02-04
Format ocaml blocks with ocamlformat-rpc

gpetiot:
Using https://github.com/ocaml-ppx/ocamlformat/pull/1692


test/bin/mdx-test/expect/errors/test-case.md
```diff
@@ -30,38 +30,34 @@ Hi!
...
-# let x =
-  1 + "42"
-Line 2, characters 7-11:
+# let x = 1 + "42";;
+Line 1, characters 13-17:
 Error: This expression has type string but an expression was expected of type
          int
 ```
 
-```ocaml non-deterministic=output
```
gpetiot, line 64: No idea what is causing this to fail, I had this issue even without integrating the formatting of blocks, and this only involves the evaluation anyway...

gpetiot:
I've added an option `--format-code` to enable this feature.

I was having a look at how we can pass an instance of a running ocamlformat-rpc server to mdx (instead of spawning the process in mdx like it's currently the case) but didn't find any solution. At first I thought the pid of the process would be enough but there is no way to get the right channels from the pid. What would be the right way to do it? I had a look at the dune rpc feature but it wasn't very helpful either

gpetiot:
Ready to be reviewed!

samoht:
> What would be the right way to do it? I had a look at the dune rpc feature but it wasn't very helpful either

Could you use unix domain socket with a know name? e.g. `/var/run/ocamlformat.sock` or something similar? Or maybe embed the version number in the filename to have one socket per ocamlformat version.

NathanReb:
> Could you use a unix domain socket with a know name? e.g. `/var/run/ocamlformat.sock` or something similar? Or maybe embed the version number in the filename to have one socket per ocamlformat version.

That does sound like a good idea!

gpetiot:
cc @emillon to get some insight on the dune side, I'm afraid ocamlformat-rpc needs another release to support sockets (or maybe just having a in_channel/out_channel is enough to support sockets)

PR #317, opened 2021-02-10
Drop OMP dependency

NathanReb:
OMP was mostly useless in MDX and its main benefit was to get pretty printers that were stable accross OCaml versions. Given the current state of OMP it will be simpler for MDX to define its own stable printers, using preprocessing for compatibility with multiple OCaml versions.


lib/top/compat_top.ml
```diff
@@ -1,4 +1,4 @@
-open Mdx.Migrate_ast
+open Mdx.Compat
```
gpetiot, line 1: To fix the CI we should add this to type_structure (from #298)
```ocaml
let type_structure e s l =
#if OCAML_VERSION >= (4, 12, 0)
    let _ = l in
    type_structure e (to_current.copy_structure s)
#else
     type_structure e (to_current.copy_structure s) l
#endif
```

NathanReb:
I added a changelog entry as I'm afraid this change might have unpredicted user facing consequences!

gpetiot (approved):
Looks good to me, thanks!

NathanReb:
@ceastlund confirmed he got it working so let's gooooo!

PR #322, opened 2021-03-30
Do not use patch_env anymore

emillon:
`Env.without_cmis` has an issue in 4.04 and 4.05: it corrupts the environment, making the toplevel unusable after (see ocaml/ocaml#1223).

The `patch_env` function in mdx makes it a noop instead. This prevents corruption, but will display different results in some cases.

This PR restricts that workaround to affected versions. This has two benefits:
    
- this gives us a way to retire this piece of code, which relies on unsafe behavior
- this aligns the mdx behavior for recent compilers (after 2017) with what the toplevel does in these corner cases

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
 
 #### Fixed
 
+- Use the same output as the normal toplevel. Mdx used to carry an unsafe patch
+  to work around a bug fixed in OCaml 4.06 and that patch would change the
+  printed types in some corner cases. (#322, @emillon)
+
 #### Removed
 
 #### Security
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -537,35 +537,6 @@ let verbose t =
 let silent t =
   add_directive ~name:"silent" ~doc:"Be silent" (`Bool (fun x -> t.silent <- x))
 
-(* BLACK MAGIC: patch field of a module at runtime *)
-let monkey_patch (type a) (m : a) (type b) (prj : unit -> b) (v : b) =
-  let m = Obj.repr m in
-  let v = Obj.repr v in
-  let v' = Obj.repr (prj ()) in
-  if v' == v then ()
-  else
-    try
-      for i = 0 to Obj.size m - 1 do
-        if Obj.field m i == v' then (
-          Obj.set_field m i v;
-          if Obj.repr (prj ()) == v then raise Exit;
-          Obj.set_field m i v')
-      done;
-      invalid_arg "monkey_patch: field not found"
-    with Exit -> ()
-
-let patch_env () =
-  let module M = struct
-    module type T = module type of Env
-
-    let field () = Env.without_cmis
-
-    let replacement f x = f x
-
-    let () = monkey_patch (module Env : T) field replacement
-  end in
-  ()
-
 let protect f arg =
   try
     let _ = f arg in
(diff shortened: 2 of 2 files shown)


test/bin/mdx-test/expect/without-cmis/test-case.md
```diff
@@ -0,0 +1,7 @@
+```ocaml
+# #require "stdio"
```
emillon, line 2: I haven't been able to write a test case that does not use `stdio` or `base`, but that would be better. I think that this is mostly caused by type aliases in external packages.

lib/patch_module.ml
```diff
@@ -0,0 +1,28 @@
...
+  let dst_obj = Obj.repr dst in
+  let src_obj = Obj.repr (src ()) in
+  if same_obj src_obj dst_obj then ()
+  else
+    match find_in_module m_obj src_obj with
+    | None -> invalid_arg "set_field: field not found"
+    | Some i ->
+        Obj.set_field m_obj i dst_obj;
+        let new_value = Obj.repr (src ()) in
+        if not (same_obj new_value dst_obj) then
```
emillon, line 27: The original code did something a bit different - it would set back the field to its original value and try the next field. I believe this was dead code, so we just error out instead

emillon: also, we can remove this assertion and do the check in `Mdx_top` instead. In that case, we can directly pass `src` as a value and the type of `set_field` becomes the simpler `set_field : _ -> src:'a -> dst:'a -> unit`.

test/bin/mdx-test/expect/without-cmis/test-case.md
emillon: this can also wait for private library support and it should be easier to replicate with a private lib

lib/patch_module.ml
gpetiot: Could we return a result type instead? `Mdx_top.patch_env` will be free to propagate or fail on error but at least `set_field` would be cleaner, after having a look at the code it looks reasonable to propagate the result to`Mdx_top.init` up to `Mdx_test.run_exn`

gpetiot:
For the record I tried to replace the module patch by using another reference for the identifier printer (`Oprint.out_ident`) -- as the `without_cmis` seems to only be useful when printing the identifier name of a value from an `outcometree` -- but without success so far.

emillon:
I investigated a bit more and found the reason of that piece of code. The PR description is updated with it. Thanks a lot @let-def for the explanation.

Regarding the feature itself: it's a breaking change for OCaml < 4.06 in some corner cases, so I think the `no-changelog` label should be removed.

When the compatibility is bumped to 4.06+, a lot of the code referenced in this PR can be removed.

emillon:
ping @NathanReb

emillon:
I updated the PR to remove this code path completely. This targets 2.0.0 so the corresponding milestone can be added.

Leonidas-from-XIV (approved):
Since we've dropped compatibility with everything <4.08 this is a good time to merge. Technically the changelog entry is unnecessary anymore since no user-visible behaviour should change anymore but I think it is nice to recognize MDX becoming better internally :+1:

emillon:
> Technically the changelog entry is unnecessary anymore since no user-visible behaviour should change anymore

Note that removing `patch_env` has visible consequences on all OCaml versions. But this workaround is not necessary anymore with the bumped ocaml dependency, so we can restore the vanilla output.

PR #323, opened 2021-03-31
Disable failing tests on Windows

emillon:
These tests do not pass today so this makes the metadata accurate. If windows runners are enabled they will succeed in building mdx, and we can fix tests on our schedule.


gpetiot (approved):
Looks good to me, thanks!

PR #324, opened 2021-04-02
Prepare the changelog for next release

gpetiot:


NathanReb (approved):
We should automate that someday!

PR #330, opened 2021-05-26
4.13 compatibility

emillon:
- Adding to `directive_table` is deprecated in 4.13. Using `add_directive` requires adding some information that will be
displayed by `#help`.
- Handle `Ctype.equal` being renamed to `Ctype.is_equal`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 #### Added
 
+- Support for OCaml 4.13 (#330, @emillon)
+
 #### Changed
 
 #### Deprecated
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,17 @@
+### unreleased
+
+#### Added
+
+#### Changed
+
+#### Deprecated
+
+#### Removed
+
+#### Fixed
+
+#### Security
+
 ### 1.10.0
 
 #### Added
--- a/lib/top/compat_top.ml
+++ b/lib/top/compat_top.ml
@@ -386,3 +386,10 @@ let top_directive_require pkg =
 #else  
   Parsetree.Ptop_dir ("require", Pdir_string pkg)
 #endif
+
+let ctype_is_equal =
+#if OCAML_VERSION >= (4, 13, 0)
+  Ctype.is_equal
+#else
+  Ctype.equal
+#endif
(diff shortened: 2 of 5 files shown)


emillon:
At the moment it's not really possible to test mdx since there's no compatible branch of odoc, but I stubbed out the odoc support and managed to get a working build with the current `trunk` (ocaml/ocaml@600aad5d29643598d4bddee346275a51930c7859). This can get merged as is or we could wait for odoc to be working (it seems better to integrate 4.13 patches in small steps rather than a large 4.13 PR at the end).

emillon:
I added some changes to make it work with the current version of odoc. This can be tested against https://github.com/Octachron/odoc/tree/weekly.

emillon:
@NathanReb I only kept the 4.13 specific bits since finding a new odoc is a separate problem being addressed in #333.

NathanReb (commented):
Awesome possum! Thanks for working on this!

I think the odoc situation has since been solved, both odoc 1.x and odoc parser have versions compatible with 4.13 now (@kit-ty-kate can you confirm ?).

Might be worth adding a changelog entry under `Fixed`? I think it's worth writing this down in the changelog.

emillon:
Done (I used `Added` for consistency with what has been done in the past but can move to `Fixed` if you prefer)

emillon:
This builds using odoc from the alpha repository, but there's an error:

```diff
diff --git a/test/bin/mdx-test/expect/warnings/test-case.md b/test/bin/mdx-test/expect/warnings.actual
index 625caaa..f7ba0ac 100644
--- a/test/bin/mdx-test/expect/warnings/test-case.md
+++ b/test/bin/mdx-test/expect/warnings.actual
@@ -6,6 +6,10 @@ let () =
   f ();;
 let x = 4
.```
+```mdx-error
+Line 3, characters 5-6:
+Warning 6 [labels-omitted]: label x was omitted in the application of this function.
+```
 
 Warning attributes must be set to print them:
 ```

(I had to add a `.` for formatting to work)

emillon:
It's because warning 6 has been turned on by default in 4.13, so that test case is not super useful anymore. I'll fix it to use another warning.

emillon:
I changed it to warning 9 and now everything's very green

gpetiot (approved):
Looks good!

NathanReb:
From reading the test I was under the impression that all warnings should be silenced by default in raw ocaml blocks and that they have to be explicitly turned on.

We need to better specify the behaviour here because if I got it right, then we need to disable it ourselves for this to work under 4.13.

emillon:
My understanding of the current behaviour is that mdx starts with the default warnings enabled (like the toplevel does), and that set varies from version to version.

NathanReb:
Ok, let's merge this and we can figure out the warnings thing later, as we collect feedback on the current behaviour!

PR #331, opened 2021-06-07
Update opam dependencies specification

NathanReb:
As shown by ocaml/opam-repository#18782, our opam file was missing some lower bounds.

This PR adds them to match our latest release's opam file.


NathanReb:
CC @kit-ty-kate it should all be there now, thanks for your help with the release!

gpetiot (approved):

PR #332, opened 2021-06-09
Show exceptions in the correct order

### LINKED ISSUE #328: Exceptions appear in wrong order
I'm trying to use MDX to document eio at https://github.com/ocaml-multicore/eio/blob/main/README.md. I'm having a problem with exceptions appearing in the wrong place. Here's a simple example:

```ocaml
# print_endline "Output"; flush stdout; failwith "Error"
Output
Exception: Failure "Error".
```

MDX wants to "correct" this by printing "Output" after the exception, even though it happened first. It doesn't seem to matter whether I output on stdout or stderr.

talex5:
Previously:

- We redirected stdout and stderr to a file.
- We copied this to an in-memory buffer from time to time.
- We wrote exceptions directly to the in-memory buffer.

This meant that exceptions raised after writing output would appear before the output in the buffer.

We now write exceptions to the file too. This means that they appear in order with the rest of the output.

Fixes #328.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,8 @@
 
 #### Fixed
 
+- Show exceptions in the correct order (#332, @talex5)
+
 #### Security
 
 ### 1.9.0
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -372,7 +372,7 @@ let errors = ref false
 
 let eval t cmd =
   let buf = Buffer.create 1024 in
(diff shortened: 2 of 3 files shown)


NathanReb (approved):
Looks great! Thanks for taking the time to fix this!

Would you mind adding a changelog entry about this? This will be good to merge then!

talex5:
Done.

gpetiot (approved):

PR #333, opened 2021-06-16
Upgrade to Odoc 2.0.0

Julow:
In Odoc 2.0.0 (which is not released yet), the parser will be packaged separately and the API will change slightly. The API might change a bit again, I'll update this PR accordingly.

An other change is the "metadata" field, allowing to attach informations on every code blocks:

```ocaml
{@ocaml labels... [ ... ]}
```

This PR updates the code to work with the new version and handle the metadata field.

An other change in this PR is that code blocks without header are no longer run. Existing code blocks that were running before need to be changed. Perhaps this change is out of scope and should be removed ?

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -35,6 +35,6 @@
   result
   (ocaml-version
    (>= 2.3.0))
-  (odoc (>= 1.4.1))
+  (odoc-parser (>= 0.9.0))
   (lwt :with-test)
   (alcotest :with-test)))
--- a/lib/mli_parser.ml
+++ b/lib/mli_parser.ml
@@ -1,7 +1,7 @@
 open! Compat
 
 module Code_block = struct
-  type t = { location : Odoc_model.Location_.span; contents : string }
+  type t = { location : Odoc_parser.Loc.span; contents : string }
 end
 
 let drop_last lst =
@@ -14,8 +14,8 @@ let drop_first_and_last = function
   | [] -> None
   | first :: tl -> Some (first, drop_last tl)
 
-let slice lines ~(start : Odoc_model.Location_.point)
-    ~(end_ : Odoc_model.Location_.point) =
+let slice lines ~(start : Odoc_parser.Loc.point) ~(end_ : Odoc_parser.Loc.point)
+    =
   let lines_to_include =
     Util.Array.slice lines ~from:(start.line - 1) ~to_:(end_.line - 1)
     |> Array.to_list
@@ -65,7 +65,7 @@ let slice lines ~(start : Odoc_model.Location_.point)
    end of "[**", but doesn't add those three characters to the within-the-file column
    number. Here, we make the adjustment.
 *)
-let account_for_docstring_open_token (location : Odoc_model.Location_.span) =
+let account_for_docstring_open_token (location : Odoc_parser.Loc.span) =
   let start_shift = 3 in
   let end_shift = if location.start.line = location.end_.line then 3 else 0 in
   {
@@ -78,8 +78,8 @@ let extract_code_blocks ~(location : Lexing.position) ~docstring =
   let rec acc blocks =
     List.map
       (fun block ->
-        match Odoc_model.Location_.value block with
-        | `Code_block contents ->
+        match Odoc_parser.Loc.value block with
+        | `Code_block (_metadata, { Odoc_parser.Loc.value = contents; _ }) ->
             let location =
               if location.pos_lnum = block.location.start.line then
                 account_for_docstring_open_token block.location
@@ -91,17 +91,17 @@ let extract_code_blocks ~(location : Lexing.position) ~docstring =
       blocks
     |> List.concat
   in
-  let parsed = Odoc_parser.parse_comment_raw ~location ~text:docstring in
+  let parsed = Odoc_parser.parse_comment ~location ~text:docstring in
   List.iter
-    (fun error -> failwith (Odoc_model.Error.to_string error))
-    parsed.warnings;
+    (fun error -> failwith (Odoc_parser.Warning.to_string error))
+    (Odoc_parser.warnings parsed);
   List.map
     (fun element ->
-      match Odoc_model.Location_.value element with
-      | #Odoc_parser.Ast.nestable_block_element as e ->
-          acc
-            [ { Odoc_model.Location_.location = element.location; value = e } ]
-      | `Tag tag -> (
+      match element with
+      | { Odoc_parser.Loc.value = #Odoc_parser.Ast.nestable_block_element; _ }
+        as e ->
+          acc [ e ]
+      | { value = `Tag tag; _ } -> (
           match tag with
           | `Deprecated blocks -> acc blocks
           | `Param (_, blocks) -> acc blocks
@@ -110,8 +110,8 @@ let extract_code_blocks ~(location : Lexing.position) ~docstring =
           | `See (_, _, blocks) -> acc blocks
           | `Before (_, blocks) -> acc blocks
           | _ -> [])
-      | `Heading _ -> [])
-    parsed.value
+      | { value = `Heading _; _ } -> [])
+    (Odoc_parser.ast parsed)
   |> List.concat
 
 let docstrings lexbuf =
@@ -128,10 +128,10 @@ let docstrings lexbuf =
   in
   loop [] |> List.rev
 
-let convert_pos (p : Lexing.position) (pt : Odoc_model.Location_.point) =
+let convert_pos (p : Lexing.position) (pt : Odoc_parser.Loc.point) =
   { p with pos_lnum = pt.line; pos_cnum = pt.column }
 
-let convert_loc (loc : Location.t) (sp : Odoc_model.Location_.span) =
+let convert_loc (loc : Location.t) (sp : Odoc_parser.Loc.span) =
   let loc_start = convert_pos loc.loc_start sp.start in
   let loc_end = convert_pos loc.loc_end sp.end_ in
   { loc with loc_start; loc_end }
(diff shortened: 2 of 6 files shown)


dune-project
```diff
@@ -35,6 +35,6 @@
   result
   (ocaml-version
    (>= 2.3.0))
-  (odoc (>= 1.4.1))
+  (odoc-parser (>= 2.0.0))
```
Julow, line 38: This makes Mdx incompatible with Odoc 1.5. Odoc 2.0.0 should be fully retrocompatible. The only reason I can think of a user keeping the current version is the change in the output.

lib/mli_parser.ml
```diff
@@ -150,12 +149,23 @@ let docstring_code_blocks str =
     (docstrings (Lexing.from_string str))
   |> List.concat
 
+let parse_metadata s =
+  match String.index_opt s ' ' with
```
Julow, line 153: It's possible that Odoc will enforce that the first word is the language tag.

Julow (commented):
The new API of the parser is still WIP: https://github.com/ocaml/odoc/pull/685

NathanReb:
Thanks for taking care of this!

I haven't looked at the code but there are a couple of high level things I'd like to discuss.

My first concern is compatibility with older code. I'd like the behaviour of MDX not to change after this patch, as much as possible. Blocks with no headers should run exactly the same as before.

My understanding was that the content between `{@` and `[` would not be interpreted by odoc at all. I'm guessing that if you're considering forcing the first component to be a language header, that means you do intend to interpret it somehow in odoc. If so, what will odoc do with this header?
If odoc does want the first bit to be a language header, then maybe it's better to make the parser aware of that so that we don't have to deal with this here and can get a pair made of the language header and the rest (which is the part we care the most about here). It's also probably a good idea to have a clear separation between language header and custom headers, so that you can have one without the other. I don't see a good reason for forcing MDX users to add a language if they want to use MDX features.

The last point is more of an open quesiton but I'm wondering if we should not use language headers at all in mli files and rely on the `kind` label instead. I guess the answer here does depend on whether odoc plans on using language headers. If not, I don't think MDX should introduce them as they don't have an actual semantic value on their own. They were used to infer block types but this has proved not to be such a good idea, leading to bad error reporting and other similar issues.
We could keep supporting toplevel blocks with no labels and enforce any other blocks to at least have the `kind` label for MDX to interpret it. I think overall it would improve MDX behaviour and make MDX document more readable for people maintaining them.

Julow:
> I'd like the behaviour of MDX not to change after this patch, as much as possible. Blocks with no headers should run exactly the same as before.

I agree that this change was a bit out of scope. I'll revert it.
My concern was that some blocks might not be intended to run or might not be OCaml code. That would make Mdx not work by default if it were run in CI for example.

> what will odoc do with this header?

The first word would be the language tag and it might be used for syntax highlighting in the html backend.

> make the parser aware of that so that we don't have to deal with this here and can get a pair made of the language header and the rest

Yes, that's what I meant.

> We could keep supporting toplevel blocks with no labels and enforce any other blocks to at least have the kind label for MDX to interpret it.

That would be great! Do you think we should change the syntax to have the language tag optional to avoid passing informations twice in this case ?

We don't have clear plans yet for the language tag, it looked interesting to add because Mdx might use it.

NathanReb:
You're right. I think supporting an optional language tag in odoc, so that it can be used to generate the right, syntax highlighted HTML is a perfectly valid reason to allow adding language tag to code blocks.

I think it should be optional indeed, both for compatibility reasons and because sometimes there might be blocks we don't know how to highlight and for which there isn't really a point in adding a language tag but where you might still want to add headers. For example, an MDX file include block that's importing the content of a text file.
To clarify I feel like users should be able to write blocks with just a language tag, just custom headers, both or none. The syntax probably need to be changed for that to work well, e.g. have a symbol marking the beginning of the header. I don't have any particular opinion on what it should look like and I trust you to make the right choice here!

I think the `kind` label doesn't necessarily dispense one from adding a language tag as they might still want to have it properly highlighted in the generated documentation.
Remains the question of how to highlight toplevel blocks, which aren't strictly speaking valid ocaml syntax but we can work this one out I guess.

Regarding simple blocks and MDX, if I remember correctly, the current behaviour is that blocks are inspected: if they follow the toplevel blocks syntax, they are evaluated by MDX, otherwise they are simply ignored. I think we should preserve this with this patch. The good news is that now, users are able to add a `skip` label to blocks they don't want MDX to temper with. I feel like it's a win-win situation!

This all looks really promising, I'm fairly excited to get that out! Let me know if there's anything I can do to help with this on either sides (MDX or odoc)!

jonludlam:
Thanks for looking this over @NathanReb! I'm curious about the problems you encountered with the language tag? I can't immediately see a useful difference between `{@ocaml foo=bar[...]}` and `{@kind=ocaml foo=bar[...]}`, but perhaps I'm misinterpreting what you're suggesting? It would be helpful to see some examples of the distinction you mean.

NathanReb:
In Markdown, the language tag is used for markdown interpeters to generate the styling for the code blocks. MDX used this information to infer how to interpret the content of the block, along with the syntax of the block content itself but it has not strong semantic meaning. There are 4 different ways to interpret a block for MDX:
1. It's an ocaml toplevel block, made of toplevel phrases interleaved with their printed evaluation, similar to what one gets running an interactive toplevel
2. A raw OCaml block, made of any valid ocaml structure. MDX will make sure this block compile by running it as a toplevel phrase but won't print any output.
3. A file include block, it's made of a bit of arbitrary text that MDX imported from a file as specified by the labels
4. A shell block, which is similar to a toplevel block except that it's made of shell commands interleaved with their standard output and error.

For example, the first 3 blocks can contain OCaml code, or something you'd want to style as OCaml code. The language tag is too ambiguous on its own and MDX can't rely on that only to find out how to interpret the block. At some point it will have to make a guess, which might differ from what the user wanted. This guess will affect how MDX report errors and it can lead to a great deal of confusion when MDX and the user had different expectations for that block.

Another issue with this is that, atm, toplevel blocks have a different syntax than the raw OCaml one, and can't be styled the same way.

In short, the block type and the language tag are two different piece of information, used for different purposes and this is why I believe they should be handled differently.
I believe the block kind is clear and explicit as to what MDX is going to do with that block both for user and for MDX itself, which in return allows MDX to guide users when they make mistake such as forgetting to specify a mandatory label or use the wrong syntax for that specific kind of block.

For compat reasons, we'll maintain that behaviour in markdown files but we'd like to encourage our users to slowly migrate to the prefered block kind approach. If I can avoid doing the same mistakes in mli files from the start, I would much rather do that!

jonludlam:
Does this mean that if a block is `included` it can't be treated as a block of ocaml? That is, I can't interpret it and add to the environment for a subsequent block to use?

jonludlam:
I think this PR should be split - one PR to switch to odoc-parser (with no support of code-block metadata) and a follow up with the rest of the changes.

NathanReb:
@jonludlam this definitely sounds sensible!

Julow:
The [parser](https://github.com/ocaml-doc/odoc-parser) is released, I updated this PR and removed the commit about the metadata field, it is ignored for now.

dune-project
Julow: This is fixed now. Odoc 1.5 is co-installable with odoc-parser.

lib/mli_parser.ml
```diff
@@ -46,62 +46,28 @@ let slice lines ~(start : Odoc_model.Location_.point)
...
-
-   According to odoc, the "b" in "bar" is at column 0 inside the docstring and at column 4
-   within the broader file. That is correct. But it says the "f" in "foo" is at column 1
-   inside the docstring and column 5 within the file. This isn't right.
-
-   The problem is that it starts counting the inside-the-docstring column number from the
-   end of "[**", but doesn't add those three characters to the within-the-file column
-   number. Here, we make the adjustment.
-*)
-let account_for_docstring_open_token (location : Odoc_model.Location_.span) =
```
Julow, line 68: The parser now keeps the location :)

Julow: I talked too fast. This is still needed, an other force-push.

Julow: I debugged this a bit more and found that there is no bug into odoc-parser but rather that it expects the starting location to account for the comment opening. That's how odoc does it: https://github.com/ocaml/odoc/blob/master/src/loader/doc_attr.ml#L56
I pushed a fix.

jonludlam: Indeed, as described here: https://ocaml-doc.github.io/odoc-parser/Odoc_parser/index.html#val-parse_comment

lib/mli_parser.ml
```diff
@@ -1,7 +1,8 @@
 open! Compat
+module Loc = Odoc_parser.Loc
```
NathanReb, line 2: I'd rather keep the explicit module name accross the code if you don't mind. I find that it helps not to use such aliases.

If the code is to bloated, I prefer the use of local opens.

NathanReb (commented):
Thanks! This looks great!

I have a minor comment but otherwise it all looks good to me.

Could you add a changelog entry? In theory it shouldn't matter as the behaviour is expected to remain the same but I think it's good to mention it, just in case we uncover bugs or changes in the future. I'll cut a release with this straight after we merge!

You mentioned a required bug fix on the odoc-parser side, has it been released? If so we can fix that module alias thing and merge!

NathanReb:
Ouh also, could you please rebase on top of master so the 4.13 build succeeds?

jonludlam:
I don't think there was a bug fix required for `odoc-parser` - it was just a difference in expectations about what the initial location meant.

Julow:
Rebased, changelog update and alias removed. Good to merge for me.

NathanReb:
This is due to the master branch updating the dune lang version!

NathanReb:
Okay, let's merge this!

Huge thanks to both of you @Julow @jonludlam for taking care of this!

PR #339, opened 2021-08-03
Recognise the metadata field on Odoc code blocks

Julow:
Odoc's parser will change again: https://github.com/ocaml-doc/odoc-parser/pull/2
It was decided to parse the "metadata" field of code blocks into two fields: the language tag and an arbitrary string. It is not possible to write the second field without the first.

If the metadata is absent, the code block is run as OCaml. This is backward compatible.

Until now, Mdx would strip it from the original file by ignoring it at parsing and not outputting it back. (<https://github.com/realworldocaml/mdx/pull/333#issuecomment-876388599>)

### DIFF
--- a/test/bin/mdx-test/expect/simple-mli/test-case.mli
+++ b/test/bin/mdx-test/expect/simple-mli/test-case.mli
@@ -19,14 +19,30 @@
       print_endline (String.concat " " [the; last; phrase])
     ]}
 
-    {[
+    With the optional header:
+
+    {@ocaml[
       # List.map (fun x -> x * x) [(1 + 9); 2; 3]
       - : int list = [100; 4; 9]
       # List.map (fun x -> x * x) [1; 2; 3]
       - : int list = [1; 4; 9]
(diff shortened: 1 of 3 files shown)


test/bin/mdx-test/expect/simple-mli/test-case.mli
```diff
@@ -19,14 +19,30 @@
...
       # List.map (fun x -> x * x) [1; 2; 3]
       - : int list = [1; 4; 9]
     ]}
+
+    A shell block:
+
+    {@sh set-FOO=Hello,set-BAR=Bash[$ echo $FOO $BAR
+Hello Bash
+
+    ]}
```
Julow, line 36: The initial newline character is moved to the end in this `sh` block. Any idea ?

Julow:
This is a draft PR because the new parser isn't released yet.

Julow:
Closed in favor of https://github.com/realworldocaml/mdx/pull/357

PR #342, opened 2021-09-02
Add warning for phrases that are missing trailing ;;

Leonidas-from-XIV:
This adds a deprecation warning about `;;` to be required in future versions of mdx along with the roadmap of how things will change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 #### Deprecated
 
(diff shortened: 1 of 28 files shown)


NathanReb:
Forgot to mention it while we were working on this but we should add a Changelog entry for this!

Leonidas-from-XIV:
There is a bug where it outputs the warning to stdout from where it gets picked up (#343), which explains the expect failure. Once this branch is rebased there it should Just Work(TM).

gpetiot (approved):
Looks good!

PR #343, opened 2021-09-03
Fix restoring stderr correctly

Leonidas-from-XIV:
When running `ocaml-mdx` it would output warnings to the "corrected" files when running with `--output -`. Since warnings are emitted on stderr this should not be happening, which means that something was messing with stderr. It turns out yes, there is code which modifies stderr but also takes care of restoring it. Unfortunately a typo caused the backup to be stderr instead so restoring would redirect stderr to stdout.

(This PR is necessary to test #342 properly without mdx writing the new warning into the corrected file)

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -22,7 +22,7 @@ type directive = Directory of string | Load of string
 
 let redirect ~f =
   let stdout_backup = Unix.dup Unix.stdout in
-  let stderr_backup = Unix.dup Unix.stdout in
+  let stderr_backup = Unix.dup Unix.stderr in
   let filename = Filename.temp_file "ocaml-mdx" "stdout" in
   let fd_out =
     Unix.openfile filename Unix.[ O_WRONLY; O_CREAT; O_TRUNC ] 0o600


Leonidas-from-XIV:
I've tested this branch with the `realworldocaml/book` repository opam-multiverse and the test is succeeding so I am concluding that this was probably indeed just a typo and not desired behaviour. Furthermore at least the book doesn't depend on the old behaviour so this should be safe to merge.

NathanReb (commented):
This looks good to me! Thanks for fixing it!

Do you think we could extract a simple regression test here?

Leonidas-from-XIV:
That's a good question, so I took a deeper look at how to deliberately cause it:

1. It needs to call the `Mdx_top.redirect` function at least once so it would misredirect `stderr`. This needs a toplevel MDX phrase be parsed once.
1. It needs to then print something to `stderr`. For this I looked at where `eprintf` is used:
   * `deprecated.ml` is warnings that we plan to remove in the future so depending a test on these seems unwise
   * `rule.ml` will only potentially print something in `duniverse_mode` which seems to happen before any phrases are evaluated
   * `deps.ml` will not call `redirect` before potentially failing
   * `main.ml` prints to `stderr` *before* potentially running `redirect`
   * `util.ml` could do the trick by triggering the error branch of `>>!` but it seems to be difficult to trigger a case where it would go into the error branch. The best I could think of is resorting the `warn` call in `rule.ml:245` under the `Mdx.run` but again, that code is just there temporarily.
 
So unless we want to introduce some special stanza that would output a warning on stderr it looks like there is no easy way unfortunately.

NathanReb (approved):

PR #345, opened 2021-09-20
Bump minimum OCaml to 4.08

Leonidas-from-XIV:
As agreed with @NathanReb this is bumping the minimum OCaml version to 4.08. Thus this PR gets to remove all the code that is necessary for support of older versions - only additional shims for 4.09+, 4.10+ 4.11+ and 4.13+ remain.


lib/top/compat_top.ml
```diff
@@ -1,52 +1,26 @@
-open Mdx.Compat
-
-#if OCAML_VERSION >= (4, 8, 0)
-let try_finally ~always f = Misc.try_finally f ~always
-#else
-let try_finally ~always f = Misc.try_finally f always
-#endif
-
 let map_error_loc ~f (error : Location.error) =
```
emillon, line 1: It looks like a bunch of functions in this module don't have `#if`s anymore. Can we remove them from this compat module? Usually it works well to inline them at the call site, or at least we can move the definition to the main library code otherwise.

lib/top/compat_top.ml
```diff
@@ -212,7 +138,7 @@ let add_directive ~name ~doc kind =
             in
             let id = Ident.create_persistent s in
             let sg = to_sig env loc id lid in
-            Printtyp.wrap_printing_env env (fun () ->
+            Printtyp.wrap_printing_env ~error:false env (fun () ->
```
emillon, line 141: what's the reason for this new parameter?

lib/top/mdx_top.ml
```diff
@@ -39,7 +38,7 @@ let redirect ~f =
     read_up_to := pos;
     Buffer.add_channel buf ic len
   in
-  try_finally
+  Misc.try_finally
```
emillon, line 41: I think we can be more future proof by porting uses of `Misc.try_finally` (in `compiler-libs` so technically unstable) to `Fun.protect` (in Stdlib, 4.08+). What do you think?

emillon (commented):
Looks good! I didn't expect so much code to go away!

lib/top/compat_top.ml
Leonidas-from-XIV: We had a `Printtyp` module which had a `wrap_printing_env` function which conditionally (dependent on OCaml version) added this parameter. Removing support for older version made this into a function that only added this parameter so I completely inlined it and removed our `Printtyp` compat shim.

lib/top/compat_top.ml
Leonidas-from-XIV: Good point, this removes even more functions (especially decreasing the size of the exposed interface). For the most part I kept the (non-exposed) helper functions since given they have reasonably useful names I found it more descriptive to keep them as such.

lib/top/mdx_top.ml
Leonidas-from-XIV: :+1: `Fun.protect` in this case is a great 1:1 replacement, only needed to rename the keyword since it provides exactly the required semantics.

emillon (approved):
Thanks!

PR #346, opened 2021-09-28
Add `;;` to toplevel blocks that are missing it

Leonidas-from-XIV:
The idea was that MDX 1.x would warn against this, MDX 2.0+ would start inserting semicolons in these fields and MDX 3.0 will drop support for this altogether.


Leonidas-from-XIV:
I removed the semisemi test and made sure the new test tests for the insertion of `;;` but without `dune promote` allowing backpromotion (thus accidentally fixing the test).

lib/deprecated.ml
```diff
@@ -27,26 +27,37 @@ let warn ?replacement s ~since =
...
+
   let report ~filename =
     if !missing_semicolon then
       Format.eprintf
         "Warning: OCaml toplevel block without trailing ;; detected in file \
          '%s'.\n\
          Non-semicolon terminated phrases are deprecated.\n\
-         MDX 2.0 will accept them as input but will output them with ;; \
-         appended.\n\
+         MDX 2.x accepts them as input but will output them with ;; appended.\n\
```
NathanReb, line 60: I wonder if it wouldn't make more sense to just say that in MDX 3.0 they won't be accepted anymore.

This line just describes the behaviour the users are just observing so it might not be necessary.

What do you think?

lib/test/mdx_test.ml
```diff
@@ -321,6 +321,7 @@ let run_exn ~non_deterministic ~silent_eval ~record_backtrace ~syntax ~silent
             Toplevel.of_lines ~syntax ~loc:t.loc t.contents
           in
           Deprecated.Missing_double_semicolon.check_block phrases;
+          let phrases = Deprecated.Missing_double_semicolon.fix phrases in
```
NathanReb, line 324: This will run through the commands twice for each of them, how about have the `fix` function also perform the check, renaming it on the way to make it clear it does both eventually?

NathanReb (commented):
Looks great! Just a couple minor comments!

lib/test/mdx_test.ml
Leonidas-from-XIV: Ah yes, I somehow didn't see the forest for the trees. This is much nicer indeed and avoids nearly identical code.

lib/deprecated.ml
Leonidas-from-XIV: Yes, this is much better, I agree.

NathanReb (approved):
Awesome, let's merge!

PR #347, opened 2021-09-28
Remove documentation of 'ocaml-mdx rule' command in the README

gpetiot:
Follow-up on #312


Leonidas-from-XIV (approved):
Ah yes, we should've remembered to remove this.

Issue #273, opened 2020-06-17
Refresh the documentation in the README


NathanReb:
We need to refresh the documentation a bit, in particular we should remove the deprecated features from it, make sure the syntax used is the latest one and that all of the main features are documented properly.

The 1.7.0 release seems like a good time to take care of that!

gpetiot:
The readme already uses the new label syntax (html comments) and presents the `mdx` dune stanza, and all reference to the deprecated binary `mdx` have already been removed.
What else needs to be updated?
The `rule` command is only deprecated since 1.7 and will be removed in 2.0.

PR #348, opened 2021-09-28
Document OCaml delimiters for code blocks

### LINKED ISSUE #338: OCaml-comment delimiters for code blocks are not documented
I was able to find the syntax by finding the PR and reading it, but I couldn't find any documentation for it in the README.

Leonidas-from-XIV:
Given the discussion in #273 I thought it would be nice to fill up some missing bits. This is one and it was even reported, so might as well document it.

Closes #338.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,9 +107,16 @@ Here is an examples of toplevel OCaml code:
 
 ### File sync
 `mdx` is also capable of including content from files in fenced code blocks
-using the label `file`. When an OCaml file is included it can be automatically
-sliced if it contains annotations such as `[@@@part "partName"]` and if the
-block has the label `part=partName`:
+using the label `file`. OCaml files can be sliced using named blocks:
+
+```ocaml
+(* $MDX part-begin=partName *)
+let meaning_of_life () =
+  print_endline "42"
+(* $MDX part-end *)
+```
+
+These can then be included in the document:
 
     <!-- $MDX file=sync_to_md.ml,part=partName -->
     ```ocaml


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 #### Added
 
+- Document OCaml comment delimiters (#338, #348, @Leonidas-from-XIV)
```
NathanReb, line 5: I think this can be considered a no changelog change!

NathanReb (commented):
Thanks for taking care of this! This looks good but I would take this one step further and entirely remove the mention of the attributes based syntax as that is also something we wish to drop support for in the future!

CHANGES.md
Leonidas-from-XIV: Good point. I was on the fence myself and saw that there was a change to the documentation with a changelog so I followed suit. Removed now.

README.md
```diff
@@ -107,9 +107,16 @@ Here is an examples of toplevel OCaml code:
 
 ### File sync
 `mdx` is also capable of including content from files in fenced code blocks
-using the label `file`. When an OCaml file is included it can be automatically
-sliced if it contains annotations such as `[@@@part "partName"]` and if the
-block has the label `part=partName`:
+using the label `file`. OCaml files can be sliced using named blocks:
+
+    ```ocaml
```
NathanReb, line 112: I think you don't want this indentend as here we're showing a fragment of ocaml code.

If we indent it like this, it will be displayed as a code block containing the full markdown block, including the delimiters.

To clarify I think what we want here is this:
```ocaml
(* $MDX part-begin=partName *)
let meaning_of_life () =
  print_endline "42"
(* $MDX part-end *)
```

but we'd currently get this:

    ```ocaml
    (* $MDX part-begin=partName *)
    let meaning_of_life () =
      print_endline "42"
    (* $MDX part-end *)
    ```
    ```

Leonidas-from-XIV: You're completely correct. I somehow didn't notice there's two layers of code blocks in the README, but it makes sense now that we want to show people how to write the code in Markdown yet show them how to mark up their OCaml code in OCaml syntax. Pushed a fix.

NathanReb (approved):
Perfect!

PR #349, opened 2021-10-01
Update failing test due to merge

Leonidas-from-XIV:
When #346 was branched this #312 wasn't merged yet so the test hasn't been run and thus not yet fixed automatically. Therefore both branches were succeeding on CI but merging them lead to a failing test.

This branch should make the `main` branch to pass CI builds again.


NathanReb (approved):
Looks good, thanks!

PR #362, opened 2021-12-13
Upgrade to `ocamlformat.0.20.1`

Leonidas-from-XIV:
Rebased version of #351, which we need for odoc-parser 1.0.0+ compatibility in #357. Now that 0.20.0 is out this should be safe to merge and the compatibility fix in 0.20.1 (that shouldn't cause reformatting) will unblock #357.

Closes #351.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,2 @@
-version=0.19.0
+version=0.20.0
 profile=conventional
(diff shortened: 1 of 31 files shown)


NathanReb (commented):
Any idea why it's removing all those empty lines? the new behaviour feels a bit inconsistent to me.

cc @gpetiot

gpetiot:
That's because `module-item-spacing = compact` is now the default value on the conventional profile.

NathanReb (approved):
Ok, all good then!

PR #363, opened 2021-12-13
Remove `require-package` label support

Leonidas-from-XIV:
This tests functionality that we don't support anymore since the way to do this is via the dune stanza.


NathanReb:
Probably worth removing the label as well as part of this PR! It should probably have a mention in the changelog, maybe simply a ref to this PR in the `remove ocaml-mdx rule` entry would suffice. What do you think?

Leonidas-from-XIV:
I've removed the label and added a changelog entry. I think it won't hurt to be specific about this in the changelog in case people wonder what happened to this label.

NathanReb:
To clarify this label was only consumed by `ocaml-mdx rule`. The only affected users would be the ones who use this label without using `ocaml-mdx rule` and they should probably be notified! I think it's alright to remove it from MDX in 2.0.0 without an explicit deprecation as it was directly tied to the rule subcommand which we did deprecate!

CHANGES.md
```diff
@@ -22,6 +22,8 @@
   is 4.08 to 4.13 now (#345, @Leonidas-from-XIV)
 - Do not install deprecated `mdx` binary anymore (#274, @gpetiot)
 - Remove deprecated `rule` command (#312, @gpetiot)
+- Remove support for `require-package` label, use the `mdx` stanza in dune
```
NathanReb, line 25: Maybe mention here that it was only used by MDX rule and that it can otherwise be safely removed!

NathanReb (approved):

PR #366, opened 2022-01-27
Allow running non-deterministic MDX stanzas using `dune-gen`

### LINKED ISSUE #365: dune-gen does not support non-deterministic mode
Hi,

It seems that dune-gen (used by stanza 0.2) does not support setting non-deterministic mode:

https://github.com/realworldocaml/mdx/blob/4942357cad00307d787c5d117ce4941678b39296/bin/dune_gen.ml#L35-L41

If we want `MDX_RUN_NON_DETERMINISTIC=1 dune runtest` to work, this needs be supported and dune to be adjusted so that it knows about the envvar dependency.

Leonidas-from-XIV:
There is a number of ways this could be implemented, but for simplicity this just uses the same environment variable as the main executable, but to avoid dependencies it doesn't use the Cmdliner flag.

Closes #365.

### DIFF
--- a/bin/dune_gen.ml
+++ b/bin/dune_gen.ml
@@ -32,8 +32,13 @@ let run (`Setup ()) (`Prelude prelude) (`Directories dirs) =
   line "    ]";
   line "  in";
   line "  let predicates = Predicate.[ byte; toploop ] in";
+  line "  let non_deterministic =";
+  line "    match Sys.getenv_opt \"MDX_RUN_NON_DETERMINISTIC\" with";
+  line "    | Some _ -> true";
+  line "    | None -> false";
+  line "  in";
   line "  run_exn ~packages ~predicates ~prelude_str:[]";
-  line "    ~non_deterministic:false";
+  line "    ~non_deterministic";
   line "    ~silent_eval:false ~record_backtrace:false";
   line "    ~syntax:None ~silent:false";
   line "    ~verbose_findlib:false ~section:None";
--- a/test/bin/mdx-dune-gen/misc/non-deterministic/dune
+++ b/test/bin/mdx-dune-gen/misc/non-deterministic/dune
@@ -30,11 +30,12 @@
 
 ;; make sure the non-deterministic is different from the deterministic
 
+(executable
+ (name different)
+ (modules different))
+
 (rule
  (alias runtest)
  (action
-  (with-accepted-exit-codes
-   1
-   (ignore-stdout
-    (run diff %{dep:dune-mdx-nondeterministic.expected}
-      %{dep:dune-mdx-nondeterministic.nondeterministic})))))
+  (run ./different.exe %{dep:dune-mdx-nondeterministic.expected}
+    %{dep:dune-mdx-nondeterministic.nondeterministic})))
--- a/test/bin/mdx-dune-gen/misc/non-deterministic/dune
+++ b/test/bin/mdx-dune-gen/misc/non-deterministic/dune
@@ -1,9 +1,7 @@
 (rule
- (target dune_gen.ml)
- (action
-  (with-stdout-to
-   %{target}
-   (run ocaml-mdx dune-gen))))
+ (with-stdout-to
+  dune_gen.ml
+  (run ocaml-mdx dune-gen)))
 
 (executable
  (name dune_gen)
@@ -12,33 +10,23 @@
  (libraries mdx.test))
 
 (rule
- (target dune-mdx-nondeterministic.deterministic)
- (deps
-  dune_gen.exe
-  (:input dune-mdx-nondeterministic))
- (action
-  (with-stdout-to
-   %{target}
-   (run ./dune_gen.exe %{input}))))
+ (with-stdout-to
+  dune-mdx-nondeterministic.deterministic
+  (run ./dune_gen.exe %{dep:dune-mdx-nondeterministic})))
 
 (rule
- (target dune-mdx-nondeterministic.nondeterministic)
- (deps
-  dune_gen.exe
-  (:input dune-mdx-nondeterministic))
- (action
-  (setenv
-   MDX_RUN_NON_DETERMINISTIC
-   1
-   (with-stdout-to
-    %{target}
-    (run ./dune_gen.exe %{input})))))
+ (setenv
+  MDX_RUN_NON_DETERMINISTIC
+  1
+  (with-stdout-to
+   dune-mdx-nondeterministic.nondeterministic
+   (run ./dune_gen.exe %{dep:dune-mdx-nondeterministic}))))
 
 (rule
  (alias runtest)
  (action
-  (diff %{dep:dune-mdx-nondeterministic.expected}
-    %{dep:dune-mdx-nondeterministic.deterministic})))
+  (diff dune-mdx-nondeterministic.expected
+    dune-mdx-nondeterministic.deterministic)))
 
 ;; make sure the non-deterministic is different from the deterministic
 
--- a/test/bin/mdx-dune-gen/misc/non-deterministic/dune
+++ b/test/bin/mdx-dune-gen/misc/non-deterministic/dune
@@ -1,0 +1,52 @@
+(rule
+ (target dune_gen.ml)
+ (action
+  (with-stdout-to
+   %{target}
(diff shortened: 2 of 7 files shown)


bin/dune_gen.ml
```diff
@@ -32,8 +32,13 @@ let run (`Setup ()) (`Prelude prelude) (`Directories dirs) =
   line "    ]";
   line "  in";
   line "  let predicates = Predicate.[ byte; toploop ] in";
+  line "  let non_deterministic =";
+  line "    match Sys.getenv_opt \"MDX_RUN_NON_DETERMINISTIC\" with";
+  line "    | Some _ -> true";
```
emillon, line 37: I think you want to allow a value that disables, e.g. `MDX_RUN_NON_DETERMINISTIC=0`. You can check how cmdliner uses that.

emillon: and `getenv_opt` is recent-ish so you might want to check that it's always available when mdx is used, or use `getenv`.

test/bin/mdx-dune-gen/misc/non-deterministic/dune
```diff
@@ -0,0 +1,52 @@
+(rule
+ (target dune_gen.ml)
+ (action
+  (with-stdout-to
+   %{target}
```
emillon, line 5: you can just write `dune_gen.ml` here and remove `(target)`. then you can inline the `action`: `(rule (with-stdout-to dune-gen.ml (run ...)))`

test/bin/mdx-dune-gen/misc/non-deterministic/dune
```diff
@@ -0,0 +1,52 @@
...
+
+(rule
+ (target dune-mdx-nondeterministic.deterministic)
+ (deps
+  dune_gen.exe
+  (:input dune-mdx-nondeterministic))
+ (action
+  (with-stdout-to
+   %{target}
+   (run ./dune_gen.exe %{input}))))
```
emillon, line 22: same here, you can inline input and target and use the short form with just the action.

test/bin/mdx-dune-gen/misc/non-deterministic/dune
```diff
@@ -0,0 +1,52 @@
...
+ (deps
+  dune_gen.exe
+  (:input dune-mdx-nondeterministic))
+ (action
+  (setenv
+   MDX_RUN_NON_DETERMINISTIC
+   1
+   (with-stdout-to
+    %{target}
+    (run ./dune_gen.exe %{input})))))
```
emillon, line 35: again here

test/bin/mdx-dune-gen/misc/non-deterministic/dune
```diff
@@ -0,0 +1,52 @@
...
+
+(rule
+ (alias runtest)
+ (action
+  (diff %{dep:dune-mdx-nondeterministic.expected}
+    %{dep:dune-mdx-nondeterministic.deterministic})))
```
emillon, line 41: `diff` knows that its arguments are dependencies so you can drop the `%{dep:}` forms

test/bin/mdx-dune-gen/misc/non-deterministic/dune
```diff
@@ -0,0 +1,52 @@
...
+
+;; make sure the non-deterministic is different from the deterministic
+
+(rule
+ (alias runtest)
+ (action
+  (with-accepted-exit-codes
+   1
+   (ignore-stdout
+    (run diff %{dep:dune-mdx-nondeterministic.expected}
```
emillon, line 51: I'm not completely sure that we can expect the `diff` binary to be available in PATH but can't think of a way to do that with just dune.

bin/dune_gen.ml
Leonidas-from-XIV: According to the [Cmdliner docs](https://erratique.ch/software/cmdliner/doc/Cmdliner/Arg/index.html#val-flag) it doesn't allow it:

> `flag i` is a bool argument defined by an optional flag that may appear at most once on the command line under one of the names specified by `i`. The argument holds true if the flag is present on the command line and false otherwise.

I was surprised myself, but thought it would be better to be consistent between the MDX cli and dune-gen. I can of course also implement disabling the option by un-setting it.

`Sys.getenv_opt` is 4.05+ and if I understand the various lower-bound flags in dune then dune requires 4.08+ so I think we should be fine. But it is a good point that the minium supported OCaml version of MDX needs to match dune's. In our case, it matches it exactly so that's good!

(Also the test compiles the `dune_gen.ml` so from this PR on we are also checking whether the dune-gen output is 4.08+ compatible)

emillon: all good then!

emillon: (I don't think we have to follow what dune supports: you can install and run dune on a 4.02 project, but if you use the mdx stanza it's 4.08+)

test/bin/mdx-dune-gen/misc/non-deterministic/dune
Leonidas-from-XIV: Ah, that's convenient. I've applied all these improvements and the file shrunk a lot. Thanks!

emillon: np. I'm like clippy but for dune files.

test/bin/mdx-dune-gen/misc/non-deterministic/dune
Leonidas-from-XIV: I was scouting what I can use because I didn't find a way to basically negate `(diff)` so `diff` (and `cmp`) was the best I could find. At least it seems to consistently use 0, 1 and >1 for exit codes between GNU and BSD.

The alternative was shelling out and dealing with `[` or `[[` but that seemed even worse and possibly even less portable.

Fortunately these are just required to run the tests so unless someone wants to run tests on Windows without WSL or MinGW I think it is acceptable for now.

Leonidas-from-XIV: An alternative I just thought of is writing a short OCaml helper binary that checks whether two files are different, and given the it can terminate on the first byte that doesn't match without any kind of complex reporting should be quite a simple program.

NathanReb: In the future we can always disable that specific test on windows and eventually write a windows version of it.

Leonidas-from-XIV: I decided to make an OCaml version of it since the functionality is simple and it doesn't have to be particularly efficient to compare two text files. Thus I could write it in a way where I don't need to redirect stdout or deal with exit codes without having to write anything too complex. In a way I find the build now more understandable since I'm not using `diff` to do the exact opposite of what people use `diff` for.

NathanReb (approved):
Looks good to me!

This still needs a change on the dune side to declare the env var dependency right? At some point it would be nice to include dune stanza tests here as well so we're confident we don't break the compat with dune on those features!

Leonidas-from-XIV:
That's a good point for the future.

Unfortunately that would require us to use a dune version that has either 0.1 support or 0.2 support (the latter is dune 3.0 which is not released yet) unless we want to do a fun thing and declare:

```
dune | dune {>= "3.0" & with-test}
```

emillon:
> Looks good to me!
> 
> This still needs a change on the dune side to declare the env var dependency right? At some point it would be nice to include dune stanza tests here as well so we're confident we don't break the compat with dune on those features!

That's done in https://github.com/ocaml/dune/pull/5391. This only affects stanza 0.2 which is unreleased at the moment.

NathanReb:
That's a good point indeed.

On the other hand I think that it wouldn't be too bad for MDX to require the version of dune that integrates well with it. We don't really offer good support for running MDX on your own anyway.

We'd need to look into the implications more carefully but on principle, I think it could work.

emillon:
We can also test this by doing what the stanza does by hand, ie something like that in a cram test:

```
$ mdx dune-gen > t.ml
$ cat > dune <<EOF
> (executable ...)
$ dune exec ...
```

It's an e2e test of dune-gen, not of the dune stanza, but this ensures that this contract works.

PR #371, opened 2022-02-28
Fix compatibility with Cmdliner 1.1.0

Leonidas-from-XIV:
Currently the build fails because of the deprecations. I guess this is a good point to upgrade then.

Since we already depend on OCaml 4.08 depending on Cmdliner 1.1.0 should not be not much of a problem.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 
 #### Fixed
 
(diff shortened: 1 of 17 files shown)


NathanReb (approved):
Looks good to me! Thanks!

PR #376, opened 2022-06-07
mdx_test: Lazily initialize OCaml toplevel

bcc32:
This changes the OCaml toplevel to be a lazy value, which is only
forced when needed to run tests in an OCaml block.

The motivation behind this change is to avoid stderr output from the
init process (e.g., errors while registering pretty-printers) spilling
into the test executable's output, which can trip up some build tools.

### DIFF
--- a/lib/test/mdx_test.ml
+++ b/lib/test/mdx_test.ml
@@ -332,8 +332,9 @@ let run_exn ~non_deterministic ~silent_eval ~record_backtrace ~syntax ~silent
     match syntax with Some syntax -> Some syntax | None -> Syntax.infer ~file
   in
   let c =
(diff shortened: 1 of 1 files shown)


samoht:
Do you mind rebasing this?

bcc32:
Done, thanks for the ping.

samoht:
Thanks! Do you think it would be possible to add a test case for this?

samoht:
Also: could you add a change entry? Many thanks!

bcc32:
In trying to come up with a good test, I realized that the original bug we observed was caused by a separate internal Jane Street patch to mdx (for our custom toplevel backend) and could not be reproduced in this repo.  I have gone through and refactored the fix to be more specifically tailored for our internal toplevel backend and I think this can be dropped.

Thank you for taking a look at this PR!

PR #378, opened 2022-07-06
Remove mi_parser unit test

Julow:
This test used to compare formatted versions of ocaml values large
enough to make understanding and updating the output impossible.

The expected values can't easily be converted to non-string values
because they do not correspond to the actual type (the 'pp' function
loses information), which is too complicated and unstable to be written
manually.

I think this test is not necessary, mli support is covered by the
black-box tests.


Leonidas-from-XIV (approved):
Seems ok to me, there is not all that much use in verifying the exact output of the parser in `Text` and `Block` bits and `test/bin/mdx-test/expect/simple-mli/test-case.mli` contains essentially the same test cases.

PR #379, opened 2022-07-06
Use a simpler location type

Julow:
On top of https://github.com/realworldocaml/mdx/pull/377 and https://github.com/realworldocaml/mdx/pull/378 as this PR might not be merged quickly.

Mdx internally uses the compiler's Location.t type but it has to handle
other locations types: Lexing.position, Odoc_parser.Loc.span.

The conversion from Odoc's locations is complicated and is sometimes
wrong, for example the 'pos_cnum' is computed and then used incorrectly.

The new location type is easier to computed and to use. It is also
simpler, for example it doesn't support ranges or absent filenames.

Also, contain some fixes to locations.


Julow:
This is obsolete since https://github.com/realworldocaml/mdx/pull/423

Leonidas-from-XIV:
It could still use a simpler location type, not much has changed? I mean, it's also not a high priority.

PR #380, opened 2022-07-07
Document the possible labels

talex5:
There didn't seem to be a list of the possible labels anywhere, so I added one, along with my guesses about what they do.

I also modified a `sudo rm -rf /` in a test case, because I don't want to wipe my computer if a test fails. Possibly the test case doesn't need to damage the computer at all, but at least the new version is easier to recover from.

### DIFF
--- a/test/bin/mdx-test/expect/shell-file-inc/script.sh
+++ b/test/bin/mdx-test/expect/shell-file-inc/script.sh
@@ -1,1 +1,1 @@
-$ sudo rm -rf /
+$ sudo chmod a-x /usr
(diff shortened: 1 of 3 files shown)


test/bin/mdx-test/expect/shell-file-inc/script.sh
```diff
@@ -1 +1 @@
-$ sudo rm -rf /
\ No newline at end of file
+$ sudo chmod a-x /usr
```
NathanReb, line 1: This is never run but simply included as text. I agree this is a poor choice of command to write though and I'm not sure why we wrote this.

PR #381, opened 2022-07-07
Allow writing out a block to a file

talex5:
This allows input data to be shown with syntax highlighing and then used in future steps. e.g.

    
    ```json
    { "value": 42 }
    ```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 #### Added
 
+- Support for writing files from blocks. (#381, @talex5)
+
 #### Changed
 
 #### Deprecated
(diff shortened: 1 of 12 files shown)


talex5:
I tried using this for the capnp-rpc tutorial. But since discovering the `dir` label in #380 I found I could keep different versions of the file in different directories and achieve the same effect, so maybe this isn't needed.

NathanReb:
This feature use to exist but was causing troubles and we removed it. Now that we have the dune stanza we might be able to eventually re-introduce it but I'd like to first have a good understanding of what you need it for!

If you found an alternative for now, I indeed think it's preferable! This otherwise can lead to quite complex promotion rules. Nothing dune can't handle but it can easily confuse users.

PR #382, opened 2022-07-08
Report exits from the toplevel correctly

talex5:
Previously, you just got:

                    Compenv.Exit_with_status(125)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,8 @@
 #### Fixed
 
 - Fixed compatibility with Cmdliner 1.1.0 (#371, @Leonidas-from-XIV)
+- Report errors and exit codes of toplevel directives (#382, @talex5,
+  @Leonidas-from-XIV)
 
 #### Removed
 
--- a/lib/top/compat_top.ml
+++ b/lib/top/compat_top.ml
@@ -195,23 +195,19 @@ let action_on_suberror b =
   if not b && not !Sys.interactive then
     raise (Exit_with_status 125)
 
-(* [load] cannot be patched to return errors because the underlying code is not exposed:
-   It would require [Topdirs.load_file] with the first argument to be [false] but the exposed
-   version hardcodes it to [true].
-  *)
-let mdx_load = patch_directive "load" (Directive_string (Topdirs.dir_load std_err))
-
 let dir_use ppf name =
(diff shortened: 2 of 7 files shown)


Leonidas-from-XIV:
Looks like there is a multi-version incompatibility:

 * 4.08 to 4.11 don't have `Compenv.Exit_with_status`
 * 4.12 to 4.13 don't throw errors on invalid `#use`
 * 4.14 works

It's quite hard to build a compatibility wrapper in `Compat_top` to work around this if there isn't a consistent way to get the error between different versions, especially since before < 4.14 these don't seem to trigger errors.

I looked a bit around in the source code of 4.12 vs 4.14 and it seems like `execute_phrase` has seen some considerable rework between the releases which has probably lead to better error reporting.

Leonidas-from-XIV:
Okay, I looked into this and I think I resolved all the issues here:

1. Redefined a lot of directives (those exposed in `Topdirs`) to write to `stderr` instead of `stdout`, so they get captured. 4.14 does that conditionally, but here we expect it to be always to be on `stderr`. There's some subtlety to this because depending on the supported version this looks different. In particular, `#load` can only be overridden in 4.13+ because older versions don't expose `load_file` which does not recurse. We could try to copy-paste the definition but I opted to avoid copying too much code and keep the complexity (somewhat) low at the cost of less than 100% identical behavior.
2. The redefined directives no longer use `ignore` to suppress the error but throw `Exit_with_status`, similar to how it is done on 4.14, but with our own exception so it works on older OCaml versions as well.

The directives are all implemented as custom `mdx_<original-name>` directives, because when I tried to override the existing ones there is a race condition between the original and the patched ones. So instead the code rewrites e.g. `#use` into `#mdx_use` on OCaml <4.14 which avoids the headache altogether.

This was a fun challenge, teaching me a thing or two how the directives in the toplevel work.

talex5:
Thanks - I hadn't realised how much work it would be to support older versions!

Leonidas-from-XIV:
I could move stuff out of `Compat_top` but besides being "somewhere else" I feel there is not particularly much that can be done, given this is all icky reimplementing of internal libs, trying to duplicate the least amount of code.

Thanks for finding the issues with the constraints, looks like I copypasted exactly the one line that had a problem.

Issue #383, opened 2022-07-13
Isolating Unix Dependency in mdx


SaySayo:
Mdx has only one library at the moment which uses unix for some parts of its processes. We'd like to isolate the part of mdx that uses unix from the parts that don't. Essentially, what I'm suggesting is to have 2 libraries for mdx, one that contains unix and one that doesn't. I'm currently working on integrating ocaml notebooks into vscode and need to use mdx in executing the cell output but do not to use the unix dependency as I won't be having access to the file system because I am compiling to javascript. 

What would be helpful is to find a way of isolating the unix dependent part from the independent part and refactoring the code to suit our purposes. We could probably try figuring it out but it would be super helpful if you have any guidelines or suggestions for us to achieve this. Of course, that's if you have the time. 

cc @tmattio

PR #384, opened 2022-07-13
Portable toplevel exit 

Leonidas-from-XIV:
Experimentation with the #383 PR opened to get CI to work.


Leonidas-from-XIV:
Pushed the changes to #382 so this can be safely closed.

PR #385, opened 2022-07-21
Introduce new syntax for explicit block type

NathanReb:
This introduce a new, more explicit labels `block-type=...` which is meant to replace the four separate labels `cram`, `ocaml`, `toplevel` and `include` in the future.

It is added for a transition period at the end of which the four labels of the apocalypse should be removed in favor of this new one.

I'm opening this as a draft as one of the test relies on MDX reporting multiple errors at once which I thought we had but turns out we still haven't so I'll work on this next to unblock this PR.

Other than that it is ready for review!


Leonidas-from-XIV (commented):
Looks good so far. Question is how we want to handle the migration:

* Should `ocaml` blocks be converted to `block-type=ocaml` blocks
* Should we do this now or should we support both in parallel?

test/bin/mdx-test/expect/block-type/test-case.md.expected
```diff
@@ -0,0 +1,23 @@
...
+
+The following blocks use a volontarily misleading language header that would
+normally lead to errors if we let MDX infer the type of block based on them.
+
+```ocaml block-type=toplevel
+# 1 + 1;;
+- : int = 2
+```
+
+```ocaml block-type=ocaml
```
samoht, line 13: This seems a bit verbose? Is `ocaml` the shortcut for `ocaml block-type=ocaml` ?

test/bin/mdx-test/failure/block-type-value/test-case.md.expected
```diff
@@ -0,0 +1 @@
+[mdx] Fatal error: File "test-case.md", line 6: invalid code block: Label `block-type` requires a value.
```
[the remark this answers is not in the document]
NathanReb: Yes hence me working on multiple error reporting before undrafting this!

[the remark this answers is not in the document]
NathanReb: Also I should rewrite the tests to use the new labels syntax!

NathanReb:
The label will never be mandatory and I think we should probably keep the inference. It just provides a way to be explicit about things, providing better error messages. The label must not be confused with the language header. We can switch the label to `type` if the length really is a problem.

NathanReb:
> for example a single character: $ocaml, $toplevel. $ is used in the extended label syntax.

The idea for this new "syntax" here is to make the label more easily readable and understandable to anyone. I agree `block-` was redundant and removed it but I believe having a regular label syntax that state what this value is about helps understanding what it does, looking up the label in documentation (which I'm currently writing and is part of why I worked on this feature), etc...

I think the 4 extra characters don't do a lot of harm but they do simplify things for the newcomers to projects using MDX quite a lot!

NathanReb:
I'll undraft once #389 is merged!

NathanReb:
Just rebased!

CHANGES.md
```diff
@@ -3,6 +3,9 @@
 #### Added
 
 - Report all parsing errors in Markdown files (#389, @NathanReb)
+- Add alternative syntax for explicitly setting the block-type.
+  The new label `block-type=...` can be set to `ocaml`, `toplevel`, `cram` or
```
Leonidas-from-XIV, line 7: This needs to be updated to be `type=`.

test/bin/mdx-test/expect/block-type/test-case.md
```diff
@@ -0,0 +1,24 @@
+It is possible to explicitly state the type of a block using the
+`block-type` label, working around language header and content based
```
Leonidas-from-XIV, line 2:
```suggestion
`type` label, working around language header and content based
```

test/bin/mdx-test/expect/block-type/test-case.md
```diff
@@ -0,0 +1,24 @@
...
+inference which can sometime lead to troublesome error messages.
```
Leonidas-from-XIV, line 3: I'd say "it can sometimes fail and lead to confusing error messages" to state why these error messages happen (because I've been curious myself).

test/bin/mdx-test/expect/block-type/test-case.md.expected
```diff
@@ -0,0 +1,26 @@
+It is possible to explicitly state the type of a block using the
+`block-type` label, working around language header and content based
```
Leonidas-from-XIV, line 2:
```suggestion
`type` label, working around language header and content based
```

But I assume this will just get `dune promote`'d :-)

test/bin/mdx-test/failure/block-type-value/test-case.md
```diff
@@ -0,0 +1,14 @@
+This tests that erros are properly reported when the `block-type` label
```
Leonidas-from-XIV, line 1:
```suggestion
This tests that erros are properly reported when the `type` label
```

test/bin/mdx-test/failure/block-type-value/test-case.md
```diff
@@ -0,0 +1,14 @@
...
+is misused.
+
+It requires a value
+
+<!-- $MDX type -->
+```ocaml
+```
+
+It only accept a fixed set of values
```
Leonidas-from-XIV, line 10:
```suggestion
It only accepts a fixed set of values
```

test/bin/mdx-test/failure/include-without-file-label/test-case.md
```diff
@@ -0,0 +1,6 @@
+Explicitly setting the block-type to include has little benefits except
```
Leonidas-from-XIV, line 1:
```suggestion
Explicitly setting the `type` to include has little benefits except
```

Leonidas-from-XIV (commented):
The code is good, my suggestions are mostly just from the renaming of `block-type` to `type`. Just overlooked details.

What I would however like to see would be some free-text documentation (e.g. in the README) about this, maybe with some advice whether and when to use `ocaml`, `type=ocaml` or inference, to make this a bit more discoverable for users as well as clearer which of the various choices make the most sense in their particular situation.

NathanReb:
Indeed, it seems to be good to clarify why we added explicit block types in the first place:

The main motivation was that people were sometime confused at error messages or behaviour. For instance they want to write a toplevel block but forget the leading `# ` and can't figure out why it doesn't behave the way they want it to because MDX just treated it as an OCaml block. By adding the explicit `type=toplevel`, they would get an error about the misisng the `#`.
There are a few other such examples but I can list them from the top of my head.

The other reason is clarity of the semantic. Using explicit labels, which are easy to look up (especially once we finish the manual), makes it clear what the intent of the block is and easier for newcomers to understand how MDX will behave.

The straight `ocaml`, `toplevel`, etc. form should be gradually removed as it doesn't help much with clarity. The explicit label name can be looked up in the documentation and will inevitably lead to describe what the four types of block are.

My personal recommendation would be to always use the explicit type label as when writing an MDX block, one usually knows what they want to do with it. I understand some people will prefer conciseness to clarity hence why we'll keep block inference.

NathanReb:
I just rebased and hopefully fixed the issues you pointed out. The CI failure on alpine seems unrelated!

Issue #368, opened 2022-02-10
spurious `;;` are added


c-cube:
I have lines like:

```
# v;; (* now v is a mutable vector *)
- : (int, CCVector.rw) CCVector.t = <abstr>
```

and mdx now gives me a diff like:

```

-# v;; (* now v is a mutable vector *)
+# v;; (* now v is a mutable vector *);;
 - : (int, CCVector.rw) CCVector.t = <abstr>
```

(to reproduce, `make test` in https://github.com/c-cube/ocaml-containers )

Leonidas-from-XIV:
With #398 this issue could be solved.

PR #386, opened 2022-07-26
Deduplicate implementation of `;;` detection

Leonidas-from-XIV:
While pondering a solution for #368 I realized that the code is full of duplicated definitions of `ends_by_semi_semi`, so I just decided to make it consistent. `Mdx.Block` might not be the best place to put it, but I gather it is still better than having it copied in 3 places. Interestingly, it was all the same implementation.

4.13 can use `String.ends_with` but that's a pretty high limit so I decided to use the implementation from `Astring` to make it a bit more readable, which we depend on already.

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -219,7 +219,7 @@ let ends_by_semi_semi c =
   match List.rev c with
   | h :: _ ->
       let len = String.length h in
-      len > 2 && h.[len - 1] = ';' && h.[len - 2] = ';'
+      len > 2 && Astring.String.is_suffix ~affix:";;" h
   | _ -> false
 
 let pp_line_directive ppf (file, line) = Fmt.pf ppf "#%d %S" line file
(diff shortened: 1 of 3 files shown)


lib/block.ml
```diff
@@ -219,7 +219,7 @@ let ends_by_semi_semi c =
   match List.rev c with
   | h :: _ ->
       let len = String.length h in
-      len > 2 && h.[len - 1] = ';' && h.[len - 2] = ';'
+      len > 2 && Astring.String.is_suffix ~affix:";;" h
```
Julow, line 222: Why `len > 2` rather than `len >= 2` ? (the latter is implicitly checked by `is_suffix`)

Leonidas-from-XIV: I assumed that it also checks that there is *something* in there besides just the `;;` but this might have been a typo in the original code. Maybe it really was just a guard for the string subindexing. Removing the check altogether makes the tests all pass so I've amended the code to remove the check.

Julow (approved):
I think it's fine to remove the `len > 2` check.

PR #389, opened 2022-08-01
Report all syntax errors in Markdown files

NathanReb:
This PR improves the MDX workflow by making sure that all parsing errors are reported at once rather than exiting on the first errors.

Obviously that only impacts errors that we can recover from e.g. mis-used or missing labels, syntax error inside a block but not for errors such as invalid markdown.

We used to do all the work in markdown/cram parser. I changed that to extract raw bits of data from the parser and leave their interpretation to a later pass.

While working on this I discovered that the locations we attach to blocks are incorrect due to inconsistencies in `eol` handling in the lexer code which I took the liberty to fix as part of this PR. I can eventually extract it as a separate PR depending on this one if you'd like but basing it on this work made it significantly simpler.

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -27,7 +27,7 @@ let locate_error_msg ~loc s =
 
 let locate_errors ~loc r =
   Result.map_error
-    (fun l -> List.map (function `Msg m -> `Msg (locate_error_msg ~loc m)) l)
+    (fun l -> List.map (fun (`Msg m) -> `Msg (locate_error_msg ~loc m)) l)
     r
 
 module Header = struct
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -20,11 +20,10 @@ let loc_error ~loc fmt =
   Format.kasprintf
     (fun s -> Error (`Msg s))
     ("%a: invalid code block: " ^^ fmt)
-    Stable_printer.Location.print_loc loc
+    Stable_printer.Location.pp loc
 
 let locate_error_msg ~loc s =
-  Format.asprintf "%a: invalid code block: %s" Stable_printer.Location.print_loc
-    loc s
+  Format.asprintf "%a: invalid code block: %s" Stable_printer.Location.pp loc s
 
 let locate_errors ~loc r =
   Result.map_error
@@ -139,7 +138,7 @@ let dump ppf ({ loc; section; labels; contents; value; _ } as b) =
   Fmt.pf ppf
     "{@[loc: %a;@ section: %a;@ labels: %a;@ header: %a;@ contents: %a;@ \
      value: %a@]}"
-    Stable_printer.Location.print_loc loc
+    Stable_printer.Location.pp loc
     Fmt.(Dump.option dump_section)
     section
     Fmt.Dump.(list Label.pp)
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -16,6 +16,21 @@
 
 open Util.Result.Infix
 
+let loc_error ~loc fmt =
+  Format.kasprintf
+    (fun s -> Error (`Msg s))
+    ("%a: invalid code block: " ^^ fmt)
+    Stable_printer.Location.print_loc loc
+
+let locate_error_msg ~loc s =
+  Format.asprintf "%a: invalid code block: %s" Stable_printer.Location.print_loc
+    loc s
+
+let locate_errors ~loc r =
+  Result.map_error
+    (fun l -> List.map (function `Msg m -> `Msg (locate_error_msg ~loc m)) l)
+    r
+
 module Header = struct
   type t = Shell of [ `Sh | `Bash ] | OCaml | Other of string
 
@@ -41,6 +56,26 @@ module Header = struct
 end
 
 type section = int * string
+
+module Raw = struct
+  type t =
+    | Include of { loc : Location.t; section : section option; labels : string }
+    | Any of {
+        loc : Location.t;
+        section : section option;
+        header : string;
+        contents : string list;
+        label_cmt : string option;
+        legacy_labels : string;
+        errors : Output.t list;
+      }
+
+  let make ~loc ~section ~header ~contents ~label_cmt ~legacy_labels ~errors =
+    Any { loc; section; header; contents; label_cmt; legacy_labels; errors }
+
+  let make_include ~loc ~section ~labels = Include { loc; section; labels }
+end
+
 type cram_value = { language : [ `Sh | `Bash ]; non_det : Label.non_det option }
 
 type ocaml_value = {
@@ -253,20 +288,20 @@ let version_enabled version =
 
 let get_label f (labels : Label.t list) = Util.List.find_map f labels
 
-let label_not_allowed ~label ~kind =
-  Util.Result.errorf "`%s` label is not allowed for %s blocks." label kind
+let label_not_allowed ~loc ~label ~kind =
+  loc_error ~loc "`%s` label is required for %s blocks." label kind
 
-let label_required ~label ~kind =
-  Util.Result.errorf "`%s` label is required for %s blocks." label kind
+let label_required ~loc ~label ~kind =
+  loc_error ~loc "`%s` label is required for %s blocks." label kind
 
-let check_not_set msg = function
-  | Some _ -> Util.Result.errorf msg
+let check_not_set ~loc msg = function
+  | Some _ -> loc_error ~loc "%s" msg
   | None -> Ok ()
 
-let check_no_errors = function
+let check_no_errors ~loc = function
   | [] -> Ok ()
   | _ :: _ ->
-      Util.Result.errorf "error block cannot be attached to a non-OCaml block"
+      loc_error ~loc "error block cannot be attached to a non-OCaml block"
 
 type block_config = {
   non_det : Label.non_det option;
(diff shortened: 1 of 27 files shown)


NathanReb:
There is a small side effect which is that errors with labels are reported with the entire block location. I think it's not too bad, especially given how precise the error messages usually are, there should be no ambiguity as to where the error comes from.

We can eventually try to refine error locations even further but since we have no editor integration at the moment I see little incentive for it.

NathanReb:
Finally, this does not impact `mli` files at the moment. The implementation of the parsing being quite different I decided to open a PR with I already had.

I'll work on the mli side of things next but doing it separately should ease the review!

lib/lexer_mdx.mll
```diff
@@ -89,44 +65,35 @@ and cram_text section = parse
...
+      { let start = Lexing.lexeme_start_p lexbuf in
+        newline lexbuf;
+        let header = "sh" in
         let requires_empty_line, contents = cram_block lexbuf in
         let contents = first_line :: contents in
-        let labels = [] in
-        let legacy_labels = false in
-        let loc = Location.curr lexbuf in
-        List.iter (fun _ -> newline lexbuf) contents;
+        let label_cmt = Some "" in
```
Leonidas-from-XIV, line 73: Shouldn't this be `None`?

lib/lexer_mdx.mll
```diff
@@ -89,44 +65,35 @@ and cram_text section = parse
...
+        let legacy_labels = "" in
+        let end_ = Lexing.lexeme_start_p lexbuf in
         let rest = cram_text section lexbuf in
+        let loc = loc ~start ~end_ in
         let block =
-          match
-            Block.mk ~loc ~section ~header ~contents ~labels
+            Block.Raw.make ~loc ~section ~header ~contents ~label_cmt
               ~legacy_labels ~errors:[]
```
Leonidas-from-XIV, line 80: I wonder if, when switching from `Block` to `Block.Raw` it wouldn't make sense to make `legacy_labels` into an `option`, instead of using `""` as a sentinel to represent emptiness.

lib/lexer_mdx.mll
```diff
@@ -89,44 +65,35 @@ and cram_text section = parse
...
-          | Error (`Msg msg) -> failwith msg
-        in
-        let legacy_labels = false in
+      { let start = Lexing.lexeme_start_p lexbuf in
         newline lexbuf;
-        let loc = Location.curr lexbuf in
-        List.iter (fun _ -> newline lexbuf) contents;
+        let header = "sh" in
+        let requires_empty_line, contents = cram_block lexbuf in
+        let label_cmt = Some (Printf.sprintf "non-deterministic=%s" choice) in
```
Leonidas-from-XIV, line 89: It's not a big thing, but having helper functions to construct known labels could be useful

lib/lexer_mdx.mll
```diff
@@ -162,12 +124,6 @@ and cram_block = parse
 let cram_token lexbuf =
     try Ok (cram_text None lexbuf)
     with
-    | Failure e ->
-      let loc = Location.curr lexbuf in
-      let msg =
-        Format.asprintf "%a: invalid code block: %s" Stable_printer.Location.print_loc loc e
-      in
-      Util.Result.errorf "%s" msg
```
Leonidas-from-XIV, line 170: Great to be dropping this duplicated code.

lib/block.ml
```diff
@@ -16,6 +16,21 @@
 
 open Util.Result.Infix
 
+let loc_error ~loc fmt =
+  Format.kasprintf
+    (fun s -> Error (`Msg s))
+    ("%a: invalid code block: " ^^ fmt)
+    Stable_printer.Location.print_loc loc
```
Leonidas-from-XIV, line 23: I wonder, would it not be reasonable to rename `print_loc` into `pp`? That way it is very clear what it is.

lib/mdx.ml
```diff
@@ -56,16 +56,27 @@ let filter_section re (t : t) =
...
+        | `Section s -> Ok (Section s)
+        | `Block rb -> Block.from_raw rb >>= fun b -> Ok (Block b))
+      l
+  in
+  let errors =
+    Util.List.concat_map (function Ok _ -> [] | Error l -> l) results
+  in
+  let ok =
+    List.filter_map (function Ok x -> Some x | Error _ -> None) results
+  in
```
Leonidas-from-XIV, line 72: Since `concat_map` has to be supplied by us for compatibility reasons, wouldn't it be better to implement `Base.List.partition_result`/`Base.List.partition_map`? That saves one iteration and is clearer on what it does.

lib/util.ml
```diff
@@ -22,12 +22,15 @@ module Result = struct
     let ( >>! ) r f =
       match r with
       | Ok x -> f x
-      | Error (`Msg e) ->
-          Printf.eprintf "[mdx] Fatal error: %s\n" e;
+      | Error l ->
+          List.iter
+            (function `Msg m -> Printf.eprintf "[mdx] Fatal error: %s\n" m)
+            l;
```
Leonidas-from-XIV, line 28: Purely stylistic but I find it somewhat weird to print a fatal error and keep going, maybe it would make sense to print them as errors and finally conclude that these are fatal and the program will exit?

Or maybe that's just me overthinking semantics.

test/bin/mdx-test/failure/part-unsupported/test-case.md.expected
```diff
@@ -1 +1 @@
-[mdx] Fatal error: File "test-case.md", lines 3-5: invalid code block: `part` label is not allowed for non-OCaml include blocks.
+[mdx] Fatal error: File "test-case.md", lines 3-4: invalid code block: `part` label is required for non-OCaml include blocks.
```
Leonidas-from-XIV, line 1: I am not sure this is correct.

test/lib/test_block.ml
```diff
@@ -25,6 +25,13 @@ let test_mk =
         Mdx.Block.mk ~loc:Location.none ~section:None ~labels
           ~legacy_labels:false ~header ~contents ~errors:[]
       in
+      let expected =
+        Result.map_error
+          (function
```
Leonidas-from-XIV, line 30: I would assume that ``(fun (`Msg m) -> …)`` would work here as well and be a bit simpler?

Leonidas-from-XIV (commented):
Thanks for the PR. I think it is fair to leave the fix for the error locations here, untangling it would probably create more work than it is worth, although it would be somewhat simpler to review.

In general I think it is fair to move the parsing out of the lexer, which also makes programming a bit easier in a file that is actual OCaml syntax instead of `ocamllex`. And well, it is more parsing than lexing anyway.

The code is fine, I don't have anything big to note. Just a few comments on the points that took me longer to understand with suggestions how to improve them.

lib/block.ml
```diff
@@ -254,20 +289,20 @@ let version_enabled version =
 
 let get_label f (labels : Label.t list) = Util.List.find_map f labels
 
-let label_not_allowed ~label ~kind =
-  Util.Result.errorf "`%s` label is not allowed for %s blocks." label kind
+let label_not_allowed ~loc ~label ~kind =
+  loc_error ~loc "`%s` label is required for %s blocks." label kind
```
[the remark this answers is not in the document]
NathanReb: Ouh yeah, probably a copy/paste error! Good catch!

lib/lexer_mdx.mll
NathanReb: No, this is a direct translation of what was done before (i.e. no labels, and legacy_labels is `false`). This basically encode that there were no label without triggering any potential warnings for using the old syntax!

lib/lexer_mdx.mll
NathanReb: Here it's specifically to deal with a special in the cram syntax which is deprecated so I went for the simplest and most straightforward implementation.

lib/lexer_mdx.mll
NathanReb: There might be a better representation but here I merely transmitted the raw data we're extracting in the lexer.

Also note that using options does not rule out `Some ""` so I think in this case, the option is a bit redundant.

NathanReb: Also `""` is not a sentinel, it's the actual value we extract from the lexer when there are no legacy labels!

lib/block.ml
NathanReb: Yeah it would make more sense indeed!

lib/mdx.ml
NathanReb: Yeah good point! I'll see how I can improve it!

lib/util.ml
NathanReb: That's what it's doing, the `1` coming after is the exit code.

test/bin/mdx-test/failure/part-unsupported/test-case.md.expected
NathanReb: Indeed it was not, fixed it!

lib/util.ml
NathanReb: Looking at how this is used by mdx_test and the dune-gen command, it's probably better to move the exit in here as well actually!

NathanReb: After a bit of testing, it seems to work well without any change! We can eventually do this in a separate PR if the need arise!

lib/lexer_mdx.mll
Leonidas-from-XIV: Yes, I guess it depends on whether you consider `""` a valid legacy label or not, but fair enough - if this is what is extracted by the lexer it somewhat makes sense to pass this as-is to the parser to then be interpreted accordingly.

Leonidas-from-XIV (approved):
Looks good, I think it is ready to be merged.

PR #395, opened 2022-08-30
Fix multilines for OCaml block in .mli file

panglesd:
This small PR fixes the fact that, in OCaml blocks, in .mli files, the indentation was erased to a "only 2 space relative to the opening bracket" indentation.

For instance,
```ocaml
(**
{[
  match None with
  | None -> ()
  | Some a -> match a with
    | None -> ()
    | Some _ -> ()
]}
*)
```
was turned into
```ocaml
(**
{[
  match None with
  | None -> ()
  | Some a -> match a with
  | None -> ()
  | Some _ -> ()
]}
*)
```
which is not ideal.

The first commit adds a test, showing the wrong behaviour,. The second commit is the patch and updates the test.
The new behaviour is the first line is indented two spaces after the opening bracket, and for other lines the indentation relative to the first one is kept, as long as they are at least as indented as the first one. In the previous example, the exact indententation would be kept. 

In the test, I also added that multilines in .mli files for toplevel block does not work!

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -162,8 +162,7 @@ let lstrip strings =
   let hpad = Misc.hpad_of_lines strings in
   List.map
     (fun string ->
-      let min a b = if a < b then a else b in
-      let first = min (Misc.hpad_of_lines [ string ]) hpad in
+      let first = Util.Int.min (Misc.hpad_of_lines [ string ]) hpad in
       Astring.String.with_index_range string ~first)
     strings
 
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -158,15 +158,20 @@ let pp_lines syntax t =
   in
   Fmt.(list ~sep:(any "\n") pp)
 
-let lstrip string =
-  let hpad = Misc.hpad_of_lines [ string ] in
-  Astring.String.with_index_range string ~first:hpad
+let lstrip strings =
+  let hpad = Misc.hpad_of_lines strings in
+  List.map
+    (fun string ->
+      let min a b = if a < b then a else b in
+      let first = min (Misc.hpad_of_lines [ string ]) hpad in
+      Astring.String.with_index_range string ~first)
+    strings
 
 let pp_contents ?syntax ppf t =
   match (syntax, t.contents) with
   | Some Syntax.Mli, [ line ] -> Fmt.pf ppf "%s" line
   | Some Syntax.Mli, lines ->
-      Fmt.pf ppf "@\n%a@\n" (pp_lines syntax t) (List.map lstrip lines)
+      Fmt.pf ppf "@\n%a@\n" (pp_lines syntax t) (lstrip lines)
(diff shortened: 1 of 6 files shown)


test/bin/mdx-test/expect/multilines-mli/test-case.mli
```diff
@@ -0,0 +1,45 @@
...
+
+  {[
+    match None with
+    | None -> ()
+    | Some a -> match a with
+      | None -> ()
+      | Some b -> b
+  ]}
+
+But it does not work fine for toplevel (see [multilines/test-case.md] for the
```
Leonidas-from-XIV, line 37: I don't think we should have a test making sure things fail unexpectedly. It makes more sense to have such a test as part of a test case fixing the issue.

lib/block.ml
```diff
@@ -158,15 +158,20 @@ let pp_lines syntax t =
...
   Fmt.(list ~sep:(any "\n") pp)
 
-let lstrip string =
-  let hpad = Misc.hpad_of_lines [ string ] in
-  Astring.String.with_index_range string ~first:hpad
+let lstrip strings =
+  let hpad = Misc.hpad_of_lines strings in
+  List.map
+    (fun string ->
+      let min a b = if a < b then a else b in
```
Leonidas-from-XIV, line 165: Given this was added in OCaml 4.13, I would suggest adding it to `Util`, as `Util.Int.min` with the same signature as the Stdlib, so if we end up dropping compatibility with older versions we can easily remove it.

Leonidas-from-XIV (changes requested):
Thanks @panglesd for your contribution! It looks nice and we should be able to merge it soon, there's just some small things that would be nice to improve beforehand.

Leonidas-from-XIV:
Oh, as I am seeing the CI runs, can you add a changelog entry?

test/bin/mdx-test/expect/multilines-mli/test-case.mli
panglesd: You are right. I sometimes read the tests to understand how a binary/library work, and can be frustrated when only the small working subset is tested (not covering my failure), but an issue will be better to track the problem.

Leonidas-from-XIV: I definitely agree with you, this should be fixed. But I find tests that make sure the code fails to be more confusing than helpful, as they make me wonder: should this fail? Kind of like you have unit tests, you fix something and suddenly a test fails because it was expecting the value to be wrong which makes you wonder whether you fixed it or you broke it.

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
Thanks @panglesd for your help, I'll try to bundle a few of these fixes that have PRs and make a release soon.

Alnsim69:
https://github.com/realworldocaml/mdx/pull/397#discussion_r987613121

PR #398, opened 2022-09-28
Undo the `;;` deprecation

Leonidas-from-XIV:
As described in https://github.com/tarides/platform-roadmap/issues/45 this undoes the `;;` deprecation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,7 +24,7 @@
 
 - Removed warning about missing semicolons added in MDX 1.11.0 and the
   automatic insertion of semicolons in the corrected files introduced in MDX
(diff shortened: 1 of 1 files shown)


talex5:
https://github.com/tarides/platform-roadmap/issues/45 gives a 404 error for me. Would be good to document the reason for this change somewhere public.

Leonidas-from-XIV:
I think it's ok to quote the roadmap item summary for a description of this change:

> In a previous version of mdx, code blocks that do not end with a double semicolon (;;) have been deprecated. However, this has been criticized as creating a poor user experience, as it can make code less readable when semicolons are added to all code blocks. We want to remove this deprecation to improve the user experience of mdx.

PR #399, opened 2022-09-28
Use `let*` operators instead of infix operators

Leonidas-from-XIV:
At least in most places. Where it is used as piping operator, it still makes sense to use the infix versions.

Given our lower bound for OCaml is 4.08, it should be okay to use them and I feel them make the code a bit easier to read, since you get to avoid `>>= fun xyzzy ->` overhead.


gpetiot (approved):
It looks better indeed!

PR #405, opened 2022-11-09
Replace `egrep` with `grep -E`, the former is deprecated

MisterDA:


Leonidas-from-XIV:
I didn't know `egrep` was [deprecated](https://unix.stackexchange.com/questions/418009/availability-of-deprecated-egrep-and-fgrep-commands) but apparently that's true!

Leonidas-from-XIV (approved):
Ok to merge when CI finishes. Thanks for the contribution!

Leonidas-from-XIV:
Checked on a machine with macOS, they also have `grep -E`, so merged since CI was stuck waiting for a macOS worker.

PR #406, opened 2022-11-09
Update to Dune 3, fix new warnings

MisterDA:


Leonidas-from-XIV:
I am not sure how this happens because the values indeed don't seem to be read, but the tests fail, can you take a look what is going on there?

MisterDA:
The failures are due to Dune 3.0. This PR categorizes the cram tests as "misbehaving" ;)
https://github.com/ocaml/dune/pull/4981
Apparently you shouldn't write syntactically incorrect cram tests or unreachable code in cram tests.

Leonidas-from-XIV:
But the tests aren't supposed to be run by dune `cram`, these are `ocaml-mdx test` cram tests (which support things like `...` etc), so dune shouldn't execute them. Pretty sure that even before Dune 3.0 they weren't valid Dune Cram tests.

According to the docs, these are [automatically enabled in Dune 3.0](https://dune.readthedocs.io/en/stable/tests.html#cram-tests-1), so I think they have to be explicitly turned off in this project otherwise dune will attempt to run the MDX cram tests through its own Cram implementation.

Leonidas-from-XIV:
Great, thanks!

PR #407, opened 2022-11-09
Optimize `ansi_color_strip` perf by copying substrings

MisterDA:
cf https://github.com/ocaml/dune/pull/6339


Leonidas-from-XIV:
Thanks. While I don't think we need the performance improvements at the expense of slightly less readable code, might as well stay in sync with the code in dune.

Can you please reformat?

MisterDA:
> Can you please reformat?

Done!

MisterDA:
> While I don't think we need the performance improvements at the expense of slightly less readable code,

At least it's not premature optimization ;)

PR #411, opened 2022-12-19
Add note about using mdx with mli files

mooreryan:
By default, Dune only processes markdown files with mdx.  This note gives an example of how to use Dune and mdx in mli files.

It's just a small change to the readme, but if it was there, it would have saved me a few minutes skimming through the Dune docs to figure out what to do.  Do you think it is a helpful addition to the readme?


Leonidas-from-XIV:
Sounds good to me, thanks for your contribution!

PR #413, opened 2023-01-10
Work around lack of `Unix.set_nonblock` on Windows using threads

polytypic:
On Windows `Unix.set_nonblock` fails.

This PR implements a workaround method for pulling the output without blocking by using threads.  I can't say I like the workaround, but implementing proper non-blocking IO on Windows would seem to require a lot more work.  This workaround has been tested at least on Mac OS and Windows.


polytypic:
I'm closing this as reverting [PR403](https://github.com/realworldocaml/mdx/pull/403) would be a much simpler solution.

Leonidas-from-XIV:
#414 merged to undo the issues that we had with #403.

PR #414, opened 2023-01-10
Revert "Rewrite redirect code with pipe and cloexec" #403 

MisterDA:
This reverts commit https://github.com/realworldocaml/mdx/commit/de43f19ab1f3e37acc439fb27a54fb7ea978935d, introduced in "Rewrite redirect code
with pipe and cloexec" #403.

It is not possible (or difficult) to set non-blocking mode on pipes on
Windows. Moreover, using pipes instead of a file as a buffer between
subprocesses makes it harder to investigate hangs and data being
transferred.

I've added explicit close-on-exec flags from the previous PR.

### DIFF
--- a/lib/top/mdx_top.ml
+++ b/lib/top/mdx_top.ml
@@ -20,36 +20,34 @@ open Compat_top
 type directive = Directory of string | Load of string
 
 let redirect ~f =
-  let stdout_backup = Unix.dup ~cloexec:true Unix.stdout in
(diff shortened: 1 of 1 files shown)


polytypic (approved):
I tested that, with this PR, `mdx> dune test` runs to completion successfully on Windows. 🚀

Leonidas-from-XIV:
@MisterDA @polytypic Thanks for the input. I'll try to wrestle the CI to get it to build.

PR #417, opened 2023-02-02
tiny change to allow using the parser that the toplevel uses

chetmurthy:
The toplevel parser can be swapped-out, and this patch allows to use whatever it is, not just the official OCaml parser.

If you guys could merge this, I can use `mdx` with Camlp5, which would be lovely!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,8 @@
 ### unreleased
 
-- tiny change to allow using the parser that the toplevel uses
-  (instead of always the official OCaml parser) (#417, @chetmurthy)
+- Switch to using the parser that toplevel uses (found in a mutable
+  `ref`, instead of always the official OCaml parser).  This allows
+  Camlp5's parser to be used with MDX. (#417, @chetmurthy)
 
 #### Added
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ### unreleased
 
+- tiny change to allow using the parser that the toplevel uses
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,5 +1,8 @@
 ### unreleased
 
+- tiny change to allow using the parser that the toplevel uses
```
Julow, line 3: Users don't care whether it's a tiny change. Perhaps you can add a specific use case, for example "Allow to use Camlp5's parser".

Julow (approved):
Given that we use `Toploop` for evaluation, it makes perfect sense to use its parser too.

If you have a specific use case, do you think you could write a test for it ?

CHANGES.md
Leonidas-from-XIV: Yes, that would be a better changelog entry. The idea is that these entries should be from a users perspective, like what can they do now, what can't they do anymore, maybe in which cases the bug manifested, something like this.

chetmurthy: I updated the Changelog entry.

As to writing a test, here's an example of how I use MDX (with Camlp5): https://github.com/camlp5/pa_ppx_perl/blob/main/README.asciidoc

BUT: to use MDX here with Camlp5, requires pulling in all of Camlp5.  After all, if we're going to replace the OCaml toplevel's parser, we need to replace it with .... a full parser for OCaml or something equivalent, right?  So I'm not sure how to provide a small testcase.  FWIW, now that I know about MDX, I plan to use it in Camlp5's unit-tests, so eventually there will be test-coverage via that path.

This stuff is like sliced bread!

PR #418, opened 2023-02-15
Parse env only once

Leonidas-from-XIV:
Reported by @emillon, the way the code currently works is to call `Mdx.Prelude.env_and_file` in the CLI, validate it is valid, then discard it and reparse it at a later point in time in `Mdx_test`. This is wasteful but most importantly also confusing.

There is the open question whether `Prelude_str` should support environments as well. At the moment it does and I kept the support for it in this PR but it would be better to either remove it if we deem that support unnecessary or like in `Prelude` move the parsing step to the command line parser.

### DIFF
--- a/bin/cli.ml
+++ b/bin/cli.ml
@@ -68,9 +68,11 @@ let verbose_findlib =
 
 let prelude =
   let parse s =
-    let env, file = Mdx.Prelude.env_and_file s in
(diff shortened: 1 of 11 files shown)


gpetiot (approved):
It makes sense to parse it only once indeed.

Looks good to me!

PR #422, opened 2023-04-12
Simplify mli_parser

jonludlam:
Use odoc's location conversion to simplify the extraction of code blocks from mli files. Also generally tidy it up and add some documentation.

### DIFF
--- a/lib/mli_parser.ml
+++ b/lib/mli_parser.ml
@@ -20,11 +20,13 @@ end
 
    The results are prepended in reverse order onto [acc]. *)
 let extract_code_block_info acc ~(location : Lexing.position) ~docstring =
-  let open Odoc_parser in
-  let parsed = parse_comment ~location ~text:docstring in
+  let module O = Odoc_parser in
+  let parsed = O.parse_comment ~location ~text:docstring in
 
   (* If odoc-parser produced any warnings, we raise them as errors here *)
-  List.iter (fun error -> failwith (Warning.to_string error)) (warnings parsed);
+  List.iter
+    (fun error -> failwith (O.Warning.to_string error))
+    (O.warnings parsed);
 
   (* Extract the useful info from what odoc has given us.
 
@@ -33,21 +35,21 @@ let extract_code_block_info acc ~(location : Lexing.position) ~docstring =
      Fortunately the location info give us enough info to be able to extract
      the code from the original text, whitespace and all.
   *)
-  let handle_code_block : Loc.span -> _ -> Code_block.t =
-    let convert_loc (sp : Loc.span) =
+  let handle_code_block : O.Loc.span -> _ -> Code_block.t =
+    let convert_loc (sp : O.Loc.span) =
       Location.
         {
-          loc_start = position_of_point parsed sp.start;
-          loc_end = position_of_point parsed sp.end_;
+          loc_start = O.position_of_point parsed sp.start;
+          loc_end = O.position_of_point parsed sp.end_;
           loc_ghost = false;
         }
     in
-    fun location (metadata, { Loc.location = span; _ }) ->
+    fun location (metadata, { O.Loc.location = span; _ }) ->
       let metadata =
         Option.map
           (fun (lang, labels) ->
-            let language_tag = Loc.value lang in
-            let labels = Option.map Loc.value labels in
+            let language_tag = O.Loc.value lang in
+            let labels = Option.map O.Loc.value labels in
             Code_block.{ language_tag; labels })
           metadata
       in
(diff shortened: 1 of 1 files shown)


jonludlam:
I don't think a changelog entry is needed for this.

lib/mli_parser.ml
```diff
@@ -1,207 +1,128 @@
...
-       2 ---[---
-       3 ---------
-       4 --
-       5 ----------
-       6 -------]--
-       7 ----------
-       8 ----------
+   The results are prepended in reverse order onto [acc]. *)
+let extract_code_block_info acc ~(location : Lexing.position) ~docstring =
+  let open Odoc_parser in
```
Leonidas-from-XIV, line 23: How bad would the code look without that `open`? I find code using `open` often hard to reason about and try to avoid it if possible.

lib/mli_parser.ml
```diff
@@ -1,207 +1,128 @@
...
-       from lines 2-6 ([lines_to_include]). *)
-  | Some (first_line, Some (stripped, last_line)) ->
-      let first_line =
-        String.sub first_line start.column
-          (String.length first_line - start.column)
-      in
-      let last_line = String.sub last_line 0 end_.column in
-      String.concat "\n" ([ first_line ] @ stripped @ [ last_line ])
+  (* If odoc-parser produced any warnings, we raise them as errors here *)
+  List.iter (fun error -> failwith (Warning.to_string error)) (warnings parsed);
```
Leonidas-from-XIV, line 27: Maybe it would be nicer to turn this into a `result` instead of terminating mdx, especially as `Failure` is annoying to catch.

lib/mli_parser.ml
```diff
@@ -1,207 +1,128 @@
...
-        (fun (b : Code_block.t) -> (b, convert_loc cmt_loc b.location))
-        blocks)
-    (docstrings (Lexing.from_string str))
-  |> List.concat
-
-let make_block ~loc code_block =
+(* Given code block metadata and the original file, this function splices the
+   contents of the code block from the original text and creates an Mdx
+   Block.t, or reports the error (e.g., from invalid tags) *)
+let make_block code_block orig =
```
Leonidas-from-XIV, line 113: I think `orig` is a poor name, maybe `let extract_block code_block contents`?

Leonidas-from-XIV (changes requested):
Very cool, I think it is a significant improvement over the mess I created when attempting to extract the content with whitespace!

I have a few minor comments that I would hope make the understanding a bit easier, but structure-wise this looks very good.

lib/mli_parser.ml
```diff
@@ -1,207 +1,128 @@
...
-    | Some offset -> offset
+  (* Fold over the results from odoc-parser, recurse where necessary
+     and extract the code block metadata *)
+  let rec fold_fn acc (elt : Ast.block_element Loc.with_location) =
+    match elt with
+    | { Loc.value = `Code_block c; location } ->
+        handle_code_block location c :: acc
+    | { Loc.value = `List (_, _, lists); _ } ->
+        List.fold_left (List.fold_left fold_fn) acc (lists :> Ast.t list)
+    | { value = `Tag tag; _ } -> (
```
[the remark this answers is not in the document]
jonludlam: I think the old code _did_ look through those blocks:

```ocaml
| `See (_, _, blocks) -> acc blocks
```

where `acc` takes the blocks list and finds `Code_block`s in them. I found it confusing when I was reading it as I tend to use `acc` as an accumulator, and returning `acc` reads to me like we're ignoring that case - that's partly why I rewrote this bit.

lib/mli_parser.ml
jonludlam: It just looked a little noisy. I was definitely against `open`s in the past, but these days I find merlin really helps disambiguation. I kept it local to the function that requires it, but if you're opposed I'll take it out.

lib/mli_parser.ml
```diff
@@ -1,207 +1,128 @@
...
-let docstrings lexbuf =
+  List.fold_left fold_fn acc (ast parsed)
+
+(* This function handles string containing ocaml code. It parses it as ocaml
+   via compiler-libs, then for each odoc-formatted comment it then parses
+   that via odoc-parser. The end result is a list of metadata about the code
+   blocks within the comments. The result is given as an in-order list of
+   [Code_block.t] values. *)
+let docstring_code_blocks str =
+  Lexer.handle_docstrings := true;
```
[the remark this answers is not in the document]
jonludlam: Sure. As I said, I kept the original semantics for this commit, and then we can change behaviour in a separate commit. I'll update this PR.

lib/mli_parser.ml
jonludlam: Agreed! I'll change it :-)

lib/mli_parser.ml
jonludlam: Removing the `open Code_block` was not a problem, happy to do that. Removing the `open Odoc_parser` led to a lot of additional noise, IMO. Here's a commit with the changes: https://github.com/jonludlam/mdx/commit/dcce65ec579364b9eef0d48493252deaf2a6afd2

lib/mli_parser.ml
jonludlam: I think `file_contents` is better - that's actually the name of the variable passed to this function below.

lib/mli_parser.ml
jonludlam: This is a slightly larger change that probably deserves a separate PR I think? Changing this would merit a changelog entry at least :-)

Leonidas-from-XIV: Ah, I didn't see that this code was in the original parser as well, it has moved places quite a bit. Agreed, as this is no change (and turning it into a result requites threading through the result monad) I think this is sufficient for now and improving on this can be done at a later point in time.

lib/mli_parser.ml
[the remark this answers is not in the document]
Leonidas-from-XIV: Indeed, the original code looks at the tags twice, I only looked at the case in `acc` (which is horribly named). This is good then :)

Leonidas-from-XIV:
@jonludlam I think you accidentally pushed the mld support commit. If you remove it the code is good to be merged in my opinion.

jonludlam:
gah! so I did...

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
Great, thanks!

PR #430, opened 2023-06-20
Vendor odoc-parser

jonludlam:
This PR vendors odoc-parser into the repo. This helps prevent the use of mdx in odoc's tests. In this I've vendored the latest release of odoc-parser, but once [ocaml/odoc#973](https://github.com/ocaml/odoc/pull/973) is merged we'll want to adjust the script to point at the odoc repo rather than the odoc-parser repo.


vendor/update-odoc-parser.sh
```diff
@@ -0,0 +1,26 @@
...
+    git checkout $version
+)
+
+SRC=$TMP/odoc-parser
+
+cp -v $SRC/src/*.{ml,mli,mll} odoc-parser/src
+cp -v $SRC/LICENSE.md odoc-parser/
+
+git checkout odoc-parser/src/dune
+git add -A .
```
panglesd, line 26: I'm not sure, what is this line for?

panglesd (approved):
The update-odoc-parser.sh script will need quite a lot of modifications when upgrading the vendored odoc-parser, but it is still good to keep it!

vendor/update-odoc-parser.sh
jonludlam: Which, the git add? I should mention that this script was modelled closely on (copied from!) dune's equilvalent scripts, e.g. https://github.com/ocaml/dune/blob/main/vendor/update-cmdliner.sh

panglesd: Yes, I was wondering why automatically adding those files to the git index... (I'm OK with it, I just wanted to know if there was a specific reason you knew for this)

vendor/update-odoc-parser.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+version=ebfd3b9489e44187da2c67d79a32b6fc1e92bda4
```
gpetiot, line 3:
```suggestion
version=5ac1ffc67ce1b96f5a990fa4902a157a5cdb42d0
```

vendor/update-odoc-parser.sh
```diff
@@ -0,0 +1,26 @@
...
+
+TMP="$(mktemp -d)"
+trap "rm -rf $TMP" EXIT
+
+rm -rf odoc-parser
+mkdir -p odoc-parser/src
+
+(
+    cd $TMP
+    git clone https://github.com/ocaml-doc/odoc-parser.git
```
gpetiot, line 15:
```suggestion
    git clone https://github.com/ocaml/odoc.git
```

vendor/update-odoc-parser.sh
```diff
@@ -0,0 +1,26 @@
...
+    cd odoc-parser
+    git checkout $version
+)
+
+SRC=$TMP/odoc-parser
+
+cp -v $SRC/src/*.{ml,mli,mll} odoc-parser/src
```
gpetiot, line 22:
```suggestion
cp -v $SRC/src/parser/*.{ml,mli,mll} odoc-parser/src
```

vendor/update-odoc-parser.sh
```diff
@@ -0,0 +1,26 @@
...
+TMP="$(mktemp -d)"
+trap "rm -rf $TMP" EXIT
+
+rm -rf odoc-parser
+mkdir -p odoc-parser/src
+
+(
+    cd $TMP
+    git clone https://github.com/ocaml-doc/odoc-parser.git
+    cd odoc-parser
```
gpetiot, line 16:
```suggestion
    cd odoc
```

vendor/update-odoc-parser.sh
```diff
@@ -0,0 +1,26 @@
...
+    git checkout $version
+)
+
+SRC=$TMP/odoc-parser
```
gpetiot, line 20:
```suggestion
SRC=$TMP/odoc
```

gpetiot (commented):
I updated `vendor/update-odoc-parser.sh` to make it point to the odoc repo since the parser has been merged back.
Btw would it be possible to use git submodules for all this?

panglesd:
> to make it point to the odoc repo since the parser has been merged back.

More modification would be needed since the new AST has changed the code block node.

Moreover, one would want to use the new syntax for "output of code blocks". Indeed, it would fix a bug where an error in the execution leads to:

``````
{[
  let f = 1 + "2"
]}```mdx-error
An expression of type int was expected, but got a string
```
``````
which is not valid odoc code.

jonludlam:
I think we should merge this as-is, release a version of mdx, then do the work to upgrade to the new parser API.

trefis:
I agree with this plan.

gpetiot:
> Moreover, one would want to use the new syntax for "output of code blocks". Indeed, it would fix a bug where an error in the execution leads to:
> 
> ```
> {[
>   let f = 1 + "2"
> ]}```mdx-error
> An expression of type int was expected, but got a string
> ```

I had a look at it, did I miss something regarding the new syntax for the output of code blocks? this is not mentioned in the [docs](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#ocaml_code_blocks)

panglesd:
More info is available here: https://github.com/ocaml-doc/odoc-parser/pull/17

I don't know if the new syntax is documented in doco `master`'s branch, but it's normal that it is not in the doc for the released version!

Issue #393, opened 2022-08-29
Conditionally skip blocks


talex5:
I'd like to skip certain blocks/sections, where the test is performed at runtime. For example, when running on macos in the opam sandbox we can't run any tests that use the network, and when running in a Docker container we might not be able to use IPv6.

Is there a good way to do this already?

If not, what's the best way to implement this? Perhaps if a block raises `Mdx_skip` then it could just be ignored?

Leonidas-from-XIV:
You mean, only correct the block unless it raises Mdx_skip? I wonder if this isn't adding functionality to MDX that would better be solved on another level, e.g. do not run the MDX test via dune in some conditions.

talex5:
Yes. For example, oUnit has a `skip_if` function that raises a special exception (https://github.com/gildor478/ounit/blob/master/src/lib/ounit2/advanced/oUnitAssert.ml#L37), and crowbar has a similar `guard` function (https://github.com/stedolan/crowbar/blob/0cbe3ea7e990a7d233360e6a74b1cb5e712501ad/src/crowbar.ml#L93).

I don't think it can be solved easily with dune. For example, most of the Eio README file should still be tested by opam-repo-ci, just not the parts using the network when run on macos within the opam sandbox.

PR #433, opened 2023-07-27
Add `os_type` label to enable/disable based on `Sys.os_type`

polytypic:
This PR adds ability to skip blocks by comparing against `Sys.os_type`.

The desire to have something like this arose when [working to allow Eio's MDX based tests to run on Windows](https://github.com/ocaml-multicore/eio/pull/589) and I'm using this feature there.

Also related is the [issue that proposes a more general ability to skip blocks](https://github.com/realworldocaml/mdx/issues/393).  The feature that this proposes is less ambitious (less general) and very similar to the existing OCaml `version` constraint supported by MDX and the implementation follows the same pattern.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 #### Added
 
+- Add `os_type` label to enable/disable based on `Sys.os_type` (#433,
+  @polytypic)
+
 - Make MDX compatible with OCaml 5.1 (#435, @polytypic and @kit-ty-kate)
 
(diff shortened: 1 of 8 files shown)


trefis:
The CI failures are all 5.1 related, but we just merged your other PR adding support for 5.1.
So if you rebase and take care of the conflict in the changelog, this should be good to merge.

polytypic:
I rebased the PR.

PR #439, opened 2023-09-29
Upgrade mdx to use last version of odoc-parser

gpetiot:
Follow-up on #430

- [x] update script updated
- [x] vendored parser updated
- [x] mdx compiles
- [x] use the new [syntax](https://github.com/ocaml-doc/odoc-parser/pull/17) for block output

Re-using code from https://github.com/realworldocaml/mdx/compare/main...jonludlam:mdx:vendor-odoc-parser-richmdx


lib/block.ml
```diff
@@ -160,14 +160,20 @@ let rec error_padding = function
...
-      let errors = error_padding errors in
-      Fmt.pf ppf "```mdx-error\n%a\n```\n"
-        Fmt.(list ~sep:(any "\n") Output.pp)
-        errors
+  | OCaml { errors; _ } -> (
+      match syntax with
+      | Some Syntax.Mli | Some Syntax.Mld ->
+          let errors = error_padding errors in
+          Fmt.pf ppf "][\n%a\n" Fmt.(list ~sep:(any "\n") Output.pp) errors
+      | Some Syntax.Cram -> ()
```
panglesd, line 171: i think some syntax for cram output should be invented (or at least keep the markdown syntax to not integrate regression!)

jonludlam:
I don't think we should delete the license file from the vendored parser

Leonidas-from-XIV:
It's weird that it doesn't fail, given there is a line to copy `LICENSE.md` and the script is `set -e`.

gpetiot:
It's now copying the license file from odoc (longer than the previous one)

lib/block.ml
```diff
@@ -160,19 +161,45 @@ let rec error_padding = function
...
 
-let pp_errors ppf t =
+let pp_error ?syntax ?delim ppf outputs =
+  match syntax with
+  | Some Syntax.Markdown ->
+      Fmt.pf ppf "```\n```mdx-error\n%a\n"
+        Fmt.(list ~sep:(any "\n") Output.pp)
+        outputs
+  | Some Syntax.Mli | Some Syntax.Mld ->
+      Fmt.pf ppf "]%a[\n{err@mdx-error[\n%a\n]err}"
```
panglesd, line 171: I think it would be nice to check that the error does not contain the string `]err}`. It's unlikely to happen, but easy to check and avoids prematurely closing the block!

lib/dune
```diff
@@ -12,6 +12,7 @@
   ocaml-version
   odoc_parser
   re
+  result
```
panglesd, line 15: I know it was introduced in #430, but I think we can remove dependency on this library altogether.

> Projects that want to use the new result type defined in OCaml >= 4.03 while staying compatible with older version of OCaml should use the Result module defined in this library.

Mdx works with `ocaml >= 4.08`...

panglesd (approved):
Looking forward to not having syntax errors after mdx reports an error in an mld file! :star_struck:

lib/block.ml
gpetiot: That's a good idea, done.

lib/block.ml
```diff
@@ -171,7 +171,7 @@ let pp_error ?syntax ?delim ppf outputs =
       Fmt.pf ppf "]%a[\n{err@mdx-error[\n%a\n]err}"
         Fmt.(option string)
         delim
-        Fmt.(list ~sep:(any "\n") Output.pp)
+        Fmt.(list ~sep:(any "\n") Output.(pp ~escape_strings:[ "]err}" ]))
```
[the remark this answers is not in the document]
gpetiot: This sounds like over-engineering to me, if we can fully decide what the delimiter is, we could just have something like `mdx_error`, users should never prefix anything with mdx.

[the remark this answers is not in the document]
panglesd: Yes, we can fully decide what the delimiter is, but I don't see why limit people.

This code:
```
let compute_delimiter s =
  let base_delim = "mdx_error" in
  let delim n =
    if n = 0 then base_delim
    else Format.sprintf "%s_%d" base_delim n in
  let is_inadequate delim = String.is_infix ~affix:("]" ^ delim ^ "}") s in
  let rec loop n =
    let delim = delim n in
    if is_inadequate delim then loop (n+1)
    else delim
  in
  loop 0
```
seems less error-prone than the escaping business, and uses delimiters how they were meant to be used!

lib/block.ml
```diff
@@ -160,19 +161,64 @@ let rec error_padding = function
...
+  | Some base_delim -> Some (loop ~base_delim 0)
+
+let pp_error ?syntax ?delim ppf outputs =
+  match syntax with
+  | Some Syntax.Markdown ->
+      Fmt.pf ppf "```\n```mdx-error\n%a\n"
+        Fmt.(list ~sep:(any "\n") Output.pp)
+        outputs
+  | Some Syntax.Mli | Some Syntax.Mld ->
+      let delim = compute_delimiter ?delim outputs in
```
panglesd, line 189: You are modifying the wrong delimiter: the one given by the user, instead of `err` (or I would prefer the more explicit `mdx_error` that you suggested).

If `outputs` contains the delimiter given by the user, it breaks the syntax...

gpetiot: Makes sense, should be good now.

panglesd: Yes!

panglesd (approved):
Looks good and ready to be merged to me!

Issue #421, opened 2023-03-27
Support for silent blocks


Lupus:
It would be nice to have support for blocks that evaluate, but do not produce any markdown output.

I'm writing some tutorial for a library in mdx, and need to have a snippet of code of 20-30 lines to present to the reader. Writing it in toplevel block directly is not very convenient as it won't get automatically formatted. Making it a file block solves the problem of formatting, but the code is not evaluated in mdx context. My best approach so far it to load it into toplevel via `#use` directive, but it outputs a lot of signatures along the way. Once loaded, I can present separate sections of the file via file blocks and then show some toplevel evaluation of the constructs defined in the file. With silent block I could load it without extra noise in the markdown output.

Leonidas-from-XIV:
I am a bit confused, given the idea of MDX is to essentially present a fixed point of evaluation (modulo non-deterministic blocks), in that case not rendering the block would make it disappear from the next run. That's also why MDX files can be promoted using dune, so documents using silent blocks would silently promote into documents missing these blocks altogether.

Also, if it doesn't produce any output, how would error reporting work, if the block never renders anything? So whether it is correct or not would not yield any feedback to the documentation writer.

Lupus:
Silent block changes the environment and if consecutive blocks depend on it, diff & promotion workflow will work as expected. So far I worked around this via `summary`/`details` markdown tags (see [this gist](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab)) to collapse the output of `#require` and `#use` at the beginning of my markdown file and hide this from the reader.

Leonidas-from-XIV:
How will it work? If the block is silent, it will be missing in the output, thus promotion will promote a version that doesn't feature it, thus breaking the document.

Lupus:
I currently have the following block at the start of my mdx (using 2 backticks to not mess the outer markdown):
```
``ocaml
# #require "ppx_yojson_conv";;
# #require "my-lib";;
# #use "test/Mdx_parts.ml";;
``
```

It does produce some output, as `Mdx_parts.ml` has some modules and functions defined, but I want to hide inferred signatures from the reader at this point.

Later on in my document I expand on some subject, and output a relevant piece of `Mdx_parts.ml` like this:

```
<!-- $MDX file=test/Mdx_parts.ml,part=myCodePart -->
``ocaml
``
```

This shows a piece of code to the reader, and after I presented this code, I want to show how one can use it with toplevel blocks.

[...]

PR #443, opened 2023-12-02
fix: make directives usable in the `ocaml` mode

sorawee:
Prior this PR, the following code fails:

```
{@ocaml ocaml[
  #require "astring";;
  let x = Astring.strf;;
]}
```

because MDX incorrectly infers that the code block is a toplevel interaction (from the fact that the block starts with `#`), resulting in an error:

    incomplete toplevel entry: unexpected character '#'.
    Did you forget a space after the '#' at the start of the line

It is possible to workaround this issue as suggested in #421 by adding a comment as a first line.

```
{@ocaml ocaml[
  (* This works! *)
  #require "astring";;
  let x = Astring.strf;;
]}
```

but ideally the workaround should not be needed.

One may wonder why the inference is needed, since the above code block is already specified to be in the `ocaml` mode.
The answer appears to be that we are expected to use the inference heuristics for a light sanity check, as the existing tests
("invalid ocaml" and "invalid toplevel" in `test_block.ml`) require:

    "let x = 2;;" in the toplevel mode should error with
    invalid toplevel syntax in toplevel blocks

    "# let x = 2;;" in the ocaml mode should error with
    toplevel syntax is not allowed in OCaml blocks

As a result, this PR keeps the light sanity check intact, but adjusts the inference heuristics to be more conservative. A block is now considered a toplevel interaction when it starts with `#` followed by a space.
This fixes the issue, making it possible to use directives. As a bonus, directives will now also work even when the mode is not specified at all. But one disadvantage is that this kind of code will no longer be considered invalid.

    #1+1;;
    ...
    #2+2;;
    ...

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -300,7 +300,7 @@ let guess_ocaml_kind contents =
     | h :: t ->
         let h = String.trim h in
         if h = "" then aux t
-        else if String.length h > 1 && h.[0] = '#' then `Toplevel
+        else if String.length h > 2 && h.[0] = '#' && h.[1] = ' ' then `Toplevel
         else `Code
   in
   aux contents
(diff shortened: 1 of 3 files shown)


lib/block.ml
```diff
@@ -300,7 +300,7 @@ let guess_ocaml_kind contents =
     | h :: t ->
         let h = String.trim h in
         if h = "" then aux t
-        else if String.length h > 1 && h.[0] = '#' then `Toplevel
+        else if String.length h > 2 && h.[0] = '#' && h.[1] = ' ' then `Toplevel
```
Julow, line 303: If needed, the check could be more strict as top-level directives necessarily start with a lowercase ASCII letter (or the sequence `\#` which is never going to be used).

samoht:
Do you mind rebasing your changes? Thanks!

sorawee:
My apologies, @samoht. I didn't see the message at the time, and also completely forgot that I have this PR open. I verified that the issue still exist in the current version, rebased, and also made a minor refactoring.

Issue #420, opened 2023-03-21
MDX hangs when closing fence is missing


talex5:
If the input file is missing a closing fence then MDX hangs:

    $ cat > test.md << 'EOF'
    ```ocaml
    # 1;;
    EOF

    $ ocaml-mdx test test.md
    [hangs]

talex5:
Looks like this was fixed by #444 - thanks!

PR #444, opened 2024-01-17
Avoid infinite loop in lexer on unclosed code block

edwintorok:
An unclosed code block will cause the lexer to loop infinitely in the 'block' rule, because 'eof' is part of the 'eol' rule and it'll keep matching the last rule.

Add a separate rule for eof that terminates immediately, the way the first rule was written it would expect closing quotes after eof which would never match eof.


patricoferris:
This should fix #420 -- We hit this in Eio whilst debugging some other issues, would be a good thing to have merged.

PR #445, opened 2024-01-17
markdown: parse fenced_code_attributes extension

edwintorok:
Pandoc supports this extension: https://pandoc.org/MANUAL.html#extension-fenced_code_attributes
``` {#identifier .language attr="value"}

```

And this:
``` language {#identifier attr="value"}
```

Recognize them in the lexer. Try to limit the complexity of the regular expression by splitting off parsing of attributes into a separate 'parse' (otherwise we hit automata size limits in `ocamllex`).

According to
https://quarto.org/docs/authoring/markdown-basics.html#ordering-of-attributes the ordering has to be:
* #identifiers
* .classes
* key-value attributes

For now on output we always normalize to this form (which isn't ideal, but could be improved later):
```language {#identifierM .otherclasses attrN="valueN"}
```

I initially tried to fully parse the attributes, but I've exceeded the maximum size of the `ocamllex` automaton, so I kept it simple in this PR (and do some minimal parsing in OCaml later, note that key-value pairs aren't split correctly, but when joined backed together they retain the original value).

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -66,11 +66,23 @@ module Raw = struct
         contents : string list;
         label_cmt : string option;
         legacy_labels : string;
+        attributes : string option;
         errors : Output.t list;
       }
 
-  let make ~loc ~section ~header ~contents ~label_cmt ~legacy_labels ~errors =
-    Any { loc; section; header; contents; label_cmt; legacy_labels; errors }
+  let make ~loc ~section ~header ~contents ~label_cmt ~legacy_labels ~attributes
+      ~errors =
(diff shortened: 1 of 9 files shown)


edwintorok:
(this PR might need some wider testing to check it doesn't break backwards compatibility, is there a larger corpus you'd normally test changes like this on? e.g. the realworldocaml book, anything else?)

samoht:
Testing on RWO would indeed be a great idea - and checking that the resulting diff doesn't contain surprising results

PR #450, opened 2024-03-04
Add an 'exec' label to execute include OCaml blocks

gpetiot:
Follow up on #446

The code of https://github.com/panglesd/mdx/commit/85b9142fb0d83b187989bf330d37fa5b9ee8556b is already squashed, so that's a more difficult to go back in time. I couldn't re-create that behavior yet.

@panglesd if you see something weird with the exec/skip interaction for the ocaml includes I'm interested :)

### DIFF
--- a/lib/block.ml
+++ b/lib/block.ml
@@ -30,6 +30,16 @@ let locate_errors ~loc r =
     (fun l -> List.map (fun (`Msg m) -> `Msg (locate_error_msg ~loc m)) l)
     r
 
+module OCaml_kind = struct
+  type t = Impl | Intf
+
+  let infer_from_file file =
+    match Filename.(remove_extension (basename file), extension file) with
+    | _, (".ml" | ".mlt" | ".eliom") -> Some Impl
+    | _, (".mli" | ".eliomi") -> Some Intf
+    | _ -> None
+end
+
 module Header = struct
   type t = Shell of [ `Sh | `Bash ] | OCaml | Other of string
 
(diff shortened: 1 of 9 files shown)


panglesd:
Will have a look, but we need some logic to add this to the dependencies as pointed [here](https://github.com/realworldocaml/mdx/pull/446#issuecomment-1976542813)!

lib/block.ml
```diff
@@ -422,23 +455,38 @@ let mk_toplevel ~loc ~config ~contents ~errors =
...
+            match OCaml_kind.infer_from_file file_included with
+            | Some _ -> `OCaml
+            | None -> `Other)
+        | _ -> `Other
+      in
+      match kind with
+      | `OCaml ->
+          let kind =
+            Util.Option.value ~default:OCaml_kind.Impl
+              (OCaml_kind.infer_from_file file_included)
```
panglesd, line 472: Nitpick: `Ocaml_kind.infer_from_file` is called twice in a few lines... Maybe store the result in the `` `OCaml `` polymorphic variant?

(This would also remove the call to `Option.value`)

test/bin/mdx-test/expect/code/test-case.md
```diff
@@ -29,5 +29,6 @@ module type Foo = sig type t end
 
 ```ocaml skip
 # Pipe.f ();;
-- : unit
+Line 1, characters 1-7:
+Error: Unbound module Pipe
```
panglesd, line 33: Do you have an idea why this PR is making this change? It seems wrong!

gpetiot:
@panglesd I completely forgot what this is about, is this still needed? I'm fine with closing it and reopening something cleaner when needed.

gpetiot:
Not needed anymore!

PR #457, opened 2024-11-24
support OCaml 5.3

anmonteiro:
the expect tests still have some failures like this if ran on 5.3. Any advice on how to make them conditional?

```
File "test/bin/mdx-test/expect/simple-mli/test-case.mli", line 1, characters 0-0:
diff --git a/_build/default/test/bin/mdx-test/expect/simple-mli/test-case.mli b/_build/default/test/bin/mdx-test/expect/simple-mli.actual
index ad197127..ad54569e 100644
--- a/_build/default/test/bin/mdx-test/expect/simple-mli/test-case.mli
+++ b/_build/default/test/bin/mdx-test/expect/simple-mli.actual
@@ -56,7 +56,7 @@ val baz : string
 ][
 {err@mdx-error[
 Line 1, characters 15-18:
-Error: This expression has type string but an expression was expected of type
+Error: This constant has type string but an expression was expected of type
          int
 ]err}]}
 *)
```

### DIFF
--- a/lib/top/compat_top.ml
+++ b/lib/top/compat_top.ml
@@ -96,7 +96,11 @@ let extension_constructor
   ; ext_private
   ; ext_loc
   ; ext_attributes
-#if OCAML_VERSION >= (4, 11, 0)
+#if OCAML_VERSION >= (5, 3, 0)
+  ; ext_uid = Uid.mk
+      ~current_unit:
+        (Some (Unit_info.make ~source_file:"mdx.ml" Impl "mdx"))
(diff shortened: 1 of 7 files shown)


samoht:
@voodoos is this the right way to fix this? What about using `~current_unit:None` instead?

samoht:
I've rebased on top of `main` to fix the tests

samoht:
Thanks! I've simplified the support as I don't think the tooling needs anything complicated here.

PR #466, opened 2025-03-22
Alternative method for slicing errors in mli_parser

jonludlam:
This is slightly cleaner than #462 I think!

cc @samoht

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+### Unreleased
+
+#### Changed
+
+- Improved implementation of handling of skipped blocks in mli/mld
+  files (#466, @jonludlam)
+
 ### 2.5.0
 
 #### Added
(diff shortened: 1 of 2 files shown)


jonludlam:
Getting the text of the error from odoc-parser's AST means that you end up with the text of the error with leading whitespace removed, unlike what the current mechanism does. However, since the output code [doesn't put any indentation in anyway](https://github.com/realworldocaml/mdx/blob/1f83f73e06233ea8745cd0c9473613d40e6999bc/lib/block.ml#L186) this might not be a problem.

The newer parser in odoc 3 will allow the extraction of the text of code blocks verbatim, but it is more restrictive about the sorts of tags that can appear in code blocks, so there's some design work needed before that can be dropped in.

samoht:
Looks great! Do you mind adding a change entry (or appending your name and PR to the existing one)?

samoht:
Thanks!

PR #470, opened 2025-07-29
minor fix to get it working with ocaml 5.4

chetmurthy:
the tests didn't pass, but I think it was something innocuous.


kit-ty-kate:
This PR seems to have issues. I've tried it on packages from opam-repository and i've seen failures in the yocaml project that seems related:
```
#=== ERROR while compiling yocaml_cmarkit.2.3.0 ===============================#
# context              2.4.1 | linux/x86_64 | ocaml-base-compiler.5.4.0~beta1 | file:///home/opam/opam-repository
# path                 ~/.opam/5.4/.opam-switch/build/yocaml_cmarkit.2.3.0
# command              ~/.opam/5.4/bin/dune build -p yocaml_cmarkit -j 1 @install @runtest
# exit-code            1
# env-file             ~/.opam/log/yocaml_cmarkit-1571-148bcc.env
# output-file          ~/.opam/log/yocaml_cmarkit-1571-148bcc.out
### output ###
# File "doc/data_validation.mld", line 1, characters 0-0:
# /usr/bin/git --no-pager diff --no-index --color=always -u _build/default/doc/data_validation.mld _build/default/doc/.mdx/data_validation.mld.corrected
# diff --git a/_build/default/doc/data_validation.mld b/_build/default/doc/.mdx/data_validation.mld.corrected
# index 55f7937..81f3b32 100644
# --- a/_build/default/doc/data_validation.mld
# +++ b/_build/default/doc/.mdx/data_validation.mld.corrected
# @@ -120,7 +120,7 @@ Let's try to validate an incorrect entry. For example, let's try to treat a
#  {@ocaml[
#  # V.int (D.bool true) ;;
#  - : int V.validated_value =
# -Error (V.Invalid_shape {V.expected = "int"; given = Yocaml.Data.Bool true})
# +Error (V.Invalid_shape {V.expected = "int"; given = Yocaml__Data.Bool true})
#  ]}
[...]
```

chetmurthy:
I'll debug this (and any other CI failures that appear in that run above), then report back.

chetmurthy:
(1) I installed yocaml_cmarkit.2.3.0 in both ocaml 5.3.0 & 5.4.0~beta1
(2) In the former (of course) the tests pass; in the latter, the tests fail as you note
(3) I then extracted a minimal trace of the test and run it in each toplevel.
(4) As you can see, the problem is not related to `mdx`.  Instead, it appears that
the way that the yocaml archives/libraries have been built, has changed between
ocaml versions.

I searched the Changes file for this release, but nothing jumps out as relevant.  Does
this tickle any memories for you?  I'm not a dune user, so it would be .... troublesome
for me to dig into this.  Perhaps the yocaml maintainer could have a look?

(5) Looking at the diff output by the test-runner [ok, I scanned the test-diffs] it appears that
all the diffs are from this change in module-name from Yocaml.Data to Yocaml__Data.
```
OCaml version 5.3.0

# #use "topfind";;
# #require "yocaml";;
# module D = Yocaml.Data;;
# module V = Yocaml.Data.Validation;;
# V.int (D.bool true) ;;
- : int V.validated_value =
Error (V.Invalid_shape {V.expected = "int"; given = Yocaml.Data.Bool true})

Process OCaml finished
OCaml version 5.4.0~beta1

# #use "topfind";;
# #require "yocaml";;
# module D = Yocaml.Data;;
# module V = Yocaml.Data.Validation;;
# V.int (D.bool true) ;;
- : int V.validated_value =
Error (V.Invalid_shape {V.expected = "int"; given = Yocaml__Data.Bool true})
```

chetmurthy:
I looked at the logs from ocaml-ci (from the CI run), but couldn't find anything that seemed like I should investigate?  I couldn't even find the run that produced the error above.  If there are other errors I should look into, I'm happy to do so, if you can point me at how to see the logs.

Reproduction is easy (thank you, opam!) -- I just need to know which packages failed.

kit-ty-kate:
Good catch, i've opened a ticket upstream in https://github.com/ocaml/ocaml/issues/14184

jonludlam:
Now that the new beta of 5.4 has been released with the fix for #14184 we should probably rerun the tests.

chetmurthy:
<strike>I installed this package on 5.4.0 beta2, and it built, but (as before) the buit-in tests failed.  The failure is innocuous, and looking at  https://github.com/realworldocaml/mdx/blob/04347a5c19b9e791b7cad0d920be2e725484e15a/test/bin/mdx-test/expect/warnings/test-case.md?plain=1#L22 it is somewhat clear how to fix it -- I need to duplicate and then modify that test-stanza with one of them having a range of versions [4.12..5.3.0].  I can't find any indication of how to do this in the documentation -- if you could point me at how to do this, I can update the PR with the requisite change, so the tests all pass.

```
### output ###
# @@ -25,9 +25,9 @@ let x { x } = x
# [...]
#  ```mdx-error
#  Line 2, characters 9-14:
# -Warning 9 [missing-record-field-pattern]: the following labels are not bound 
# -y
# -Either bind these labels explicitly or add '; _' to the pattern.
# +Warning 9 [missing-record-field-pattern]: the following labels are not bound
# +  in this record pattern: y.
# +  Either bind these labels explicitly or add ; _ to the pattern.
#  ```
#
```
</strike>

chetmurthy:
I figured out how to do the version constraint thing; will checkin a fix that makes the builtin tests pass.

chetmurthy:
There are a few test-failures in the CI run.

(1) some openbsd failures, don't know what's causing those
(2) some system packages failed to install, aborting some of the test runs

But most everything else passed.

P.S. And there seems to be some lint-format error: it didn't recognize CPPO directives.

jonludlam:
OK, I think this is green enough now :-)

PR #476, opened 2026-05-19
Report location of hangs

talex5:
If a test hangs it's useful to know where.

This commit adds a `?progress` callback to `Mdx_test.run_exn`, which is called when starting each test with its location. The mdx-test binary installs a signal handler for SIGINT which displays the last-known location when called.

Example:

    $ dune exec -- ocaml-mdx-test test.md
    ^C
    test.md:7: interrupted

Sending USR1 will print the current location without exiting.

It also spawns a monitoring thread that reports when part of a test is taking a long time, e.g.

    $ dune exec -- ocaml-mdx-test test.md
    warning: test.md:7 has been running for 5 seconds
    warning: test.md:7 has been running for 10 seconds
    warning: test.md:10 has been running for 5 seconds
    warning: test.md:10 has been running for 10 seconds
    warning: test.md:10 has been running for 20 seconds

The timeout doubles each time. This is intended for use in CI, where currently you just get told that your tests timed out after an hour, with no idea where.

There are a couple of problems when used with dune however:

1. On Ctrl-C, dune terminates the test with SIGKILL, so there's no opportunity to print the current progress.

2. Dune buffers output by default, so you don't see the warnings until the test is over (if it finishes at all), which is pointless. So for this to be useful, you need to do `dune test --no-buffer`.

I expect these could both be fixed with modifications to dune.

TODO:
- [x] Cram tests currently only report the location of the block, not the individual line that's hanging.
- [x] Check this doesn't break things on Windows.

Let me know if this seems like a sensible approach.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 
 - Mark the tests as conflicting with dune 3.24.0 due to a bug in
   that release; see https://github.com/ocaml/dune/issues/14724 (@avsm)
+- If a test takes longer than 5 seconds, display a warning with the test's location
+  Also display the location on Ctrl-C (#476, @talex5 @avsm).
 
 ### 2.5.2
 
--- a/README.md
+++ b/README.md
@@ -440,3 +440,18 @@ Those variables are then available in the subsequent blocks
     bar
     - : unit = ()
     ```
+
+### Hanging tests
+
+If a test takes a long time, MDX will write a warning to stderr with the location.
+For example:
+
+<!-- $MDX skip -->
+```sh
+$ dune exec -- ocaml-mdx-test test.md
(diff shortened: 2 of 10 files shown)


avsm:
This seems sensible, and the `dune test --no-buffer` should be the default in most CIs anyway.

talex5:
OK, I've added more accurate locations for cram tests and disabled trying to install signal handlers on Windows.

The CI failure is unrelated and I've reported it separately: #477.

avsm:
I chucked when reading 89833bab42c8a460ee361a19e3b0c26d095b603c. That test does seem a little aggressive!

avsm:
Rebased against main

PR #478, opened 2026-06-02
Fix the target called by the mdx test runner

### LINKED ISSUE #477: MDX CI is failing
I just pushed the current `main` branch to a new branch in my fork to check, and the tests are failing with e.g.
```
File "test/bin/mdx-test/failure/dune.inc", lines 2-9, characters 0-237:
2 | (rule
3 |  (target block-locations.actual)
4 |  (deps (package mdx) (source_tree block-locations))
5 |  (action
6 |   (with-accepted-exit-codes 1
7 |    (with-outputs-to %{target}
8 |     (chdir block-locations
9 |      (run %{bin:ocaml-mdx} test test-case.md))))))
(cd _build/default/test/bin/mdx-test/failure/block-locations && ../../../../../bin/main.exe test test-case.md) &> _build/default/test/bin/mdx-test/failure/block-locations.actual
Command exited with code 125.
File "test/bin/mdx-test/misc/inserting-double-semicolon/dune", lines 1-8, characters 0-162:
1 | (rule
2 |  (target test-case.actual)
3 |  (deps
4 |   (package mdx))
5 |  (action
6 |   (with-outputs-to
7 |    %{target}
8 |    (run %{bin:ocaml-mdx} test --output=- %{dep:test-case.md}))))
(cd _build/default/test/bin/mdx-test/misc/inserting-double-semicolon && ../../../../../bin/main.exe test --output=- test-case.md) &> _build/default/test/bin/mdx-test/misc/inserting-double-semicolon/test-case.actual
Command exited with code 125.
```

Adding some debugging, it looks like the error is:
```
 ocaml-mdx: internal error, uncaught exception:
           Unix.Unix_error(Unix.ENOENT, "create_process", "../../../../../bin/test/main.exe")
```

The test does use `(deps (package mdx))`, but maybe that isn't enough?

I'm not sure what version of dune it's using. Possibly a nightly, based on https://get.dune.build/install.

shonfeder:
When mdx's tests are run, they use the private executable from the `_build` directory, thus invoking with the private `main.exe` name, which causes us to hit line 33:

```
    | "main.exe" -> dir / "test" / "main.exe"
```

However, under the stricter rules imposed by https://github.com/ocaml/dune/pull/14432 (pending release in dune 3.24), there is cleaner separation when executing binaries, and the `.exe` target isn't available in some scenarios (notably, when running with `dune install mdx --with-test`). The byte code target is always available in this location, and, IIUC, is the intended target.

AFAIU, this should not be visible to users at all, since the would only call mdx thru the public executable installed as `ocaml-mdx`.

Should close https://github.com/realworldocaml/mdx/issues/477

The diagnosis and solution was provided by @alizter at https://github.com/ocaml/dune/issues/14724#issuecomment-4549904300

### DIFF
--- a/.github/workflows/dune-build.yml
+++ b/.github/workflows/dune-build.yml
@@ -14,6 +14,4 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
       - name: Use dune
-        uses: ocaml-dune/setup-dune@v0.0.1
-        with:
-          automagic: true
+        uses: ocaml-dune/setup-dune@v2
--- a/bin/test.ml
+++ b/bin/test.ml
@@ -30,7 +30,7 @@ let run (`Setup ()) _ _ _ _ _ _ _ _ _ _ =
   let dir = Filename.dirname Sys.argv.(0) in
   let cmd =
     match base with
-    | "main.exe" -> dir / "test" / "main.exe"
+    | "main.exe" -> dir / "test" / "main.bc"
     | x when String.length x > 6 && String.sub x 0 6 = "ocaml-" ->
         let x_without_ext, x_ext = split_exe_extension x in
         (dir / x_without_ext) ^ "-test" ^ x_ext


avsm:
Everything's still failing in ocaml-ci, but I'll try to advance this along with #479 to get things working again.

Alizter:
This looks like a separate failure to the one being fixed here. Seems to be because of https://github.com/ocaml/dune/pull/15156/.

shonfeder:
I've proposed a fix in https://github.com/realworldocaml/mdx/pull/482 .

IIRC, this fix was proposed while we working on the prelease for 3.24, before https://github.com/ocaml/dune/pull/15156 landed, and since we had added in the filter on dune versions in the tests on the opam-repository for published versions at that point, we didn't catch or notice this additional issues after the path representation change was added in to the release candidate.

PR #482, opened 2026-07-22
Fix path representations that are causing expect-like tests to fail in dune 3.24

shonfeder:
In https://github.com/ocaml/dune/pull/15156 , release in dune 3.24, the representation of paths was change so that paths would always include their local directory prefix. Test fixtures that depend on the unnormalized representation of paths coming from dune start breaking as a result.

The two changes offered here fix the test breakage reported at https://github.com/realworldocaml/mdx/pull/478#issuecomment-5036254826 and should make https://github.com/realworldocaml/mdx/pull/479 unnecessary for new release.

An alternative approach would be just promote the changes to the file representation into the test fixtures. However, normalizing and fully specifying path representations is the more robust choice if they get printed into text fixtures, IMO.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+### unreleased
+
+### Changed
+
+- Paths are normalized to drop the directory prefix ("./" on MacOs and Linux)
+  before pretty printing them (#482, @shonfeder).
+
 ### 2.6.0
 
 #### Added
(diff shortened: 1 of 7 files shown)


test/bin/misc-test-cases/mdx-pp/dune
```diff
@@ -10,8 +10,6 @@
 
 (rule
  (alias runtest)
- (deps
```
shonfeder, line 13: This is not the fix... wip...

shonfeder:
Not sure if a changelog entry is actually wanted for this, but it could be user-visible so I've offered one.

